PR #2, opened 2015-03-01
users can sheet and view a list of their own sheets

robertpulson:


spec/features/showing_all_tweets_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  scenario 'when visiting the homepage' do
+    visit '/'
+    expect(page).to have_content('This is a Sheet!')
+  end
+  
+  scenario 'ordered by time' do
+    visit '/'
+    page.body.index('This is also a Sheet!').should < page.body.index('This is a Sheet!')
```
tansaku: you should use the new rspec 'expect' syntax - we challenge you to re-write this with 'expect'

spec/features/showing_all_tweets_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'spec_helper'
+
+feature 'User browses the list of tweets' do
+
+  before(:each) {
```
tansaku: when multi-line a block should be 'do ... end' and not '{}'

spec/features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,219 @@
+# Taken from the cucumber-rails project.
```
tansaku: you aren't using cucumber, don't include files like this that are not necessary (they do more harm than good) - note also that your directory structure here is strange.  The contents of step_definitions and support directories are completely redundant and should be removed.

app/server.rb
```diff
@@ -0,0 +1,90 @@
...
+  erb :index
+end
+
+get '/users/new' do
+  @user = User.new
+  erb :'users/new'
+end
+
+post '/users' do
+  @user = User.create(:name => params[:name],
```
tansaku: adjusting params could be extracted to a separate method to reduce the size of this one

app/server.rb
```diff
@@ -0,0 +1,90 @@
...
+  session[:user_id] = nil
+  redirect to('/')
+end
+
+post '/sheets/new' do
+  erb :'sheets/new'
+end
+
+post '/sheet' do
+   Sheet.create(:text => params[:sheet],
```
tansaku: indenting is incorrect here - 3 spaces in bad

app/server.rb
```diff
@@ -0,0 +1,90 @@
...
+DataMapper.auto_upgrade!
+
+enable :sessions
+set :session_secret, 'super secret'
+use Rack::Flash
+use Rack::MethodOverride
+
+set :public_folder, Proc.new { File.join(root, '..', "public") }
+
+get '/' do
```
tansaku: we recommend using Sinatra::Application here

app/server.rb
```diff
@@ -0,0 +1,90 @@
...
+
+post '/user/profile' do
+  @sheets = Sheet.all(:user_id => session[:user_id])
+  erb :'users/profile'
+end
+
+helpers do
+
+  def current_user
+    @current_user ||=User.get(session[:user_id]) if session[:user_id]
```
tansaku: ensure you have a space after '||='

spec/sheet_spec.rb
```diff
@@ -0,0 +1,7 @@
+require 'spec_helper'
```
tansaku: this file is not really doing anything - so you should remove it, or add unit tests to it.

app/views/layout.erb
```diff
@@ -0,0 +1,49 @@
...
+    <title>5hitter</title>
+    <link rel='stylesheet' type='text/css' href='style.css'>
+  </head>
+
+  <body>
+
+    <h1>5hitter</h1>
+  
+    <% if flash[:notice] %>
+      <div id='notice'>
```
tansaku: don't use divs

app/helpers/session
```diff
@@ -0,0 +1,6 @@
+    def sign_in(username, password)
```
tansaku: this seems like it's not doing anything - please delete

spike01:
:checkered_flag:

tansaku:
Great to see it all working including sign up and log in etc.  Functionality is all there.  Please just ensure to remove any extraneous code and files that are not being used.  Well done for completing the challenge - you've earned the 'Data' badge.

app/server.rb
```diff
@@ -0,0 +1,90 @@
+require 'sinatra'
+require 'data_mapper'
+require 'rack-flash'
+
+env = ENV['RACK_ENV'] || 'development'
```
tansaku: consider extracting all the database setup to a separate file

PR #3, opened 2015-03-01
Chitter [Costas]

costassarris:


Gemfile
```diff
@@ -0,0 +1,18 @@
...
+gem 'database_cleaner'
+gem 'bcrypt-ruby'
+gem 'rack-flash3'
+gem 'sinatra-partial'
+
+group :development, :test do
+  gem 'capybara'
+  gem 'cucumber'
+  gem 'cucumber-sinatra'
+  gem 'launchy'
```
danldb: You are not using cucumber, so it should not appear in the gemfile.

app/controllers/application.rb
```diff
@@ -0,0 +1,4 @@
+get '/' do
+  @peeps = Peep.all
+  erb :index
```
danldb: How would you go about creating a specific order for the peeps? The brief specifies chronological order.

app/controllers/peeps.rb
```diff
@@ -0,0 +1,6 @@
+post '/peeps' do
+  content = params["content"]
```
danldb: ``` post '/peep/new' ``` could be a more descriptive path

app/server.rb
```diff
@@ -0,0 +1,24 @@
...
+require_relative 'controllers/users'
+require_relative 'controllers/peeps'
+require_relative 'controllers/sessions'
+require_relative 'controllers/application'
+
+
+
+enable :sessions
+set :session_secret, 'super secret'
+use Rack::Flash
```
danldb: 'super secret' is not very secret - use a randomly generated string

app/server.rb
```diff
@@ -0,0 +1,24 @@
+require 'sinatra'
+require 'data_mapper'
+require 'rack-flash'
+require 'sinatra/partial'
+
+DataMapper::Logger.new($stdout, :debug)
+
```
danldb: having the logger here is quite annoying - use it for specific debugging, then switch it off or set it to output to a file instead of stdout

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  scenario "when browing the homepage" do
+    expect(Peep.count).to be 0
+    visit '/'
+    post_peep("Hello Chitter", "Costas")
+    expect(Peep.count).to eq 1
+    peep = Peep.first
+    expect(peep.content).to eq "Hello Chitter"
+    expect(peep.user).to eq "Costas"
+  end
+
```
danldb: multiple expectations
You could consider testing the database with unit tests, thus keeping the responsibility of the features to test the user experience - so for instance, instead of testing Peep.count to eq 1, test if you can see 1 peep on the page.

spec/peep_spec.rb
```diff
@@ -0,0 +1,19 @@
...
+    it 'should be created and then retrieved from the db' do
+      expect(Peep.count).to eq 0
+      Peep.create(content: 'This is my first peep')
+      expect(Peep.count).to eq 1
+      peep = Peep.first
+      expect(peep.content).to eq 'This is my first peep'
+      peep.destroy
+      expect(Peep.count).to eq 0
+    end
+
```
danldb: multiple expectations - split into 3 tests

danldb:
:checkered_flag: 

Hi Costas!

Good work - you have clearly grasped the basics of sinatra and capybara testing. 

* the readme should contain instructions for installation and execution as well as a rundown of the learning objectives, technologies, screenshots etc
* More unit tests for the models would have been preferable
* I noticed that your flash notices were duplicating - try making a user that already exists, you get 2 different flash notices saying the same thing

Well done - let us know if you have any questions

PR #5, opened 2015-03-01
Chitter

kevinlanzon:


Rakefile
```diff
@@ -0,0 +1,12 @@
+require 'data_mapper'
+require './app/data_mapper_setup'
+
+task :auto_upgrade
```
spike01: Missing `do`

app/controllers/peeps.rb
```diff
@@ -0,0 +1,13 @@
+class Chitter < Sinatra::Base
+
+  get '/peeps/new' do
+    erb :"peeps/new"
+  end
+
+  post '/peeps' do
+    @peep = Peep.create(:post => params[:post],
+                     :user_name => params[:user_name])
```
spike01: Indentation!

app/helpers/application.rb
```diff
@@ -0,0 +1,10 @@
+class Chitter < Sinatra::Base
+
+  helpers do
+
+    def current_user
+      @current_user ||=User.get(session[:user_id]) if session[:user_id]
```
spike01: Space after the `||=`

app/views/flash.erb
```diff
@@ -0,0 +1,15 @@
+<center>
```
spike01: Centering should happen in your CSS, not your tags

spec/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      peep = Peep.first
+      expect(peep.user_name).to eq('klanzon')
+      expect(peep.post).to eq('this is my first peep!')
+      peep.destroy
+      expect(Peep.count).to eq(0)
+    end
+
+  end
+
+end
```
spike01: Yes, this test was in the bookmark manager tutorial to show you how Datamapper works, but no, this should not be in your challenge

app/views/index.erb
```diff
@@ -0,0 +1,24 @@
...
+  <% @peeps.each do |peep| %>
+    <% @peeps %>
+    <li><%= peep.post%></li>
+  <% end %>
+</ul>
+<% end %>
+<div id="new-peep">
+  <form action="/peeps" method="post">
+    <li><p>Create Peep: <input type="text" size="50" name="post"></p></li>
+    <li><p>Username: <input type="text" name="user_name"></p></li>
```
spike01: Would this not be the user that is currently logged in?

spec/features/user_management_spec.rb
```diff
@@ -0,0 +1,64 @@
...
+include SessionHelpers
+
+feature "User signs up" do
+
+  scenario "when first visiting the homepage" do
+    expect{ sign_up }.to change(User, :count).by(1)
+    expect(page).to have_content("Welcome, chitty1")
+    expect(User.first.name).to eq("John Smith")
+    expect(User.first.username).to eq("chitty1")
+    expect(User.first.email).to eq("test@test.com")
```
spike01: Don't test your database objects directly in a feature test

spike01:
I've not left many comments on this because a.) I couldn't get any tests to run or even the server to start up without a lot tweaking and b.) most of what I'll say will be covered when we do a session with your array. The main things to think about are: 

- cleaning up your tests
- testing at the right layer
- not having multiple expectations
- implementing the features that were required on the spec

:construction:

app/views/index.erb
tansaku: kevin; sorry I didn't immediately follow up on this - end of the day and all that - looking over it I think @spike01 's point is that you shouldn't ask the user to specify who creates the peep - you should set the user who is currently logged in as the user associated with the peep automatically, and BTW, do ask follow up questions within the pull request - it's the developer's way!

Rakefile
```diff
@@ -0,0 +1,12 @@
+require 'data_mapper'
+require './app/data_mapper_setup'
```
spike01: You'll need to require the models as well for this to work, you could also include the server

PR #6, opened 2015-03-01
challenge submission

GabeMaker:


spec/demo_cheet_spec.rb
```diff
@@ -0,0 +1,28 @@
+# require 'spec_helper'
+
+describe Cheet do
```
stephenlloyd: all the tests dont run because of this constant Cheet. Should you be requiring that file?

spec/demo_cheet_spec.rb
```diff
@@ -0,0 +1,28 @@
...
+                  time: 'Just now')
+      # We ask the database how many cheets we have, it should be 1
+      expect(Cheet.count).to eq(1)
+      # Let's get the first (and only) cheet from the database
+      cheet = Cheet.first
+      # Now it has all properties that it was saved with.
+      expect(cheet.text).to eq('Hello World')
+      expect(cheet.time).to eq('Just now')
+      # If we want to, we can destroy it
+      # cheet.destroy
```
stephenlloyd: Please no more comments. No one reads them and they make your code ugly.

spec/features/all_cheets_spec.rb
```diff
@@ -0,0 +1,14 @@
+# require 'spec_helper'
+
+feature "User sees all cheets" do
+  before(:each) {
+    Cheet.create( text: "Hello World!",
+                  time: Time.now.to_s)
+  }
+
+  scenario "when on home page" do
```
stephenlloyd: try and make this name more specific e.g. 'Hompage has correct text'

spec/features/cheeting_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+
+feature "User makes a new cheet" do
+  scenario "when on the homepage" do
```
stephenlloyd: again, When I run rspec all I see it 'when on the homepage' twice. This is a far more complicated test than 'when on the homepage'. context 'on the home page',  'creating a tweet', 'displaying a tweet'.

lib/cheet.rb
```diff
@@ -0,0 +1,9 @@
+class Cheet
```
stephenlloyd: it's a small one now but we could call this file model rather than lib. The definition is quite though but usually if you have a file which is mapped to an object in a database it's a model. Something which has functionality and no reason to be saved in a database you could save in a library directory.

app.rb
```diff
@@ -0,0 +1,79 @@
...
+DataMapper.finalize
+DataMapper.auto_upgrade!
+
+enable :sessions
+set :session_secret, 'encryption key'
+use Rack::Flash
+use Rack::MethodOverride
+
+get '/' do
+  @cheets = Cheet.all
```
stephenlloyd: could you do this so they're definitely in the correct chronological order. Right now they're just i the same order as they appear in the database which might not always be chronological.

app.rb
```diff
@@ -0,0 +1,79 @@
...
+end
+
+post '/sessions' do
+  username, password = params[:username], params[:password]
+  user = User.authenticate(username, password)
+  if user
+    session[:user_id] = user.id
+    redirect to('/')
+  else
+    flash[:errors] = ["The email or password is incorrect"]
```
stephenlloyd: this error would persist once they get the password correct. Maybe you want to user .now on the flash message to ensure it's only shown once.

app.rb
```diff
@@ -0,0 +1,79 @@
...
+env = ENV['RACK_ENV'] || 'development'
+
+DataMapper.setup(:default, "postgres://localhost/chitter_#{env}")
+
+require './lib/cheet'
+require './lib/user'
+
+DataMapper.finalize
+DataMapper.auto_upgrade!
+
```
stephenlloyd: I really like the modular way to create a sinatra application over the classical way. Look up the differences and use the one which is more widely accepted.

app.rb
```diff
@@ -0,0 +1,79 @@
...
+delete '/sessions' do
+  flash[:notice] = "Good bye!"
+  session[:user_id] = nil
+  redirect to('/')
+end
+
+helpers do
+  def current_user
+    @current_user ||=User.get(session[:user_id]) if session[:user_id]
```
stephenlloyd: ||= User.
It looks messy without the space

views/layout.erb
```diff
@@ -0,0 +1,34 @@
+<body>
+
```
stephenlloyd: I had no idea how to sign in so I couldn't get to the point of seeing my username displayed. So I won't mark this section of the test.

stephenlloyd:
This was good and you get a sticker. However it would have been nice if you thought about the user experience. As it stands it would be impossible for a user to sign in. I did actually manage it by going directly to the url which is how i found the error persisted issue. All and all really good. :checkered_flag:

spec/demo_cheet_spec.rb
GabeMaker: They work fine on my setup.
$ rspec => all green
Am I missing something?

GabeMaker:
Thanks for your notes Stephen. The challenge became mainly about time management for me which is why I submitted it in such a state - comments, no log-in, no css, etc - I'm sorry! I got to the end of the weekend and I was exhausted and submitted what I had. I'll really try to get a better final product handed in next time :) I did learn a lot more about Sinatra over the weekend though and this challenge brings a lot of things together and allows for plenty of extension.

PR #7, opened 2015-03-01
Challenge submission

alexparkinson1:


spec/features/session.rb
```diff
@@ -0,0 +1,20 @@
+module SessionHelpers
```
tansaku: seems like this file is not being used - to use it you would need to 'include' it in the correct places in your spec files

lib/views/layout.erb
```diff
@@ -0,0 +1,22 @@
...
+</head>
+<body>
+  <% if current_user %>
+    Welcome, <%= current_user.email %>
+    <form method="post" action="/sessions">
+      <input type="hidden" name="_method" value="delete">
+      <input type="submit" value="Sign out">
+    </form>
+  <% end %>
+  <% if flash[:errors] && !flash[:errors].empty? %>
```
tansaku: you are only displaying flash[:error] here and not flash[:notice] - this is why your test is failing

spec/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
```
tansaku: ideally this should be in the features directory

spec/posting_peeps_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
```
tansaku: ideally this should be in the features directory

lib/server.rb
```diff
@@ -0,0 +1,80 @@
...
+  end
+
+  post '/sessions' do
+    email, password = params[:email], params[:password]
+    user = User.authenticate(email, password)
+    if user
+      session[:user_id] = user.id
+      redirect to('/')
+    else
+      flash[:errors] = ["The email or password is incorrect"]
```
tansaku: this should be 'flash.now[:errors]' in order to avoid it persisting over two requests

lib/server.rb
```diff
@@ -0,0 +1,80 @@
+require 'data_mapper'
+require 'sinatra/base'
+require 'rack-flash'
+
+env = ENV['RACK_ENV'] || 'development'
+
+DataMapper.setup(:default, "postgres://localhost/bookmark_manager_#{env}")
```
tansaku: ideally database setup would be extracted to a setup file that was then required here

lib/server.rb
```diff
@@ -0,0 +1,80 @@
...
+    def current_user
+      @current_user ||=User.get(session[:user_id]) if session[:user_id]
+    end
+  end
+
+  enable :sessions
+  set :session_secret, 'super secret'
+
+  get '/' do
+    @peeps = Peep.all
```
tansaku: how would you adjust this to ensure the peeps are in chronological order?

tansaku:
Great work getting the functionality working.  You do have a failing test, which is a bad sign - we've indicated how to fix that in the comments.  Please fix that up and you'll be eligble for the 'data' badge.  Please do also neaten up according to other comments (flash.now, position of feature specs etc.), but well done completing the challenge! :checkered_flag:

PR #9, opened 2015-03-01
Challenge submission

guidovitafinzi:


spec/spec_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  config.before(:each) do
+    DatabaseCleaner.start
+  end
+
+  config.after(:each) do
+    DatabaseCleaner.clean
+  end
+  
+    expectations.include_chain_clauses_in_custom_matcher_descriptions = true
```
stephenlloyd: this line was breaking all your tests

spec/spec_helper.rb
```diff
@@ -0,0 +1,30 @@
...
+  end
```
stephenlloyd: as was this line

app/server.rb
```diff
@@ -0,0 +1,83 @@
...
+end
+
+post '/sessions' do
+  email, password = params[:email], params[:password]
+  user = User.authenticate(email, password)
+  if user
+    session[:user_id] = user.id
+    redirect to('/')
+  else
+    flash[:errors] = ["The email or password are incorrect"]
```
stephenlloyd: this error persists when you get it correct the second time. If you change it to flash[:errors].now it will only be shown once.

app/views/layout.erb
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Chitter</title>
+</head>
+<body>
+<% if current_user %>
```
stephenlloyd: You have some logic here for seeing if there is a user but in order to log in i had to manually type out the url. Maybe you should have given me a link to click on so I can easily sign in/up when required.

app/server.rb
```diff
@@ -0,0 +1,83 @@
...
+enable :sessions
+set :session_secret, 'super secret'
+use Rack::Flash
+use Rack::MethodOverride
+
+DataMapper.finalize
+DataMapper.auto_upgrade!
+
+get '/' do 
+  @peeps = Peep.all
```
stephenlloyd: nothing here is specifying the order. It would be nice to ensure they are always in chronological order.

app/models/peep.rb
```diff
@@ -0,0 +1,9 @@
+class Peep
+
+  include DataMapper::Resource
+
+  property :id,   Serial
+  property :time,   DateTime
```
stephenlloyd: this is a datatime. In your tests you're using a string to create this...?

spec/features/listing_all_peeps_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'spec_helper'
+
+feature 'User browses the list of peeps' do
+
+  before(:each) {
+    Peep.create(text: 'my first peep',
+                time: '01/01/2015 00:0')
```
stephenlloyd: why is this a string and not a datetime. You could have done Time.now

stephenlloyd:
This is good. Definitely enough to get the sticker bit it would have been nice if you thought about the user experience more. Although what you have done is clean.

stephenlloyd:
:checkered_flag:

PR #10, opened 2015-03-01
Chitter challenge

tommasobratto:


app/models/message.rb
```diff
@@ -0,0 +1,9 @@
+class Message
+
+  include DataMapper::Resource
+
+  property :id,       Serial
+  property :content,  Text
+
+  # belongs_to :user
```
pitchinvasion: Remove comments!

app/models/user.rb
```diff
@@ -0,0 +1,25 @@
...
+  property :name,     String 
+  property :email,    String
+  property :password_digest, Text
+
+  def password=(password)
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(name, password)
+    user = self.first(:name => name)
```
pitchinvasion: you don't need self.first here, you can just call first as self is implicit in ruby

app/models/user.rb
```diff
@@ -0,0 +1,25 @@
...
+    if user && BCrypt::Password.new(user.password_digest) == password
+      user
+    else
+      nil
+    end
+  end
+  # has n, :messages, :through => :message
```
pitchinvasion: comments

app/views/users/new.erb
```diff
@@ -0,0 +1,9 @@
+
+Fill the spaces to sign up and post messages today
+  <form action="/users/new" method="post">
+  Email:<input type="text" name="email" value=<% @user.email %>>
```
pitchinvasion: This isn't the best way of marking up the form - in particular you're lacking labels which help screen readers understand which form input has which label attached to it:

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/How_to_structure_an_HTML_form#The_<label>_element

features/step_definitions/user_send_message_steps.rb
```diff
@@ -0,0 +1,16 @@
+# Given(/^I am on the user page$/) do
```
pitchinvasion: Delete file!

spec/spec_helper.rb
```diff
@@ -0,0 +1,31 @@
+require './app/server'
+require 'capybara/rspec'
+require 'database_cleaner'
```
pitchinvasion: Don't forget to set RACK_ENV to test

pitchinvasion:
Looks good, seems like you were pretty comfortable with adapting bookmark manager, just be careful to keep your code tidy deleting comments and unused files.

pitchinvasion:
:checkered_flag:

PR #11, opened 2015-03-01
Challenge week 4 for review! 

SebastienPires:
Hey guys, here's my fully functional code for this weekend challenge. Thanks in advance for your feedback! Here's the link to the app: [http://awesome-chitter.herokuapp.com/](http://awesome-chitter.herokuapp.com/).


MihaiLiviuCojocar:
1. There is no README.md file. Make sure you always have this file.

Rakefile
```diff
@@ -0,0 +1,12 @@
...
+
+task :auto_upgrade do
+  DataMapper.auto_upgrade!
+  puts "Auto-upgrade complete (no data loss)"
+end
+
+task :auto_migrate do
+  DataMapper.auto_migrate!
+  puts "Auto-migrate complete (data could have been lost)"
+end
```
MihaiLiviuCojocar: Rake tasks do not work because the models are not required.

MihaiLiviuCojocar: Now the migration will work. It doesn't work for the test database though, so you can do it like this: rake auto_migrate RACK_ENV=test.

app/helpers/sessions.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def sign_in(email, password)
+    visit '/sessions/new'
+    fill_in 'email', :with => email
+    fill_in 'password', :with => password
+    click_button 'Sign in'
+  end
+
+
+end
```
MihaiLiviuCojocar: These methods are helper methods for the test files and they should be in the spec folder.

app/views/layout.erb
```diff
@@ -0,0 +1,78 @@
...
+    <br>
+    <br>
+    <br>
+       <% if flash[:notice] %>
+        <div id="notice">
+       <%= flash[:notice] %>
+        </div>
+      <% end %>
+</body>
+</html>
```
MihaiLiviuCojocar: So many divs ... :(

spec/features/adding_tweet_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+end
+
+
+  def add_tweet(message, user)
+    within('#new-tweet') do
+      fill_in 'message', :with => message
+      click_button 'Send tweet'
+    end
+  end
```
MihaiLiviuCojocar: This method(add_tweet) is a helper method and it should be moved with the rest of the helper methods.

app/controllers/sessions.rb
```diff
@@ -0,0 +1,21 @@
...
+end
+
+post '/sessions' do
+  email, password = params[:email], params[:password]
+  user = User.authenticate(email, password)
+  if user
+    session[:user_id] = user.id
+    redirect to('/')
+  else
+    flash[:errors] = ["The email or password is incorrect"]
```
MihaiLiviuCojocar: The way it is now, flash[:errors] persists for two requests. So, if I try to login with wrong credentials an error will be displayed and when I use the right credentials the second time I will logon successfully but I will see the error from before. To fix this, we can write flash.now[:errors] = ... and now, the flash message will persist only for one request.

app/controllers/users.rb
```diff
@@ -0,0 +1,19 @@
...
+                     :name => params[:name],
+                     :username => params[:username])
+    if @user.save
+      session[:user_id] = @user.id
+      redirect to('/')
+    else
+      flash.now[:errors] = @user.errors.full_messages
+      erb :"users/new"
+    end
+end
```
MihaiLiviuCojocar: In post '/users' we don't need to create @user, just user is enough because we are not using this user object in the view, so if we don't use it we don't want to send unnecessary data to the view.

MihaiLiviuCojocar:
Over all is good, I'm ok with it but like everything else, there is room for improvement. :)

:checkered_flag:

PR #13, opened 2015-03-02
Chitter Challenge

ErikAGriffin:
Did not get a chance to really work on the challenge this weekend unfortunately.  I will try to knock out as much as I can in the morning but in its present state there's not much to comment on.


spec/features/lists_all_peeps_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'spec_helper'
+
+feature "User sees the list of peeps" do
+
+  before(:each) do
+    Peep.create(content: 'Peep Peep', creation_date: Time.now)
+  end
+
+  scenario "When visiting the home page" do
```
stephenlloyd: this name isn't very expressive. Maybe you should say 'see the peeps on the homepage'

app/model/chatter_user.rb
```diff
@@ -0,0 +1,15 @@
+class ChatterUser
+
+  include DataMapper::Resource
+
+
+  property :id, Serial
+  property :name, String
+  property :creation_date, DateTime
+  #has n, :peeps
```
stephenlloyd: you started with the associations but didn't complete this. Why not?

app/server.rb
```diff
@@ -0,0 +1,41 @@
...
+require_relative 'custom-hobbit'
+require 'data_mapper'
+
+class Server < Hobbit::Base
+  include Hobbit::Mote
+  include Hobbit::Session
+
+
+
+  # # # # # # # # # # # # # # # #
```
stephenlloyd: nice try but this would have been so much nicer and more explicit just to save it in a descriptive file name and required it.

app/views/index.mote
```diff
@@ -0,0 +1,5 @@
+<ul>
+    % peeps.each do |peep|
```
stephenlloyd: totally awesome job on using mote. Haml is also really good. Impressed.

app/server.rb
```diff
@@ -0,0 +1,41 @@
+require 'hobbit'
```
stephenlloyd: really glad you've done this. Using hobbit over sinatra is a great way to up you web dsl's

app/views/layout.mote
```diff
@@ -0,0 +1,15 @@
+<!DOCTYPE html>
+<html>
+    <head></head>
+    <body>
+
+
+
+        {{content}}
```
stephenlloyd: none of this loads up so I can't mark any interactions. The content isn't there and it just shows the dev tools links.

ErikAGriffin: Content is there, just index is a blank file unless you've loaded some peeps into the database.  If you look at the source you'll see that the <ul> tags from index is there

stephenlloyd:
Really good effort on using hobbit and mote however I was unable to run it. I'm sure I could have got to the bottom of it however there was nothing in your readme to help me out. There are two reasons why you should go off piste, To learn something new and to deliver something better. I hope you learnt a lot but unfortunately what you've delivered isn't as advanced as a lot of the other tests I've been looking at. You have only 3 tests and it doesn't run. Don't get me wrong I'm really happy you tried it out but i'll mark this as still in construction until it's working :construction:

PR #14, opened 2015-03-02
Eddy B's attempt at the chitter challenge

eddbrown:


app/models/user.rb
```diff
@@ -0,0 +1,10 @@
+class User
+
+  include DataMapper::Resource
+  property :id,         Serial 
+  property :username,   String
+  property :password,   String
+
+  has n, :peeps, :through => Resource
```
stephenlloyd: good job on the associations :)

app/server.rb
```diff
@@ -0,0 +1,43 @@
...
+DataMapper.auto_upgrade!
+
+
+
+class Chitter < Sinatra::Base
+
+  enable :sessions
+
+  get '/' do 
+    Peep.create(:message => "Hello there!")
```
stephenlloyd: You shouldn't be creating this here. If you wanted to have a default tweet you should create some seeded data using a seed.rb file which you can run a take task to populate.

app/server.rb
```diff
@@ -0,0 +1,43 @@
...
+    @peeps = Peep.all
```
stephenlloyd: Could you have made so that it orders them for you by date. By default it comes out in order of creation but this may not always be chronological.

app/server.rb
```diff
@@ -0,0 +1,43 @@
...
+    erb :index
+  end
+
+  get '/users/new' do
+    erb :new_user
+  end
+
+  post '/users/welcome' do
+    @username = params[:username]
+    @password = params[:password]
```
stephenlloyd: do these need to be instance variables. Will you be using the @password in the view?

spec/peep_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      # this creates it in the database, so it's stored on the disk
+      Peep.create(message: 'Hello!')
+      # We ask the database how many peeps we have, it should be 1
+      expect(Peep.count).to eq(1)
+      # Let's get the first (and only) peep from the database
+      peep = Peep.first
+      # Now it has all properties that it was saved with.
+      expect(peep.message).to eq('Hello!')
+      
+      # If we want to, we can destroy it
```
stephenlloyd: horrible comments. Just looks like mess.

Gemfile
```diff
@@ -0,0 +1,11 @@
+gem 'data_mapper'
+gem 'dm-postgres-adapter'
+gem 'capybara'
+gem 'sinatra'
+gem 'database_cleaner'
+gem 'rack_up'
```
stephenlloyd: this breaks bundle.

app/views/index.erb
```diff
@@ -0,0 +1,7 @@
+<h1>Chitter!</h1>
+
+<% if !@peeps.empty? %>
```
stephenlloyd: I was expecting to see something which would let a user create a peep. As it stands they can just sign up.

stephenlloyd:
A lot of what you've done is good and clean however there is no way to create a peep. This is the purpose of the website and if you were to create an MVP having peeps without users would probably be your skateboard. It's a shame as it was looking clean and minimalist. If you make a few changes which I've commented on and give it a front end so a user of the website can create a peep then you can have the sticker. Also a readme is always required, We had a bit of trouble  getting it up and running and it's usually good practice to create a Readme with installation and running guides. 
:construction:

PR #15, opened 2015-03-02
Challenge submission

sphaughton:
Really struggled to make progress beyond a very early point with this due to an error I struggled with all of Sunday. I'm getting ```Failure/Error: fill_in :name, :with => name, Capybara::ElementNotFound: Unable to find field :name``` as an error for my ``` 1) User signs up when being a new user visiting the site``` rspec test, and cannot see my fault with any of the specified line errors.


Rakefile
```diff
@@ -0,0 +1,12 @@
+require 'data_mapper'
+require 'data_mapper_setup'
+
```
MihaiLiviuCojocar: The rake task isn't working. There is a load error. It can be fixed by changing require 'data_mapper_setup' to require './app/data_mapper_setup'.

app/data_mapper_setup.rb
```diff
@@ -0,0 +1,10 @@
+env = ENV['RACK_ENV'] || "development"
+
+DataMapper.setup(:default, "postgres://localhost/bookmark_manager_#{env}")
```
MihaiLiviuCojocar: Connect to bookmark_manager_... database? :)

app/data_mapper_setup.rb
```diff
@@ -0,0 +1,10 @@
...
+
+require_relative './lib/tweet'
+require_relative './lib/user'
+
+DataMapper.finalize
+DataMapper.auto_upgrade!
+DataMapper.auto_migrate!
```
MihaiLiviuCojocar: we don't want to do the upgrade or the migrate in this file but in the Rakefile.

app/helpers/application.rb
```diff
@@ -0,0 +1,25 @@
...
+    expect(page.status_code).to eq(200)
+    fill_in :name, :with => name
+    fill_in :username, :with => username
+    fill_in :email, :with => email
+    fill_in :password, :with => password
+    fill_in :password_confirmation, :with => password_confirmation
+    click_button "Sign up"
+  end
+
+end
```
MihaiLiviuCojocar: We have here two types of helper methods: application helpers and test helpers and we shouldn't put them together... sign_in() and sign_up() should be moved into a different file in the features folder.

app/server.rb
```diff
@@ -0,0 +1,71 @@
...
+require 'rack-flash'
+require 'sinatra'
+require 'sinatra/partial'
+
+require './app/lib/tweet'
+require './app/lib/user'
+require_relative 'helpers/application'
+require_relative 'data_mapper_setup'
+
+env = ENV['RACK_ENV'] || 'development'
```
MihaiLiviuCojocar: Setting up env it's already done in data_mapper_setup so we are repeating ourselves here.

app/server.rb
```diff
@@ -0,0 +1,71 @@
...
+end
+
+post '/sessions' do 
+  email, password = params[:email], params[:password]
+  user = User.authenticate(email, password)
+  if user
+    session[:user_id] = user.id
+    redirect to('/')
+  else
+    flash[:errors] = ["The email or password is incorrect"]
```
MihaiLiviuCojocar: In this case, flash[:errors] will persist for two requests. If we use flash.now[:errors] it will persist just for one request.

app/server.rb
```diff
@@ -0,0 +1,71 @@
...
+    erb :"sessions/new"
+  end
+end
+
+delete '/sessions' do 
+  session.clear
+  redirect to('/')
+  erb :index
```
MihaiLiviuCojocar: Rendering a view after we do a redirect is redundant because it will not be rendered anyway.

MihaiLiviuCojocar:
The way it is now, I can add tweets even if I'm not signed in. :)

MihaiLiviuCojocar:
There is still a bit work to do. :)

:construction:`

app/data_mapper_setup.rb
sphaughton: this makes sense and having trouble with an error since changing this and making an association between ```user.rb``` and ```tweet.rb``` that reads ``` undefined method `save_parents' for "Sean":String ``` on ```tweet_spec.rb:9```, i don't really understand what this is?

screen shot 2015-03-02 at 17 21 59

PR #16, opened 2015-03-02
Chitter submission

tomcoakes:
Took me a while to figure out how to set up a Heroku database but just about managed it!

My app is live here: http://chitterapp-tomcoakes.herokuapp.com/

There's a hell of a lot of refactoring I want to do, especially on my cucumber tests - and I'm not sure if I was supposed to unit test my model files. Can't really figure out how to, so any advice would be appreciated. Cheers!


MihaiLiviuCojocar:
You used cucumber... Awesome! :)

app/controllers/users_controller.rb
```diff
@@ -0,0 +1,46 @@
...
+      user = User.new(name: params[:name], username: params[:username], email: params[:email], password: params[:password])
+      if user.save
+        session[:user_id] = user.id
+        redirect to '/'
+      else
+        flash.now[:errors] = user.errors.full_messages
+        erb :"users/sign_up"
+      end
+    end
+  end
```
MihaiLiviuCojocar: This controller looks a bit scary! :) I see that you do a validation for password. If you want to do it yourself it should be done in the model, the way data mapper does it. So, data mapper already does this for you... all you need to add in the model is 'validates_confirmation_of: ... '

app/data_mapper_setup.rb
```diff
@@ -0,0 +1,6 @@
+  env = ENV['RACK_ENV'] || 'development'
+  DataMapper.setup(:default, ENV['DATABASE_URL'] || "postgres://localhost/chitter_#{env}")
+  require './app/models/user'
+  require './app/models/cheep'
+  DataMapper.finalize
+  DataMapper.auto_upgrade!
```
MihaiLiviuCojocar: Every time the app is racked up there is an upgrade on the database happening and it shouldn't. A rake task can fix this issue and you can run it only when you need it.

app/views/layout.erb
```diff
@@ -0,0 +1,45 @@
...
+  <section>
+    <%= yield %>
+  </section>
+
+  <footer>
+    <p><a href="https://github.com/tomcoakes/chitter-challenge">Check out the source code on my GitHub page</a></p>
+  </footer>
+
+</body>
+</html>
```
MihaiLiviuCojocar: Why divs? :)

MihaiLiviuCojocar:
Overall, it's quite good! :) Awesome! :)
:checkered_flag:

app/views/layout.erb
```diff
@@ -0,0 +1,45 @@
...
+<body>
+
+  <header>
+      <div id="header_content">
+        <a href="/"><img src="/images/chitter_logo.png" height="60px" alt="chitter logo"></a>
+
+        <h1 id="title">Chitter</h1>
+        
+        <div id="sign_in_links">
+          <% if current_user %>
```
pitchinvasion: This should be an if/else statement

app/views/users/sign_up.erb
```diff
@@ -0,0 +1,21 @@
...
+    <% flash[:errors].each do |error| %>
+      <li id="error"><%= error %></li>
+    <% end %>
+  </ul>
+<% end %>
+
+<h1>Enter your details to create an account.</h1>
+
+<form action="/users/sign_up" method="post" class="pure-form">
+  <input type="text" name="name" placeholder="Name"><br/>
```
pitchinvasion: Don't use `<br>` tags 

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/How_to_structure_an_HTML_form

features/signing_in_and_out.feature
```diff
@@ -0,0 +1,31 @@
...
+
+  Scenario: Signing in with incorrect details
+    Given I am on the sign in page
+    And I enter "incorrect@email.com" into the "email" field
+    And I enter "wrongpassword" into the "password" field
+    And I press "Sign In"
+    Then I should see "Those login details are incorrect."
+    And I should see "Please Sign In"
+
+#  Scenario: Resetting a password
```
pitchinvasion: Comments!

PR #17, opened 2015-03-02
C Christenson Chitter-Challenge

c-christenson:


app/views/layout.erb
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+
+<html>
+<head>
+  <title>Parliament</title>
+  <link rel="stylesheet" type="text/css" href="/css/application.css">  
+</head>
+<body>
+  <%= partial :welcome_message %>
+ <%= partial :flash %>
```
danldb: indentation

app/server.rb
```diff
@@ -0,0 +1,26 @@
...
+require_relative 'controllers/hoots'
+require_relative 'controllers/application'
+
+
+
+  enable :sessions
+  set :session_secret, 'super secret'
+  use Rack::Flash
+  set :partial_template_engine, :erb
+
```
danldb: You should set your root and your public folders - that will ensure you can locate your images and css.

http://www.sinatrarb.com/intro.html#Static%20Files

danldb:
:checkered_flag: 

Hi Carrie - thanks for your submission!

Great work - I had to actually struggle to find anything to criticise. You have mentioned that you could not render your images and css, for which I have suggested further reading of the documentation. 

Reasonable test coverage - consider testing your database with unit tests rather than feature tests - this would separate your concerns so that the feature tests focus purely on user experience (which does not typically include database counts, but might include seeing 2 tweets instead of 1, for example).

I think you have already mentioned all missing features and problems in your readme, which brings me to another point - the readme should contain details of how to install and run the application, as well as an overview of its functionality, screenshots, learning objectives and so on.

Well done - let us know if you have any questions.

PR #18, opened 2015-03-02
Julian Veling Chitter Challenge

veliancreate:
I spent most of the time trying to understand the relationship between the technologies. Because I was so unsure, I didn't focus on the testing side of it because I experimented alot with what was going on behind the scenes.


app/controllers/peeps.rb
```diff
@@ -0,0 +1,8 @@
+
+post '/peeps' do
```
pitchinvasion: I like you split this up into controllers - nice way of keeping your code tidy!

app/server.rb
```diff
@@ -0,0 +1,25 @@
+require 'sinatra'
+require 'thin'
```
pitchinvasion: This is awesome I love how tidy this file is

app/views/index.erb
```diff
@@ -0,0 +1,20 @@
...
+      <input type="submit" value="Peep!">
+    </form>
+  </div>
+<% end %>  
+
+<h2>Peep feed!</h2>
+
+<ul>
+    <% if @peeps %>
+      <% @peeps.reverse.each do |peep| %>
```
pitchinvasion: You've sneaked a bit of logic into your views here - you could instead have encapsulated this inside a class method inside of your peep model (where the logic belongs), maybe having a method like `Peep.all_for_feed` or something

app/views/layout.erb
```diff
@@ -0,0 +1,73 @@
...
+  </head>
+  <body class="clearfix container">
+    
+    <header class="clearfix">
+      <div class="row">  
+        <h1 class="col-sm-12"><a href="/">ChitterBug!</a></h1>
+      </div>
+      <div class="row">
+        <nav class="col-sm-12">
+          <% if !current_user %>
```
pitchinvasion: Instead of this you could do 

```ruby
if current_user
  ...
else
 ...
end
```

app/views/sessions/new.erb
```diff
@@ -0,0 +1,9 @@
+<h2>Please sign in.</h2>
+
+<form method="post" action="/sessions">
+  <h3>Email:</h3>
```
pitchinvasion: `<h3>` is not the correct semantic tag here, you want to use `<label>`

https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Forms/How_to_structure_an_HTML_form

public/css/grid.css
```diff
@@ -0,0 +1,667 @@
+.container {
```
pitchinvasion: Nice to see you using a grid system here!

pitchinvasion:
This looks good but I can't consider this finished without any tests - a lot of what we are looking for in these challenges is proof that you can understand how to code but equally how to write well-tested code. I understand you might not have felt you had time to add tests but I would really urge you while on the course to focus on testing any code you build! :construction:

PR #21, opened 2015-03-02
Chitter challenge

jjnewman:


Rakefile
```diff
@@ -0,0 +1,12 @@
...
+
+# task :auto_upgrade do
+#   DataMapper.auto_upgrade!
+#   puts "Auto-upgrade complete (no data loss)"
+# end
+
+# task :auto_migrate do
+#   DataMapper.auto_migrate!
+#   puts "Auto-migrate complete (data could have been lost)"
+# end
```
spike01: Try not to leave commented code lying around - people coming to your code later on (or you) won't know whether it's dead code to be deleted or something important

app/controllers/peeps.rb
```diff
@@ -0,0 +1,9 @@
+get '/peeps/new' do
+  erb :"peeps/new"
+end
+
+post '/peeps' do
+  content = params["content"]
```
spike01: You could probably in-line this variable into the line below

spike01: Also try to be consistent with how you access your hashes - in other controllers you're accessing with symbols, and here with a string. Both are fine, but when both being used it makes me wonder if there's something weird going on with params, which there isn't

app/controllers/users.rb
```diff
@@ -0,0 +1,20 @@
...
+  erb :"users/new"
+end
+
+post '/users' do
+  @user = User.new(:first_name => params[:first_name],
+                    :last_name => params[:last_name],
+                    :username => params[:username],
+                    :email => params[:email],
+                    :password => params[:password],
+                    :password_confirmation => params[:password_confirmation])
```
spike01: Could this extraction be dealt with in the model? i.e. pass params directly through and create a tested method in the model

app/helpers/application.rb
```diff
@@ -0,0 +1,5 @@
+helpers do
+  def current_user
+    @current_user ||=User.get(session[:user_id]) if session[:user_id]
```
spike01: Needs a space after `||=`

app/server.rb
```diff
@@ -0,0 +1,29 @@
...
+set :session_secret, 'top secret'
+set :partial_template_engine, :erb
+use Rack::Flash
+
+
+
+
+
+
+
```
spike01: Try to get rid of pointless whitespace

features/step_definitions/web_steps.rb
```diff
@@ -0,0 +1,219 @@
...
+  if actual_params.respond_to? :should
+    actual_params.should == expected_params
+  else
+    assert_equal expected_params, actual_params
+  end
+end
+
+Then /^show me the page$/ do
+  save_and_open_page
+end
```
spike01: You have websteps, but no Cucumber tests - if you aren't using it, don't include it

spec/features/adding_peeps_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+                :password_confirmation => 'test')
+
+    scenario "when browsing the homepage" do
+      expect(Peep.count).to eq(0)
+      visit 'peeps/new'
+      add_peep("Bob")
+      expect(Peep.count).to eq(1)
+      peep = Peep.first
+      expect(peep.content).to eq("Bob")
+    end
```
spike01: You're testing at the wrong level of abstraction here - feature tests should simulate user interaction with your app

spec/spec_helper.rb
```diff
@@ -0,0 +1,29 @@
+ENV['RACK_ENV'] = 'development'
```
spike01: Wouldn't `RACK_ENV` here be test?

app/views/peep.erb
```diff
@@ -0,0 +1,9 @@
+<li class='peep'>
+  <div class='content'>
+    <%= peep.content %>
+  </div>
+  <span class='source'>
+    peeped at <%= peep.created_at.to_datetime.strftime("%l:%M%p %A %e %B") %> by <%= User.get(peep.user_id).username %>
```
spike01: You should never be calling your models directly from the view! Also, multiply chained methods like this are a definite code smell - check out the Law of Demeter

app/views/index.erb
```diff
@@ -0,0 +1,8 @@
+<div id='peeps_container'>
+  <ul id='peeps'>
+    <% @peeps.sort.reverse.each do |peep| %>
+      <%= partial :peep, :locals => {:peep => peep}  %>
```
spike01: Good use of partials

app/views/layout.erb
```diff
@@ -0,0 +1,23 @@
+<!DOCTYPE html>
+<html>
+<head>
+  <title>Chitter</title>
+  <link rel="stylesheet" type="text/css" href="/css/reset.css">
```
spike01: Good use of a CSS reset

spec/features/adding_peeps_spec.rb
spike01: Also, avoid multiple expectations in tests

spec/features/listing_all_peeps_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+    Peep.create(:content => "awful weather")
+  
+    scenario "when opening the home page" do
+      visit '/'
+      expect(page).to have_content("hello")
+      expect(page).to have_content("goodbye")
+      expect(page).to have_content("have a nice day")
+      expect(page).to have_content("awful weather")
+    end
+  end
```
spike01: These tests won't run - they're in the before loop

spec/features/adding_peeps_spec.rb
spike01: And, these tests won't run as they're in the before loop

spike01:
A good attempt, but only 6 tests ran when I ran rspec. Even with the non running tests, you haven't covered the features specified in the challenge - for example, checking the ordering of the peeps. Be careful with generated code (in this case, the Cucumber websteps) - if you're not using it, delete it. :construction:

PR #22, opened 2015-03-02
Basic version of the famous Chitter is ready to deploy (-;

TStrothjohann:
Urgent needs:
- further refactoring
- layout


README.md
```diff
@@ -1,29 +1,24 @@
...
-* In order to avoid others using my account as a maker I want to log out
-* In order to let people know what I am doing as a maker I want to post a message (peep) to chitter
+* In order to use chitter as a maker I want to sign up to the service | done
+* In order to use chitter as a maker I want to log in | done
+* In order to avoid others using my account as a maker I want to log out | done
+* In order to let people know what I am doing as a maker I want to post a message (peep) to chitter | done
 * In order to see what people have to say as a maker I want to see all peeps in chronological order
 
+
+ToDo>
```
pitchinvasion: It's a minor thing but make sure your README is presentable as it may be read by employers - even little things like capitalisation look sloppy if you leave it out

app/models/user.rb
```diff
@@ -0,0 +1,40 @@
...
+    @password = password
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(args)
+    email = args[:email]
+    username = args[:username]
+    password = args[:password]
+
+   user = first(:username => username)
```
pitchinvasion: indentation is out here. Also you can extract some of this logic into private methods, as your method here is very long. For example this could be extracted into a method called `find_by_email_or_username(args)`

app/server.rb
```diff
@@ -0,0 +1,103 @@
...
+
+
+  class Chitter < Sinatra::Base
+    enable :sessions
+    use Rack::Flash
+    use Rack::MethodOverride
+    register Sinatra::Partial
+    set :partial_template_engine, :erb
+   
+
```
pitchinvasion: you don't need all this extra spacing

app/server.rb
```diff
@@ -0,0 +1,103 @@
...
+    
+  get '/users/new' do
+    erb :'users/new'
+  end
+
+  get '/sessions/new' do
+    erb :'sessions/new'
+  end
+
+  post '/users/new' do
```
pitchinvasion: often with a large server file like this we extract these routes into controller files to keep things a bit tidier here

app/server.rb
```diff
@@ -0,0 +1,103 @@
...
+      erb :index
+    else
+      flash[:errors] = ["ups"]
+      redirect('/sessions/new')
+    end
+  end
+
+  delete '/sessions' do
+    session.clear
+    flash[:notice] = "bis bald!"
```
pitchinvasion: You're mixing English and German here! Pretty minor thing, but if you're interested what you really want to be doing is internationalisation https://github.com/sinefunc/sinatra-i18n

app/server.rb
```diff
@@ -0,0 +1,103 @@
...
+    message.save
+    erb :index
+  end
+
+
+  helpers do
+    def current_user
+      if session[:user_id]
+        @current_user = User.get(session[:user_id]).username
+
```
pitchinvasion: Again minor but make sure your spacing is correct you don't need the extra line here

app/views/sessions/new.erb
```diff
@@ -0,0 +1,11 @@
+<form action="/sessions" method="post">
+  
+  <fieldset>
+    <legend>Log in with password and username or email</legend>
+    <label>username or email</label>
```
pitchinvasion: Good using labels, but if your label element does not wrap around your input element then you need to use the `for` attribute so that screen readers know which input field your form refers to (this helps people with accessibility issues

app/views/users/new.erb
```diff
@@ -0,0 +1,26 @@
+<form action="/users/new" method="post" id="signup-form">
+  <fieldset>
+    <legend>Please sign up to use chitter</legend>
+    <p>
+      <label>Name</label>
+      <input type="text" name="name">
+    </p>
+    <p>
+      <label>email</label>
```
pitchinvasion: Capitalisation!

app/views/users/new.erb
```diff
@@ -0,0 +1,26 @@
+<form action="/users/new" method="post" id="signup-form">
+  <fieldset>
+    <legend>Please sign up to use chitter</legend>
+    <p>
```
pitchinvasion: Nice use of `<p>` tags to break up each element, this is what I would consider the best way of marking up a form :)

capybara-201503020021381859313317.html
```diff
@@ -0,0 +1,45 @@
...
+  
+    <form id="sign-out" method="post" action="/sessions">
+      <input type="hidden" name="_method" value="delete">
+      <input class="btn" type="submit" value="Sign out">
+    </form>
+    <a class="btn" href="/messages/new">Create message</a>
+  
+
+  
+    <h3>Hey TStrothjohann </h3>
```
pitchinvasion: This is not dynamic, what if I sign up?

capybara-201503020021381859313317.html
```diff
@@ -0,0 +1,45 @@
...
+  
+</header>
+      
+
+
+
+
+    <form action="/messages/new" method="post" id="create-message">
+  <fieldset>
+    <legend>Write a message</legend>
```
pitchinvasion: Indentation all over the place here, it matters in HTML too!

capybara-20150302002203934073388.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
+<html lang="en">
```
pitchinvasion: I've just realised you've checked in your capybara files. I've left my comments above even though they don't apply just to show why we really, really encourage you to read over your pull request before you leave it for us to review, as we then end up having to go over code which isn't actually relevant to the challenge

Gemfile
```diff
@@ -0,0 +1,14 @@
...
+gem 'bcryp'
+gem 'dm-validations'
+gem 'rack-flash3'
+
+group :test do
+  gem 'rspec'
+  gem 'capybara'
+  gem 'capybara/cucumber'
+  gem 'database_cleaner'
+end
```
pitchinvasion: You need to check in your gemfile.lock http://viget.com/extend/bundler-best-practices

pitchinvasion:
Although I know you said it needs refactoring this is a good start a lot of the right elements are in there. Just be careful you look over the code after you've created the pull request! :checkered_flag:

PR #23, opened 2015-03-02
Monday morning pull request

wardymate:
1 test fails - need to rewrite the test as the functionality has changed. - test now passes
need to write tests for the display of the time and user.
need to work on adding CSS and links to sign in and signup on the home page. -started


pitchinvasion:
I'm still getting failing tests

```
ERROR:  null value in column "user_id" violates not-null constraint
```

app/controllers/application.rb
```diff
@@ -0,0 +1,4 @@
+get '/' do
```
pitchinvasion: Nice work splitting into controllers

app/controllers/peeps.rb
```diff
@@ -0,0 +1,7 @@
+post '/peeps' do
+  content = params["content"]
+  author = params["author"]
+  datetime = DateTime.now
```
pitchinvasion: You can do this automatically in datamapper http://datamapper.org/docs/dm_more/timestamps.html

app/models/peep.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  include DataMapper::Resource
+
+  property :id, Serial
+  property :content, Text
+  property :author, Text
+  property :created_at, DateTime
+
+  def display_date_and_time
+    text = self.created_at.to_s
```
pitchinvasion: The fact that the method name here has "display" in it is indicative of a code smell - I would argue that the model's job is not to handle how data is displayed, just to return it from the database. Instead this could be a view helper

app/views/index.erb
```diff
@@ -0,0 +1,21 @@
...
+    <% end %>
+  </ul>
+ </section>
+
+<% if current_user %>
+  <section id="new-peep">
+   <form action="/peeps" method="post">
+    <fieldset>
+      <legend>Please submit a new peep</legend>
+      <p><label>Message: </label><textarea rows="3" name="content"></textarea></p>
```
pitchinvasion: Don't forget that a label needs a for tag if it doesn't wrap around an input or screen-readers won't know which input the label applies to

spec/features/listing_all_peep_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'spec_helper'
+
+feature "user browses the list of peeps" do
+
+  before do
+    Peep.create(:content => "This is the first chitter peep",
+                :created_at => DateTime.now)
+  end
+
+  scenario " when opening the home page" do
```
pitchinvasion: You haven't done a test for the reverse-ordering of tweets...

pitchinvasion:
Apart from failing tests and missing the ordering of tweets this is really good, well-tested and structured. You just need to make sure you are following the briefs (your tests can help if you write the tests based on your briefts). But nice work (once you've fixed the test!) :checkered_flag:

app/models/peep.rb
```diff
@@ -0,0 +1,17 @@
+class Peep
+
```
pitchinvasion: There needs to be an association with users here, you're adding the username, but what if the user changes their username?

PR #26, opened 2015-03-02
chitter challenge - it was done last night as well

RizAli:


lib/server.rb
```diff
@@ -0,0 +1,43 @@
+require 'data_mapper'
+require 'sinatra'
+require_relative 'user'
+require_relative 'peep'
+require './lib/helpers/application'
+
+env = ENV['RACK_ENV'] || 'development'
+
+# DataMapper::Logger.new($stdout, :debug)
```
tansaku: avoid leaving comments in like this - commit them and then you'll have them in your github history - you don't want dead code in things you submit to github

lib/server.rb
```diff
@@ -0,0 +1,43 @@
...
+DataMapper.setup(:default, "postgres://localhost/chitter_challenge_#{env}")
+
+require './lib/user'
+DataMapper.finalize
+# DataMapper.auto_migrate!
```
tansaku: avoid leaving comments in like this - commit them and then you'll have them in your github history - you don't want dead code in things you submit to github

lib/server.rb
```diff
@@ -0,0 +1,43 @@
...
+DataMapper.auto_upgrade!
+
+enable :sessions
+set :session_secret, 'super secret'
+
+
+get '/' do
```
tansaku: You should prefer to have this inside a class that extends Sinatra::Application

lib/helpers/application.rb
```diff
@@ -0,0 +1,7 @@
+helpers do
```
tansaku: nice to see this in a separate file

Gemfile
```diff
@@ -0,0 +1,16 @@
+source 'https://rubygems.org'
+
+gem 'sinatra'
+gem 'data_mapper'
+gem 'dm-postgres-adapter'
+gem 'capybara'
```
tansaku: should be in test group

lib/helpers/application.rb
```diff
@@ -0,0 +1,7 @@
...
+
+  def current_user
+    @current_user ||=User.get(session[:user_id]) if session[:user_id]
```
tansaku: please ensure space after '||='

tansaku:
All the tests pass which is good. I can post peeps, but I don't see any signin/signup functionality on the main page.  I was able to access that via users/new, but as you probably know you should be updating your layout.erb to include the rest of the necessary functionality.

This is a good start, but you'll need to follow through if you want to get a badge.

:finished:

spike01:
:construction:

PR #35, opened 2015-04-12
Chitter! where all your messages are a secret because they don't get posted with a username ;)

M1lena:
Well I was pretty proud of getting this project set up but then things began to fail at the point where I was trying to implement having the user handle appear with each "tweet". 

Also, I got posting messages constraints to work (to post only if logged in) and it worked on localhost but it broke some other tests.

The code I'm submitting here is when everything was still passing minus the functionality of logging in before posting.


tansaku:
Hi Milena,

Well done getting the basic functionality working.  Very pleased to see sign in and sign out working.  It's great that the tests all pass, although I note that the messages are created but not displayed (view issue) --> so the listing all messages test is not complete.

I'd encourage you focus specifically on getting the features in the README working:

```sh
As a Maker
I want to sign up for Chitter

As a Maker
I want to log in to Chitter

As a Maker
So that I can avoid others posting messages on Chitter as me
I want to log out of Chitter

As a maker

As a maker
```

So for example there's no need for tags in the first pass.  I worry that you are coding ahead of the tests - let me know if that's not the case, but the key going forward is to imagine the test before you imagine the implementation.

A final minor comment is that you have a repeat of the sign_in method in the user_management_spec file.

Overall this is a good start, and looking forward to seeing you complete the specified features.

:construction:

PR #37, opened 2015-04-12
Chitter Challenge Submission

RBGeomaticsRob:


app/models/user.rb
```diff
@@ -0,0 +1,26 @@
+require 'bcrypt'
+class User
+
+  attr_reader :password
+  attr_accessor :password_confirmation
+
+  include DataMapper::Resource
+
+  validates_confirmation_of :password
```
RBGeomaticsRob: Not able to get this to work as hound requests, I think it may be Rails specific I'll keep looking for a Datamapper example

spec/peep_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'spec_helper'
+
+describe Peep do
+  it 'it expects to be created and retrieved from the database' do
+    user = User.create(name: 'Rob',
```
RBGeomaticsRob: Still investigating how to effectively stub this out, the datamapper save_parents method is causing me a problem as I cannot find a way to instigate the relationship using a double, I can respond to the command, but not get the code doing what is should, which is sort of making me think that if I achieved this I would be testing the complexity of the stub rather than the code in peep?

.rubocop.yml
```diff
@@ -67,4 +76,13 @@ Style/Semicolon:
...
+  AllowAsExpressionSeparator: false
+
+Metrics/LineLength:
+  Max: 120
+  # To make it possible to copy or click on URIs in the code, we allow lines
+  # contaning a URI to be longer than Max.
+  AllowURI: true
+  URISchemes:
+    - http
+    - https
```
sjmog: Nice hamstringing :)

README.md
```diff
@@ -32,37 +18,44 @@ So that I can avoid others posting messages on Chitter as me
...
+The app runs on a Postgres database, you will need to create two databases `chitter_test` & `chitter_development`.
 
-* In order to start a conversation as a maker I want to reply to a peep from another maker.
+You will also need to `bundle install` to obtain all the necessary gems.
 
-And/Or:
+##Contribution##
 
-* Work on the css to make it look good (we all like beautiful things).
+This is a challenge set to identify progress in Wk3 at Makers Academy, therefore is designed to be tackled individually.
```
sjmog: Wk5!

Rakefile
```diff
@@ -1,9 +1,22 @@
...
 RuboCop::RakeTask.new :cop
 RSpec::Core::RakeTask.new :spec
 Cucumber::Rake::Task.new :cuke
 
-task default: [:cop, :spec, :cuke]
\ No newline at end of file
+task default: [:cop, :spec, :cuke]
+
+task :auto_upgrade do
+  require './app/data_mapper_setup'
```
sjmog: I'd really like to find a better way to do this - to avoid using `require` within a task

app/chitter.rb
```diff
@@ -0,0 +1,29 @@
+require 'sinatra/base'
+require 'data_mapper'
+require 'rack-flash'
+require 'sinatra/partial'
+require_relative 'data_mapper_setup'
```
sjmog: Why is this here?

app/chitter.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class Chitter < Sinatra::Base
+
+  register Sinatra::Partial
+  set :partial_template_engine, :erb
```
sjmog: V excited to see your usage of Partials later :)

app/chitter.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  helpers do
+    def maker
+      @maker ||= User.get(session[:user_id]) if session[:user_id]
+    end
+  end
+
+  require './app/controllers/index'
+  require './app/controllers/chits'
+  require './app/controllers/users'
```
sjmog: I expect any `require`d modules to come in at the top

app/controllers/chits.rb
```diff
@@ -0,0 +1,27 @@
+class Chitter < Sinatra::Base
```
sjmog: I'd expect my controllers to be classes in their own right -> we'll get to this when we look at Rails.

app/chitter.rb
RBGeomaticsRob: If it is removed the database is not setup and datamapper does not have an 'adapter' to work with?

PR #38, opened 2015-04-12
chittter challenge

rjlynch:


rjlynch:
I added the specific version of do_postgres that travis wants and it's in the gemfilelock but it still thros an error :-/

Rakefile
```diff
@@ -1,9 +1,20 @@
...
+
+task :auto_upgrade do
+  DataMapper.auto_upgrade!
+  puts 'auto_upgrade complete (no data loss)'
+end
+
+task :auto_migrate do
+  DataMapper.auto_migrate!
+  puts 'auto_migrate complete (data could have been lost)'
+end
```
MihaiLiviuCojocar: rake auto_migrate and rake auto_upgrade do not work
try this:
require_relative 'data_mapper_set_up' instead of require 'data_mapper'
and I'll point that changes you need to make in data_mapper_set_up

data_mapper_set_up.rb
```diff
@@ -0,0 +1,10 @@
+env = ENV['RACK_ENV'] || 'development'
```
MihaiLiviuCojocar: require 'data_mapper'

data_mapper_set_up.rb
```diff
@@ -0,0 +1,10 @@
...
+
+DataMapper.setup(:default, "postgres://localhost/chittter_#{env}")
+
+require './lib/user'
+require './lib/peep'
+require './lib/reply'
+
+DataMapper.finalize
+DataMapper.auto_upgrade!
```
MihaiLiviuCojocar: there is no need for 'DataMapper.auto_upgrade!' here

MihaiLiviuCojocar: when you'll do the migrate for the test database you can do this: rake auto_migrate 'RACK_ENV'=test

lib/user.rb
```diff
@@ -0,0 +1,29 @@
...
+    @password = password
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(email, password)
+    user = first(email: email)
+    user if user && BCrypt::Password.new(user.password_digest) == password
+  end
+
+end
```
MihaiLiviuCojocar: the associations between models are not quite right

MihaiLiviuCojocar: let's take the 
user:  he can have many peeps and many replies
peep: can belong only to one user and can have many replies(which can be from different users)
reply: belongs to only one user and only to one reply at the same time
Data mapper provides a way to this(instead of 'though: Resource can be through another model(like 'peep')')

MihaiLiviuCojocar: And on the page would be nice to see something like this:

'Hello!"' peep created by 'Rich'
       "Hello, Richard! I'm Mihai!" reply created by 'Mihai'
       "How are you?"                   reply created by 'Rich'
"It's such a beautiful day today" peep created by 'Mihai'
       "We had better"                  reply created by 'Rich'

........

MihaiLiviuCojocar: Ah, there were some spaces but they were removed ...

controllers/post_reply.rb
```diff
@@ -0,0 +1,12 @@
+class Chittter < Sinatra::Base
+
+  post '/reply' do
+    reply = Reply.create(reply: params[:reply],
+                         user: session[:username])
+    peep = Peep.first(id: session[:reply_id])
+    peep.replys << reply
+    peep.save
```
MihaiLiviuCojocar: if we would have proper associations we could do something like this:
peep.replies.create(....)

MihaiLiviuCojocar:
Overall, it is ok ... ish... :) Awesome!
You can improve it a little bit(especially the associations)

:checkered_flag:

PR #45, opened 2015-04-12
Attempt at the challenge

GBouffard:
Not finished but planning to carry on on Monday and also retry during revision week.


MihaiLiviuCojocar:
:construction:

PR #46, opened 2015-04-12
Pull Request - Basing Chitter no styling (still working on styling in a different fork) 

Nadav-Rosenberg:


app/lib/user.rb
```diff
@@ -0,0 +1,27 @@
...
+    @password = password
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(email, password)
+    user = first(email: email)
+    user if user && BCrypt::Password.new(user.password_digest) == password
+  end
+
+end
```
MihaiLiviuCojocar: I see that there is no association between message and user... there is something but it is done manually, instead of using DataMapper resources.

The message can belong only to one user and the user can have many messages ... this will add that user, and better, the user_id to the message, which will create the association.

MihaiLiviuCojocar:
I see that it is in progress. Try to use the hint that I gave you. :)
:construction:

PR #47, opened 2015-04-13
Chitter challenge - James

james-miller:
Still could do with more refactoring and would like to add some other features but the basic parts are all there. I have managed to add the 'reply' functionality as well as 'hashtag' functionality and it is possible to search by username and hashtag.


MihaiLiviuCojocar:
Awesome job! :)

:checkered_flag:

PR #50, opened 2015-04-13
Week 4 Challenge

kanishkwalia:
Basic version of blabber working on localhost but not deployed to heroku.


MihaiLiviuCojocar:
I cloned it and it looks ok. One thing though... when I ran rspec there were many errors that need to be fixed. Please make sure before submitting that you don't have failing tests! :)

:construction:

PR #51, opened 2015-04-13
chitter-challenge

massud:
Still quite a bit to do especially refactoring my code. Also still a very basic look which i will definitely spend time on improving over the week.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,6 +12,7 @@ group :test do
   gem 'cucumber'
   gem 'rubocop-rspec'
   gem 'rubocop'
+  gem 'rack-flash3'
   gem 'sinatra'
   gem 'coveralls', require: false
 end
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,8 @@ source 'https://rubygems.org'
 ruby '2.2.1'
 
 group :test do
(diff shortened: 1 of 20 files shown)


tansaku:
Hi Massud,

Well done getting the basic functionality working.

There's a few simple things to do to improve:

1) remove the repeated sign up and sign in not necessary in user_management_spec file
2) fix the various rubocop issues
3) in gemfile some prod/dev stuff is in test e.g. sinatra/db --> should go outside test group
4) fix the poor indentation in server.rb --> I believe this is obscuring the error you have in rspec (sign out not working)

I notice a few other things - your 'listing_all_messages' is not a very rigorous test' - it could be passed with a simple hard coding.

Having your models in lib and the server code in app is interesting, although perhaps not so conventional.   Ideally move over completely to app, with the models in app/models

Your peep messages don't display because they are in the url link ...

You've got a good base here, but you should try and get the models linked up (which is in the features) and not worry about password reset (which is not).  Do your best to complete specifically the features in the readme:

```sh
As a Maker
I want to sign up for Chitter

As a Maker
I want to log in to Chitter

As a Maker
So that I can avoid others posting messages on Chitter as me
I want to log out of Chitter

As a maker

As a maker
```

:construction:

PR #56, opened 2015-04-13
Should be worth a sticker

Munded:


Munded:
didn't get comments to belong to user, username is not displayed with comment

tansaku:
Hi Ed,

There is great work here and I really want to give you a badge here, but the main thing holding me back is that you didn't complete implementing the user stories.  You added some other great functionality, but the real key is to only add that extra functionality once you have completed the core functionality.

Here are my detailed notes:

* rubocop clean - awesome!
* tests all pass - awesome!
* 'Convo' not the most self-explanatory model name
* could use more consistent spacing in adding_links_spec  - and place helper methods either at top or bottom (I prefer bottom) - avoid commented out lines and pending specs without explanation of pending status --> dead code
* feature tests not as specific as they might be - could pass with hard coding
* user lost password not part of user stories, so should be added once all user stories are completed
* some styling - awesome
* broken image link on http://localhost:9292/user/new? :-( (absolute vs relative image links)
* no welcome message on sign in (who am I?) - message there but not visible
* not using consistent layout (how to navigate from welcome message to home page?) <-- ask me to show you if this is unclear
* message posting and tags working --> bonus challenge: can you make clicking on tag display all messages that are associated with that tag?
* "sign up now" still available when signed in
* broken image link on password reset page :-( (absolute vs relative image links)
* password reset then takes you to new user page --> ideally should be log in page (or just log you in)
* no visible error message on failed signin
* reset password totally worked - awesome!
* however note that reset password is bonus - you should complete specified features first, i.e. have messages associated with users before doing tags and password reset
* api key in mailgun code --> move to ENV
* repetition of code in server and in controllers --> bad
* inconsistent indenting in css file

I hope that doesn't come across as too negative - you've obviously put a lot of work in and made a lot of progress.  I think it's just critical that we ensure that the work that you put it is focused on completing the project specification first before moving on to other features.  If you can convince me you've understood that I'll give you a badge :-)

Also at a higher level there's a lot of room for improving the consistency of the user experience.  That wasn't a requirement for this project - but since you are clearly starting to manage the application complexity, the next step up is to be considering about how various coding decisions influence the user experience.  If the meaning of that (or any of my comments above) are unclear - let's talk in person to clear it up.  Perhaps we might even do a group session with your app?

:construction:

tansaku:
:checkered_flag:

PR #60, opened 2015-05-22
Chitter challenge

dwatson62:


app/helpers/helpers.rb
```diff
@@ -0,0 +1,42 @@
+helpers do
```
silvabox: Helpers are intended to provide methods to your views, not your controller.  These should all just be methods in `server.rb`

app/helpers/helpers.rb
```diff
@@ -0,0 +1,42 @@
...
+  def prepare_feed
+    @all_peeps = Peep.all
+    @all_peeps.reverse!
+    @all_users = User.all
+  end
+
+  def prepare_public_feed
+    prepare_feed
+    @public_feed = []
+    @all_peeps.each do |peep|
```
silvabox: Avoid loading unnecessary records from the database then filtering them in the model layer.  Can you use DataMapper instead to access only public peeps?

app/helpers/helpers.rb
```diff
@@ -0,0 +1,42 @@
...
+      @public_feed << peep if peep.receiver == 'public'
+    end
+  end
+
+  def prepare_private_feed(sender, receiver)
+    prepare_feed
+    @private_feed = []
+    @sender, @receiver = sender, receiver
+    @all_peeps.each do |peep|
```
silvabox: As above - use DataMapper to restrict the records selected from the database.

app/helpers/helpers.rb
```diff
@@ -0,0 +1,42 @@
...
+        @private_feed << peep
+      end
+    end
+  end
+
+  def create_peep(sender, receiver)
+    message = params[:message]
+    if message && message.length > 140
+      flash[:error] = 'Peeps must be less than 140 characters'
+      redirect to '/main'
```
silvabox: redirects should only happen in the controller.

app/views/index.erb
```diff
@@ -0,0 +1,35 @@
...
+  <br>
+  <a href='/signup'><button type="button" class="btn btn-primary">Sign up</button></a>
+  <br>
+  <br>
+  <h3>Past peeps: </h3>
+  <br>
+  <br>
+  <% if @public_feed %>
+    <% @public_feed.each do |peep| %>
+      <% user = @all_users.get(peep.user_id) %>
```
silvabox: Use DataMapper to handle this logic: `user = peep.user` - that's what associations are for.

app/views/mainpage.erb
```diff
@@ -0,0 +1,41 @@
...
+      <button type="submit" class="btn btn-primary" name='Go'>Go</button>
+    </div>
+  </form>
+  <br>
+  <h2>Past peeps:</h2>
+  <br>
+  <br>
+  <% if @public_feed %>
+    <% @public_feed.each do |peep| %>
+      <% user = @all_users.get(peep.user_id) %>
```
silvabox: `user = peep.user`

app/views/private_peep.erb
```diff
@@ -0,0 +1,25 @@
...
+      <label for="inputPeep">Write a new private peep here: </label>
+      <input type='text' placeholder='Peep here...' name='message'>
+      <input type="hidden" name="receiver" value= <%= @receiver %> >
+      <button type="submit" class="btn btn-primary" name='Peep'>Peep</button>
+    </div>
+  </form>
+  <br>
+  <% if @private_feed %>
+    <% @private_feed.each do |peep| %>
+    <% user = @all_users.get(peep.user_id) %>
```
silvabox: `user = peep.user`

spec/features/user_login_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'spec_helper'
+require 'helpers/user_helpers'
+
+feature 'A user tries to login' do
```
silvabox: I don't like this way of structuring tests.  I can see what you want to do, but for me it makes the tests unreadable in code form.  `A user tries to login` is not a feature.

spec/features/user_login_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  before(:each) do
+    User.create(email: 'test@test.com',
+                password: '123',
+                password_confirmation: '123',
+                name: 'John',
+                username: 'test_user1')
+  end
+
+  scenario 'with correct details and is successful' do
```
silvabox: reading this standalone, it isn't clear what the expectation is.  Something like 'succeeds with correct details' might be better.

spec/features/user_logout_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+                password: '123',
+                password_confirmation: '123',
+                name: 'John',
+                username: 'test_user1')
+    login('test_user1', '123')
+  end
+
+  scenario 'from the main page' do
+    visit '/main'
+    expect(page).to have_content 'Welcome back John'
```
silvabox: Is this expectation part of the behaviour being tested?

spec/features/user_signup_spec.rb
```diff
@@ -0,0 +1,55 @@
+require 'spec_helper'
+require 'helpers/user_helpers'
+
+feature 'A new user wants to sign up' do
+  scenario 'for the first time' do
+    visit '/'
+    expect(User.count).to eq 0
+    expect(page).to have_content 'Welcome to Chitter'
```
silvabox: are these 2 expectations part of the behaviour being tested?

spec/peep_spec.rb
```diff
@@ -0,0 +1,42 @@
+require 'spec_helper'
+require 'helpers/user_helpers'
+
+describe Peep do
+
+  context 'can create a' do
```
silvabox: This is not the recommended way to use contexts.  Context is intended to provide a block in which to place tests with similar conditions, so that they can share `let`s and `before`s

app/views/index.erb
```diff
@@ -0,0 +1,35 @@
+<body>
```
silvabox: Don't include `body` tags in your views when you already have `body` tags in your layout.

app/views/index.erb
```diff
@@ -0,0 +1,35 @@
...
+  <br>
+  <h1>Welcome to Chitter</h1>
+  <br>
+  <br>
```
silvabox: Don't use `<br>` tags for layout.  HTML is for structure; use CSS for layout.

app/views/index.erb
```diff
@@ -0,0 +1,35 @@
...
+  <form class="form-inline" method='post' action='/login'>
+    <legend>Please login here</legend>
+    <div class="form-group">
+      <label for="inputUsername">Username: </label>
```
silvabox: the `for` attribute should be the `id` for the related `input` tag.  In this case, the `input` tag that follows should have `id="inputUsername"`

silvabox:
Well done Daryl, this is a good submission.  There are a number of improvements you can make to the structure of your code and I would definitely recommend changing the way you write your tests.  But it all works and CI is clean - so good job.

:checkered_flag:

PR #61, opened 2015-05-23
Sinatra, rspec/capybara, bootstrap + heroku #april2015

rodcul:
https://chitter-rodcul.herokuapp.com/


ptolemybarnes:
Great challenge submission! A few things:  
- Your tests could be refactored considerably with a gem called FactoryGirl. check it out.
- more to come..

ptolemybarnes:
@rodcul and I had a 1 to 1 on this and went over a few things:
- using FactoryGirl to refactor the setting up of state for tests.
- refactoring test helper methods. For example, being able to pass a model into a "sign_in" helper, rather than just a list of strings, would clean things up a lot.
- Refactoring the create(post) route of controllers with helper methods. What is the best practice for this? I suggested the Rails way of doing it. Have a look here: http://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html
- Other questions I have:
 - What is the "base_url" helper method doing? I am curious.

PR #63, opened 2015-05-25
Chitter-challenge submission [AH] - Minimalitter-chitter

AlexHandy1:


ptolemybarnes:
Nice submission! Great work. Few things:
- Your creation of state for tests is not very DRY. Check out a gem called FactoryGirl which will help with this... pill forthcoming :stuck_out_tongue: 
- Not sure what this test is for:
```ruby
scenario 'and peep can be stored on website' do
  visit ('/')
  @peep = Peep.create(message: "Hello world",
  user_id: @user1.id)
  expect(@peep.message).to eq "Hello world"
end
```
You're basically testing just DataMapper functionality here, which is (hopefully) already tested.  
more to come!

app/controllers/users.rb
```diff
@@ -0,0 +1,18 @@
+get '/users/new' do
+  @user = User.new
+  erb :'users/new'
+end
+
+post '/users' do
+  name, username, email, password, password_confirmation = params[:name], params[:username],params[:email], params[:password], params[:password_confirmation]
+  @user = User.create(name: name, username: username, email: email, password: password, password_confirmation: password_confirmation)
```
ptolemybarnes: Consider passing the params hash itself to User.create, rather than extracting out individual variables. Of course, you then have the problem of there possibly being form data in the params hash that you do not want (after all users can always submit form data that you do not intend them to). The rails solution to this is the following: http://apidock.com/rails/ActionController/Parameters/permit
You could do something similar in sinatra, i.e have a helper method within each controller that filters the params for only the values needed to create the relevant resource.

app/helpers/current_user.rb
```diff
@@ -0,0 +1,5 @@
+helpers do
+  def current_user
+    @current_user ||= User.get(session['user_id']) if session['user_id']
```
ptolemybarnes: I think what you're trying to do here is protect from current_user returning nil values. (is that right?) Unfortunately this method will still return nil when session['user_id'] is empty. Notably your tests still pass when the conditional is deleted.

PR #65, opened 2015-05-25
Chitter Challenge - Basic User Stories met

MollieS:


ptolemybarnes:
Great submission!!! A few things.
- Overall I am really impressed by your cucumber test coverage. You might refactor a little though as you have this sequence of step definitions a few times:
```cucumber
And I fill in "email" with "peter@test"          
And I fill in "name" with "Peter Smith"    
And I fill in "password" with "test"           
And I fill in "password_confirmation" with "test"
And I fill in "username" with "Peter"
```
This is overkill. For "Scenario: with an email already signed up", for example, you might have a single step definition: "And I enter an email address that is already in use". Then within our step definition we could do all the background stuff (filling in "name" etc) without confusing the output from cucumber with steps that are not relevant to the functionality we're trying to test. Does that make sense?

PR #67, opened 2015-05-25
User stories complete

DanielJohnston:
Wrote feature tests for each story, then implemented in modular frontend code with MVC split. No backend code led to no unit tests being created, somewhat unfortunately. That and a bunch of other extra features and proper CSS styling. This pull request is the first integration test, so fingers crossed.


DanielJohnston:
Looks like CI is failing because it's expecting a cucumber features directory? Can't immediately spot how to change that.

spec/features/peeps_management_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+                password_confirmation: 'testpassword')
+  end
+
+  scenario 'and receives confirmation of posting' do
+    sign_in
+    post_peep
+    expect(page).to have_content('Your peep has been posted')
+  end
+
+  scenario 'while not logged in' do
```
ptolemybarnes: Your scenario here doesn't describe the behaviour we would expect when a user posts a peep without being logged in.

spec/features/user_management_spec.rb
```diff
@@ -0,0 +1,74 @@
+require 'spec_helper'
+require_relative 'helpers/session'
+
+include SessionHelpers
+
+feature 'User signs up' do
```
ptolemybarnes: Again, your scenarios here should describe to the reader the behaviour we would expect given certain conditions...
So something like:
- Feature: A user is welcomed to Chitter
  - Scenario: when signed up with correct details
- Feature: A user is prompted to re-enter details
  - Scenario: having submitted non-matching passwords.
  - Scenario: having submitted an email address already in use.  

Not totally convinced that is best practice but I think it's an improvement.

spec/features/user_management_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    User.create(email: 'uhoh@danieljohnston.co.uk',
+                handle: 'uhoh',
+                password: 'testpassword',
+                password_confirmation: 'testpassword')
+  end
+
+  scenario 'when starting at the login page' do
+    visit '/'
+    expect(page).not_to have_content('Welcome, uhoh@danieljohnston.co.uk')
+    sign_in('uhoh@danieljohnston.co.uk', 'testpassword')
```
ptolemybarnes: See my gist on how to refactor this sort of thing: https://gist.github.com/ptolemybarnes/2dfda607b85d01e113b0

spec/features/peeps_management_spec.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  scenario 'and sees their peep on the front page' do
+    sign_in
+    post_peep
+    visit '/'
+    expect(page).to have_content('This is a test peep')
+  end
+end
+
+feature 'User views list of peeps' do
```
ptolemybarnes: I had a go at refactoring this: https://gist.github.com/ptolemybarnes/323b9559049264dc6c2d
Again, not convinced this is the best way of going about testing that peeps appear in the order you want but it's a little cleaner.

ptolemybarnes:
Great submission! 
- You've figured out the modular app layout for sinatra. 
- Your routes are very restful.
:checkered_flag:

app/controllers/users.rb
```diff
@@ -0,0 +1,30 @@
...
+
+    class Users < Base
+
+      get '/users/new' do
+        @user = User.new
+        erb :'users/new'
+      end
+
+      post '/users' do
+        @user = User.new(email: params[:email],
```
ptolemybarnes: Here you could pass the params hash to User.new, rather than extracting out each values. Of course, you then risk users submitting random form data which will cause your app to crash. So you then might have a helper method within each controller that would filter the params hash for only those attributes permitted for the creation of the associated resource (in this case, User). This is something Rails offers: http://edgeapi.rubyonrails.org/classes/ActionController/StrongParameters.html

PR #69, opened 2015-05-26
chitter_challenge

bagolol:


ptolemybarnes:
Great submission! A few things:
 - Watch out for mis-indentation
 - You're including the CurrentUser module in a global scope :-S Switching to a modular style would avoid this problem, or using Sinatra's helpers block to define methods.
- Your tests are pretty good but a few seem to have a couple of lines of code that aren't doing anything or are doing something not all that relevant to the test they are within. They're a little cluttered.
.... more to come!

PR #71, opened 2015-05-26
Chitter Challenge

CharlieKenny:


ptolemybarnes:
Hey, not a bad submission though looks unfinished. Make time for it in lab week maybe? A few things:
- I'm getting a DataObjects::IntegrityError when I run your spec. This is because, according to the Peep model, peeps should be created by users but in your 'post /peeps' route, you're not creating peeps belonging to any user.
- Your route look nice and RESTful. Keep it up!
I'm going to leave this as :construction: . Let me know when you get around to finishing it and we'll review again!

PR #72, opened 2015-05-26
Chitter Challanger

curlygirly:
Please note my database kept dropping and I had to restart whole challenge 3 times over the weekend as my code base broke as the database dropped. 

Yesterday (Monday) evening I had a fully working (but not fully refactored or Css'd ) code which is what has been submitted. 

Unfortunately the database dropped in the evening again whilst I was refactoring and adding CSS - I am not sure if the code will work, but it was working properly when I committed to github.

I would like to finish the challenge properly, as I would like to fully execute the plan I had created.

JAM.


ptolemybarnes:
Sorry you had such a hard time with databases! What you've got so far looks great. Perhaps have a crack at finishing this in lab week? One thing going forward:  
- You needn't have both a features and spec folder. It seems like you've got the directory set up for both Cucumber and RSpec feature testing.

:construction:

PR #73, opened 2015-05-26
user password working

timoxman:


ptolemybarnes:
Looks good, though maybe unfinished? A few things:
- The output from your "registering an account" scenario is too verbose. Consider reducing the 6 "And I enter..." steps down to "And I enter correct details." 
- I'm getting "NameError at /peep/new" when I try to peep. I think this is because in the "/peep/new" route you're calling #save on the Peep class, rather than an instance of the Peep class. =)  
Anyway, try to have another crack at this over Lab Week and we'll review again.

PR #85, opened 2015-07-06
A horrendous attempt at the Chitter challenge by Simon

leggsimon:


silvabox:
Hi Simon, as we discussed, the next step is to write a test for the next feature - signing in or creating a peep.  You can refer to bookmark manager for instructions on how to do the sign in.

:construction:

PR #90, opened 2015-07-06
Challenge completed

natstar93:


spec/features/user_management_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  scenario 'requires an email address to be entered' do
+    user = build :user, email: ''
+    expect { sign_up_as(user) }.not_to change(User, :count)
+    expect(page).to have_content 'Email must not be blank'
+  end
+
+  scenario 'requires a unique email address' do
+    user = build :user
+    dup_user = build :user, password: '1233', password_confirmation: '1233', name: 'Bob', username: 'Bobo'
+    sign_up_as(user)
```
silvabox: You don't need to sign up as the first user - just create them in the database.

spec/features/user_management_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    expect { sign_up_as(dup_user) }.not_to change(User, :count)
+    expect(page).to have_content 'Email is already taken'
+  end
+end
+
+feature 'User sign in' do
+
+  scenario 'with correct credentials' do
+    user = build :user
+    sign_up_as(user)
```
silvabox: You don't need to sign up to test sign in - just create the user.

spec/features/user_management_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+    sign_in(user)
+    expect(page).to have_content 'Hi, Natso'
+  end
+end
+
+feature 'User sign out' do
+  scenario 'while being signed in' do
+    user = build :user
+    sign_up_as(user)
```
silvabox: You don't need to sign up as the user - only sign in.

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,35 @@
+feature 'Viewing peeps' do
+
+  scenario 'can view peeps' do
+    Peep.create(message: 'Hello, world')
```
silvabox: Use FactoryGirl for peeps too!

spec/models/user_spec.rb
```diff
@@ -0,0 +1,15 @@
+describe User do
+
+  let!(:user) do
+    User.create(email: 'natalie@gmail.com', password: '1234', password_confirmation: '1234', name: 'Natalie', username: 'Natso')
```
silvabox: Just use FactoryGirl

views/peeps/peeps.erb
```diff
@@ -0,0 +1,8 @@
+<h1>Peeps</h1>
+<% if @peeps %>
```
silvabox: Don't make the view test the existence of `@peeps` - it should be the controller's responsibility to ensure this instance variable is available.

app/models/peep.rb
```diff
@@ -0,0 +1,19 @@
...
+require 'dm-postgres-adapter'
+
+class Peep
+
+  include DataMapper::Resource
+
+  property :id, Serial
+  property :message, String
+  property :post_time, Text
+  property :user_id, Serial
```
silvabox: Don't define a relationship by just introducing your own foreign key column as it will not create the correct constraints in the database.  Use `belongs_to :user` instead.

app/models/peep.rb
```diff
@@ -0,0 +1,19 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+
+class Peep
+
+  include DataMapper::Resource
+
+  property :id, Serial
+  property :message, String
+  property :post_time, Text
```
silvabox: Use time and/or date columns to store times and dates

app/models/peep.rb
```diff
@@ -0,0 +1,19 @@
...
+  property :user_id, Serial
+  property :username, String
+  property :name, String
+
+  def time=(time)
+    self.post_time = "#{time.hour}:#{time.min} #{time.day}/#{time.month}/#{time.year}"
```
silvabox: Use ruby time formats to create string representations (`.strftime`)

app/models/user.rb
```diff
@@ -0,0 +1,35 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+require 'bcrypt'
+
+class User
```
silvabox: Specify the relationship with peep: `has n, :peeps`

app/models/user.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  attr_reader :password
+  attr_accessor :password_confirmation
+
+  validates_confirmation_of :password
+  validates_presence_of :password, :username, :name, :email
+  validates_uniqueness_of :email, :username
+
+  property :id, Serial
+  property :email, String
```
silvabox: You need to specify `unique: true` here instead of relying on the `validates_uniqueness_of :email`

app/models/user.rb
```diff
@@ -0,0 +1,35 @@
...
+  property :password_digest, Text
+
+  property :name, String
+  property :username, String
```
silvabox: use `required: true` to ensure the `NOT NULL` database constraint is created.

app/models/peep.rb
```diff
@@ -0,0 +1,19 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+
+class Peep
+
+  include DataMapper::Resource
+
+  property :id, Serial
+  property :message, String
```
silvabox: Is `message` a required field?

app/chitter.rb
```diff
@@ -0,0 +1,89 @@
...
+    erb :'sessions/new'
+  end
+
+  get '/peeps/new' do
+    erb :'peeps/new'
+  end
+
+  post '/peeps' do
+    if session[:user_id]
+      if (params[:peep] != '')
```
silvabox: This should be a validation inside the Peep model.

silvabox:
Great work Natalie - this is a nice clean solution.  There are some issues as we discussed, but they are easily fixed.  Would be great to see the comments/reply functionality implemented.

:checkered_flag:

app/models/peep.rb
```diff
@@ -0,0 +1,19 @@
...
+  property :post_time, Text
+  property :user_id, Serial
+  property :username, String
```
silvabox: Don't store `username` in peep!  The username is in the users table and is made accessible to the peep through the association.

PR #125, opened 2015-08-17
Ben Hawker - Chitter

benhawker:


app/chitter.rb
```diff
@@ -0,0 +1,110 @@
...
+    erb :"users/password_reset"
+  end
+
+  post "/users/password_reset" do
+    user = User.first(email: params[:email])
+    user.update(password_token: rand_token)
+    flash[:notice] = 'Check your emails'
+  end
+
+  get "/users/password_reset_reset/:token" do
```
silvabox: This route is not well named and doesn't feel RESTful

app/views/users/password_reset_reset.erb
```diff
@@ -0,0 +1,12 @@
+<link rel="stylesheet" href="../style.css" type="text/css" />
+<link href='http://fonts.googleapis.com/css?family=Open+Sans' rel='stylesheet' type='text/css'>
+
+
+<% if token? %>
```
silvabox: And if not?

app/views/users/password_reset_reset.erb
```diff
@@ -0,0 +1,12 @@
...
+  <p>Almost there. Please enter a new password</p>
+  <form action="/users/password_reset" method="post">
```
silvabox: This form posts to the previous part of the process - where does the password actually get changed?

app/views/layout.erb
```diff
@@ -0,0 +1,46 @@
...
+
+	<%= erb :"partials/_notice" %>
+
+	<%= erb :"partials/_flash" %>
+  </div>
+    <%= yield %>
+  
+  <div id="footer"> Footer
+  <ul class="footer-links">
+  <li>Home</li>
```
silvabox: Should these contain `<a>` tags?

silvabox:
Hi Ben,  thanks for submitting your Chitter challenge.  I had a few problems running it as there are no links on your UI, but your code looks reasonably OK.  Watch out for controller logic in your view layer - for example `if token?`.
:checkered_flag:

PR #126, opened 2015-08-17
lots to do

bgraves14:
Sorry for not getting it complete on time but I felt over the weekend that I needed a review weekend as lots of the core knowledge I needed to do this challenge I didn't have. Not sure if it was due to me missing a few days last week or what but personally I feel im behind and in need of lots of review which I am doing now with all my free time. Hopefully I can come back to it soon and blaze right through it.


silvabox:
Hi Brandon, obviously there's not much to review here!  We'll meet a lot of these concepts again in Rails week, so hopefully you'll get a chance to consolidate.

:construction:

PR #127, opened 2015-08-17
Late request due to major block

7091lapS:
Had a very obscure database persistence problem where one of my validations (password length: 6 characters minum) would for some reason prevent peeps from getting saved in the database. Took me 2 days and Ptolemy's help to figure out the problem. Sorry for the late submission. I'll try and refactor modular in the next days, add timestamp and fix another couple of issues if I have time.


app/chitter.rb
```diff
@@ -0,0 +1,68 @@
...
+  end
+
+  post '/peeps' do
+    user = current_user
+    peep = user.peeps.create(content: params[:new_peep])
+    redirect to('/peeps')
+  end
+
+  get '/users/new' do
+    @user = User.new  # nb: this is to store the values in the form in case of unsuccessful registration
```
silvabox: Note your comment here is incorrect - this `@user` is only used to initialize the form the first time.  It can never include any values as this route is not posted to.

app/chitter.rb
```diff
@@ -0,0 +1,68 @@
...
+    erb :'users/new'
+  end
+
+  post '/users' do
+    user = User.create(username: params[:username], name: params[:name], email: params[:email], password: params[:password], password_confirmation: params[:password_confirmation])
```
silvabox: What happens if the fields are invalid?

spec/features/peeps_db_management_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Users have their own peeps stored in database' do
+  scenario 'a users peep is stored' do
+    user = (create :user, username: 'owner', email: 'a@a.com')
```
silvabox: incorrect use of parentheses

spec/features/peeps_db_management_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+    sign_in(user)
+    visit '/peeps'
+    fill_in 'new_peep', with: 'posted this'
+    click_on 'Peep'
+    expect(User.get(user.id).peeps.map(&:content).sample).to eq('posted this')
```
silvabox: why `.sample`???

spec/features/user_db_management_spec.rb
```diff
@@ -0,0 +1,37 @@
+#require 'byebug'
+feature 'User sign up' do
+
+  scenario 'needs a valid email' do
```
silvabox: think about testing the user interface - should I see an error message?

silvabox:
Hi Antonio, thanks for submitting your Chitter challenge.  There's a bit still to do and a few minor issues to address, but it's a good start.

:construction:

PR #140, opened 2015-09-26
Adrian's Chitter Challenge. 

Adrian1707:


sjmog:
Nice challenge, Adrian. Overall I'm happy that you understand the concepts from this week and I get the feeling you are eager to play with more.

### Do this again

- Good focus on security, and use of `session_secret`.
- Nice skinny controllers. Challenge yourself to modularise them further!
- Nice use of a CSS reset
- Very thorough specs, especially clear feature specs. This is wonderful.
- Great use of FactoryGirl for your factories. Bear in mind FactoryGirl can handle associations pretty easily, too.

### Work on this

- Try to use a 'cautious save' pattern. That is, use a conditional to save things instead of `create`:
```ruby
@foo = Foo.new(foo_params) # i.e. not Foo.create(foo_params)
if @foo.save
  flash[:notice] = "All good"
  redirect '/'
else
  flash[:errors] = @foo.errors # this is a handy pattern
  redirect '/foo/new'
end
```
- Check your understanding of `flash` vs `flash.now`.
- Try to use `post`-`redirect` loops: redirecting to a different `get` request at the end of a `post` action instead of rendering a view right in the `post` action (cf. `app.rb:63`)
- `Peep` should `belong_to` user, not `has 1`. And it definitely should go `through` anything (that is for 'many-to-many' relationships, `Peep`s can only have one `User`).
- `User` should `has n` `Peep`s, no `through` table required.
- Use early returns instead of explicitly conditionally returning `nil`:
```ruby
if something
  return :foo
end
return
```

### Weird things

- Why did you include `Rack::MethodOverride`?

All in all, a great pass. Nice job, dude! :checkered_flag:

PR #142, opened 2015-09-26
Tim Hyson's Chitter Challenge

timhyson:


christopheralcock:
looks good, but problems with factory girl implementation that I couldn't solve

stephenlloyd:
:checkered_flag:

PR #157, opened 2015-09-27
Joe Zhou Chitter Challenge

chn-challenger:


app/controllers/peep_controller.rb
```diff
@@ -0,0 +1,23 @@
...
+  module Routes
+    class PeepController < Base
+      get '/peeps/new' do
+        erb :'peeps/new'
+      end
+
+      post '/peeps' do
+        time = Time.new
+        peep = Peep.new(content: params[:content],
+        creation_time: "Peeped on #{time.day}-#{time.month}-#{time.year}")
```
sjmog: indent this

sjmog:
Great challenge, Joe. You currently have a failing test so I cannot pass this. Just fix that and you're golden!

### Do this again

- Lovely custom `.rspec`.
- Nice to see `pry` and `byebug` in the test environment!
- Highly modular MVC structure. This is a pain in Sinatra, nice job. You'll enjoy Rails ;)
- Security: good use of `session_secret`
- Great use of `post` `redirect` loops.
- I like your use of `session` to store the `current_peep`, and your extraction of it to a helper. For what you are doing, you could also use an `<input type="hidden" name="peep_id" value="<%= @peep.id %>">`.
- Nice model associations
- Great use of partials for more readable and versatile views.
- Nice to see FactoryGirl in action!
- Generally, great specs. Just fix up one failing one (a path somewhere).

### Avoid this in future

- Very basic README. I have no idea how to set this up or get started. Include those things in future.
- You construct `Peep`s and `Reply`s in several parts (e.g. `peep_controller.rb:10`). Do this all at once.
- You're not always using a 'cautious save' pattern. That is, use a conditional to save things instead of `create`:
```ruby
@foo = Foo.new(foo_params) # i.e. not Foo.create(foo_params)
if @foo.save
  flash[:notice] = "All good"
  redirect '/'
else
  flash[:errors] = @foo.errors # this is a handy pattern
  redirect '/foo/new'
end
```
- Just store the `Time` object in the Database, don't convert it to a string first. If you want to display times in future it'll be a lot easier to work with the raw `Time` object than anything else (you can interpolate it in the view).
- `reply_spec.rb` is horrible, it's all one test jam-packed full of multiple expectations. Refactor it into separate expectation clauses, one per feature.

### Weird things

- You use `Routes` as a namespace, when (judging from the naming) they're all `Controllers`. Why not use `Controllers` as the namespace instead?
- You've created a `BaseController` but you don't use it. Deliberately?

Just get that spec fixed up, and any of the above points, and this baby'll turn into a checkered flag.

:construction:

chn-challenger:
:)  BaseController is a class which every other controller inherits from, for refactor purpose.

PR #158, opened 2015-09-27
Harry Venables Chitter Challenge

hvenables:


app/chitter.rb
```diff
@@ -0,0 +1,105 @@
...
+                      reply: params[:reply],
+                      created_at: Time.now)
+    reply.peep = peep
+    reply.user = current_user
+    reply.save
+    redirect to('/peeps')
+  end
+
+  # start the server if ruby file executed directly
+  run! if app_file == $0
```
ljones140: I recommend modularise your controller. A controller for each concern

app/chitter.rb
```diff
@@ -0,0 +1,105 @@
...
+    @user = User.new
+    erb :'users/new'
+  end
+
+  post '/users' do
+    @user = User.new(name: params[:name],
+                     username: params[:username],
+                     email: params[:email],
+                     password: params[:password],
+                     password_confirmation: params[:password_confirmation])
```
ljones140: There is a way to tidy this up with a permitted params constant and some methods to check the params against them

app/chitter.rb
```diff
@@ -0,0 +1,105 @@
...
+    @user = User.new
+    erb :'sessions/new'
+  end
+
+  post '/sessions/new' do
+    user = User.authenticate(params[:email], params[:password])
+    if user
+      session[:user_id] = user.id
+      session[:username] = user.username
+      redirect to('/')
```
ljones140: Do you need to put both username and user id in the session. Can you look the username up from the id?

app/chitter.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  get '/peeps/new' do
+    @peeps = Peep.all
+    erb :'peeps/new'
+  end
+
+  get '/peeps/:id/reply' do
+    session[:id] = params[:id]
+    erb :'reply/new'
+  end
```
ljones140: needs a bit of a tweak to work. Try creating the reply associated with the peep

app/helper/current_user.rb
```diff
@@ -0,0 +1,5 @@
+module Helpers
+  def current_user
+    @current ||= User.first(id: session[:user_id])
+  end
```
ljones140: good method

app/views/layout.erb
```diff
@@ -0,0 +1,28 @@
...
+    <link rel="stylesheet" type="text/css" href="/style.css">
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap.min.css">
+    <link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/css/bootstrap-theme.min.css">
+
+    <script src="https://maxcdn.bootstrapcdn.com/bootstrap/3.3.5/js/bootstrap.min.js"></script>
+
+  </head>
+  <body>
+
+    <%= partial :'partials/signout' %>
```
ljones140: well done on using partials

app/chitter.rb
```diff
@@ -0,0 +1,105 @@
...
+
+  delete '/sessions' do
+    session[:user_id] = nil
+    flash.now[:notice] = "Goodbye!"
+    erb :'sessions/goodbye'
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
```
ljones140: two issues here. 1 you are getting all the peeps out of the database then asking ruby to reverse them in your view. Second the database cannot be guaranteed to give you the data in the order you put it in.

You need to explicitly tell the database how you want the data ordered when you query ot
 @peeps = Peep.all(order: [:time.desc])

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,30 @@
+feature 'Creating Peeps' do
+
+  let(:user) do
+    User.create(name: 'test',
+                username: 'test',
+                email: 'test@test.com',
+                password: '12345',
+                password_confirmation: '12345')
```
ljones140: you are using Factory Girl so don't need to add the user fields. They are in the factory already. This just overwrites them

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    Timecop.freeze(Time.local(2015))
+  end
+
+  after do
+    Timecop.return
+  end
+
+  scenario 'I can create a new peep' do
+    sign_in(email: 'test@test.com', password: '12345')
+    visit '/peeps/new'
```
ljones140: I suggest you use a sign_in_as(user) method and pass in a factory

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+                username: 'test',
+                email: 'test@test.com',
+                password: '12345',
+                password_confirmation: '12345')
+  end
+
+  before do
+    Timecop.freeze(Time.local(2015))
+  end
+
```
ljones140: good use of time cop

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    peep.user = @user
+    peep.save
+    visit '/peeps'
+    expect(page.status_code).to eq(200)
+    within 'div#peep' do
+      expect(page).to have_content('testing testing 123')
+      expect(page).to have_content('01/01/2015 00:00')
+    end
+  end
+end
```
ljones140: Cannot see a test for order of peeps. I recommend orderly gem so you can use appears_before assertion

stephenlloyd:
:checkered_flag:

Issue #170, opened 2015-10-30
Create challenge review rubric


silvabox:

tansaku:
in progress here: https://github.com/makersacademy/course-design/blob/master/challenge-feedback/chitter-challenge/feedback.md

tansaku:
can't seem to attach the pull request: https://github.com/makersacademy/chitter-challenge/pull/174

tansaku:
another related pull request that I can't seem to attach https://github.com/makersacademy/course/pull/450

tansaku:
a separate section for more opinion based stuff?

PR #174, opened 2015-11-05
WIP towards completing code review rubric

tansaku:
1st draft completed.  I'm struggling with this (distracted by ronin and js week), and would love input from those working on bookmark manager related stuff and others with strong opinions :-)

Trying to wrap up #170

### DIFF
--- a/docs/review.md
+++ b/docs/review.md
@@ -35,11 +35,10 @@ Particularly now that we have a database involved, it becomes even more importan
 ```
 $ git clone https://github.com/tansaku/chitter_challenge
 $ bundle
-$ psql < CREATE DATABASE chitter_development
+$ echo 'CREATE DATABASE chitter_development' | psql
 $ rake auto_migrate
 $ rspec
 $ rackup
-
 ```
 
 And as we saw in previous weeks you'll want to have screenshots of how the app works, or perhaps even a link to the deployed version on heroku?
@@ -57,13 +56,26 @@ You will need to host your images somewhere, e.g.:
 
 ## Read Over Your Pull Request Before Submitting
 
-The final process should be reading the created pull request and making changes that are seen before sending it to us, as otherwise we waste time going over indentation, dead code in comments, and unnecessary files ++++
+Always leave a space to check over your pull request before submission.  Please try to check for simple issues like:
 
-For example if you've been using the launchy gem to `save_and_open_page` then you'll have a load of `capybara-<TIMESTAMP>.html` files in your root directory that you don't want committed to git.  Try updating .gitignore like so:
+* indentation
+* dead code in comments
+* unecessary files
+
+For example if you've been using the [launchy gem](https://github.com/copiousfreetime/launchy) to `save_and_open_page` then you'll have a load of `capybara-<TIMESTAMP>.html` files in your root directory that you don't want committed to git.  Try updating `.gitignore` like so:
+
 
 ```
 capybara-*.html
 ```
+
+and if you already have unwanted files committed to git then delete them from git:
+
+```
+$ git rm capybara-*.html
+```
+
+then commit and push
 
 ## Ensure Rakefile has appropriate tasks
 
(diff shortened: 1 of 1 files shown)


tansaku:
closes #170

tansaku:
fixes #170

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+# Step 2: Structure and supporting files
+
+## Installation Instructions in README
+
+Particularly now that we have a database involved, it becomes even more important to ensure that correct installation instructions are included in your readme so that other developers (and yourself in the future) know how to get set up with the application:
+
+```
+$ git clone https://github.com/tansaku/chitter_challenge
+$ bundle
+$ echo 'CREATE DATABASE chitter_development' | psql
```
jafrog: Shouldn't it be

```shell-script
  echo 'CREATE DATABASE chitter_development | psql postgres
```
?

I think on Unix `psql`'s default DB name == username so if you didn't configure it beforehand it will throw an error

tansaku: works on osx ... maybe your suggestion is better ...

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+Use [factory girl](https://github.com/thoughtbot/factory_girl) appropriately and configure correctly
+
+```ruby
+# spec/support/factory_girl.rb
+RSpec.configure do |config|
+  config.include FactoryGirl::Syntax::Methods
+end
+```
+
+TODO - do we want factory girl used? Is it optional?  Only for acceptance tests?  To be used with care in case it creates unrealistic acceptance tests?
```
pitchinvasion: We haven't introduced Factory Girl yet

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+
+But I realise that this might not be popular - but I think it's an important part of following the acceptance test, unit test cycle ...
+
+Maybe this should be part of moving to something like:
+
+* https://github.com/jeremyevans/sequel
+* https://github.com/kematzy/minitest-sequel
+* https://github.com/openhood/rspec_sequel_matchers
+
+I guess we can't have any of the above when the bookmark_manager project doesn't address any of this ...
```
pitchinvasion: I agree I would love to see our models unit tested (as well as some more unit-tested service objects) but for the moment this is inconsistent with the current curriculum so remove this for the moment

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+```
+
+See https://github.com/makersacademy/course/blob/master/pills/spec_helper_methods.md
+
+
+## Ensure asset routes are set correctly
+
+Do we like http://recipes.sinatrarb.com/p/asset_management/sinatra_assetpack ?
+
+Or should we just be telling them to set the public folder correctly
```
pitchinvasion: Just set the public folder correctly for the moment, again we'd need to add this into the curriculum first otherwise

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+
+# Step 4: Application code and \*.rb files
+
+## Controllers
+
+### Prefer Sinatra::Application vs Sinatra::Base
+
+TODO work out if we care
+
+* http://www.sinatrarb.com/intro.html#Modular%20vs.%20Classic%20Style
```
pitchinvasion: I don't care, but we do modular in bookmark manager

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+    redirect('/password_reset')
+  end
+end
+```
+
+It could be refactored like so:
+
+```ruby
+post '/password_reset' do
+  if User.reset_password(email: params[:email])
```
pitchinvasion: I'm a big fan of this, although is this done in the curriculum? Could still be something to help them out

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+```
+
+TODO - do we need the `use` keyword here?
+TODO - watch out for unnecessary inheritance - does PeepController need to extend anything?
+
+## Models
+
+### Not encapsulating business logic in the model
+
+see 'refactor long controller methods' above where business logic is pulled from controller into User model
```
pitchinvasion: Again, love this but do we do this in the curriculum?

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+
+## View Helpers
+
+Avoid the following in your app file:
+
+```ruby
+class Chitter < Sinatra::Base
+  helpers do
+    def current_user
+      User.get(session[:user_id])
```
pitchinvasion: I think we do

```
@user ||=  User.get(session[:user_id])
```

docs/review.md
sdorunga1: or just `createdb chitter_development`

docs/review.md
sdorunga1: I think I agree.

docs/review.md
```diff
@@ -0,0 +1,579 @@
...
+
+    get '/peeps/new' do
+      erb :'peeps/new'
+    end
+  end
+end
+```
+
+TODO - do we need the `use` keyword here?
+TODO - watch out for unnecessary inheritance - does PeepController need to extend anything?
```
sdorunga1: how do you mean?

docs/review.md
tansaku: right, but what about Application vs Base?  I'm not how much difference it makes - I read the following:

```
The Application class should not be subclassed, unless you want to inherit all settings, routes, handlers, and error pages from the top-level. Subclassing Sinatra::Base is highly recommended for modular applications.
```
in http://www.rubydoc.info/github/sinatra/sinatra/Sinatra/Application 

I could dig deeper, but does anyone have any experience about the different between Application vs Base - or it makes little difference and the students can mix and match?

I think we all agree to push modular, but modular extensions of Sinatra::Base, or modular extensions of Sinatra::Application?

docs/review.md
tansaku: I think this is the kind of thing to encourage either way, so leaning to keep in this, at least as a suggestion for good practice

docs/review.md
tansaku: we do do the breaking up like this, but in the curriculum we just monkey patched the Sinatra::Base class rather than using this `use` method.  I saw some students using this in their challenges last time - not clear if it is "better practice" or not ...

docs/review.md
tansaku: we do don't we  - will update

docs/review.md
tansaku: I mean we could do:

```ruby
module Routes
  class PeepController < Sinatra::Application
  end
end
```

or 

```ruby
module Routes
  class PeepController < Sinatra::Base
  end
end
```

or 

```ruby
module Routes
  class PeepController
  end
end
```

and do we need it wrapped in a module?

and how do we pull the routes in:

```ruby
class Chitter < Sinatra::Application
  use Routes::PeepController
end
```

or 

```ruby
class Chitter < Sinatra::Application
  use PeepController
end
```

or 

```ruby
class Chitter < Sinatra::Application
end
require 'peep_controller'  # <-- what we did when you and I were pairing ...
```

or mix and match?

pitchinvasion:
:shipit: if stefan is happy with your comments on his thing

PR #179, opened 2015-11-07
Tom's Chitter Challenge

forty9er:


tansaku:
You had your chitter_challenge reviewed by **dan**.
### The good points are:
* Gemfile uses test groups
* Asset routes set correctly
* Used modular style
* Routes split over separate controller files
* View Helpers used appropriately

* Use Spec 'Helpers' Appropriately
* Correct Semantics for Form MarkUp

app/views/index.erb
```diff
@@ -0,0 +1,21 @@
...
+<body>
+  <div id='peeps'>
+  <br>
+  <br>
+  <form action='/peeps/new' method='get'>
+    <input type='submit' class='peep_button' value='NEW PEEP!'>
+  </form>
+  <br>
+  <br>
+  <% Peep.all(order: [:id.desc]).each do |peep| %>
```
danldb: Direct communication between the view and the model - this should always go via the controller to ensure correct encapsulation. I would expect to see `@peeps.each` here instead of `Peep.all`.

spec/features/logging_in_spec.rb
```diff
@@ -0,0 +1,17 @@
+require_relative '../web_helpers'
+
+feature 'Logging in:' do
+  include Helpers
+
+  scenario 'A user can log in' do
+    sign_up
+    log_in
+    expect(page).to have_content 'Hi, Yev Insomniak'
```
danldb: The string 'Hi, Yev Insomniak' is a little bit mysterious. You can make it more obvious what you are testing if your sign_in method contains an argument 'Yev Insomniak'. That way, I could instantly see that the important thing here is that the name is being stored

spec/features/signing_up_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+  scenario "User can't sign up without a valid email address" do
+    visit '/users/new'
+    expect(page.status_code).to eq(200)
+    fill_in :name, with: 'Yev Insomniak'
+    fill_in :username, with: 'yev_insomniak'
+    fill_in :email, with: 'incredibleyev.com'
+    fill_in :password, with: 'bla-bla-bla'
+    fill_in :password_confirmation, with: 'bla-bla-bla'
+    click_button 'Sign up!'
+    expect(page).to have_content("Doesn't look like an email address to me")
```
danldb: If you were using factorygirl this would be a lot simpler and less brittle - `fill_in :name, with: user.name` would mean your tests are much more future proof.

spec/peep_spec.rb
```diff
@@ -0,0 +1,9 @@
+require_relative '../app/models/peep'
+
+describe Peep do 
+  subject(:peep) { described_class }
+  let(:message) { 'check out my new band: http://theyev.com/' }
+  let(:name) { 'Yev Insomniak' }
+  let(:username) { 'yev_insomniak' }
+  
+end
```
danldb: I don't understand what you are doing here, this code seems vacuous.

spec/web_helpers.rb
```diff
@@ -0,0 +1,41 @@
...
+    fill_in :password, with: 'the-yev'
+    click_button 'Log in'
+  end
+
+  def peep
+    visit '/peeps/new'
+    fill_in :peep, with: 'My test peep'
+    click_button 'PEEP!'
+  end
+
```
danldb: These methods could all take arguments that would later make your tests more readable.

danldb:
:checkered_flag: 

Well done, really strong submission. The biggest error was calling your model directly from the view - never, ever do this. The controller is there to provide an interface layer between your view and your model.

PR #182, opened 2015-11-08
Chitter challenge

blancopado:


tansaku:
You had your chitter_challenge reviewed by **Norm Fasey**.
### The good points are:
* Gemfile uses test groups
* Forms use correct markup

* Ensure Rakefile has Appropriate Tasks
* Ensure Separate DataMapper setup file
* Prefer Modular over Classic Style
* Use View Helpers

app/app.rb
```diff
@@ -0,0 +1,91 @@
...
+                  name: params[:name],
+                  nickname: params[:nickname],
+  								password: params[:password],
+  								password_confirmation: params[:password_confirmation])
+  	if @user.save # #save returns true/false depending on whether the model is successfully saved to the database.
+	    session[:user_id] = @user.user_id
+	    redirect to('/peep')
+
+  	else
+			flash.now[:errors] = @user.errors.full_messages
```
danldb: indentation! very important to get into the habit of proper indentation - It would be a real shame to fail a tech test because your code is poorly indented.

app/app.rb
```diff
@@ -0,0 +1,91 @@
...
+class Chitter < Sinatra::Base
+
+	register Sinatra::Flash
+	enable :sessions
+	set :session_secret, 'super secret'
+
+	use Rack::MethodOverride
+  
+  get '/' do
+  	redirect '/sign_up'
```
danldb: what if the user is already signed in?

app/models/peep.rb
```diff
@@ -0,0 +1,15 @@
...
+class Peep
+
+  include DataMapper::Resource
+
+  property :id,     Serial
+  property :text,  String
+
+end
+
+DataMapper.setup(:default, ENV['DATABASE_URL'] || "postgres://localhost/chitter_#{ENV['RACK_ENV']}")
```
danldb: these should not go here - you could extract the datamapper setup stuff into a separate file and then require it in the main server file.

app/app.rb
```diff
@@ -0,0 +1,91 @@
...
+	    session[:user_id] = @user.user_id
+	    redirect to('/peep')
+
+  	else
+			flash.now[:errors] = @user.errors.full_messages   		
+			erb :'users/sign_up'
+		end
+  end
+
+  get '/peep' do
```
danldb: your paths should all be plural i.e. `get '/peeps' do `

app/app.rb
```diff
@@ -0,0 +1,91 @@
...
+	enable :sessions
+	set :session_secret, 'super secret'
+
+	use Rack::MethodOverride
+  
+  get '/' do
+  	redirect '/sign_up'
+  end
+
+  get '/sign_up' do
```
danldb: this is not restful - should be `/users/new`

danldb:
:construction: 
Good submission. Please fix the indentation - it was actually really hard to review this as I couldn't read the code properly. To be honest, if this was a real tech test, I would not progress the application to interview, purely for this reason.

PR #183, opened 2015-11-08
Rob Youngman's changes so far

youngmanr:


tansaku:
You had your chitter_challenge reviewed by **jamie**.
### The good points are:
* Gemfile uses test groups
* Forms use correct markup

* Ensure Correct Location of Feature/Acceptance and Unit Tests
* Prefer Modular over Classic Style
* Use View Helpers

spec/features/user_log_in_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+                password: 'secret1234',
+                password_confirmation: 'secret1234',
+                name: 'user',
+                user_name: 'user1')
+  end
+
+  scenario 'with correct credentials' do
+    sign_in(email: user.email, password: user.password)
+    expect(page).to have_content "Welcome, #{user.email}"
+  end
```
danldb: good sign_in method - the argument makes it really clear that you should be seeing the users email.

app/app.rb
```diff
@@ -0,0 +1,87 @@
...
+      erb :'peeps/new'
+    else
+      flash.next[:notice] = 'User not logged in'
+      redirect '/peeps'
+    end
+  end
+
+  post '/peeps' do
+    Peep.create(message: params[:message], name: current_user.name,
+      user_name: current_user.user_name, time_stamp: Time.now)
```
danldb: what happens if the user later changes their username?

danldb:
:checkered_flag: 

Really good submission, I don't have a lot to add. Good separation between view / model / controller. If I was being really picky I could mention that the controllers could be refactored into separate files, but the app is really small so even that might be overkill.. Well done!

app/models/user.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.authenticate(email, password)
+    user = first(email: email)
+    if user && BCrypt::Password.new(user.password_digest) == password
+      user
+    else
+      nil
+    end
+  end
+end
```
danldb: no relationship between user and peeps? How would you get all the peeps that belong to a user?

PR #190, opened 2015-11-08
Andrew Dowell (Ronin)

k0zakinio:


tansaku:
You had your chitter_challenge reviewed by **Tom Barnes**.
### The good points are:
* Used modular style
* Forms use correct markup

* Gemfile should use Test Groups
* Set up Database Cleaner Correctly
* Use View Helpers

app/models/peep.rb
```diff
@@ -0,0 +1,12 @@
+class Peep
+  include DataMapper::Resource
+
+  property :id,         Serial
+  property :content,    String,
+           length:      250,
+           required:    true
+  property :created_at, DateTime
+
+  has 1, :user, through: Resource
```
danldb: I would say a Peep *belongs_to* a User, whilst the user *has n* Peeps.

app/views/index.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Most recent peeps</h1>
+
+<% if peeps %>
+  <% peeps.each do |peep| %>
+    <div class='peeps'><p><%= peep.content %></p>
+    <%= peep.created_at.strftime("Posted on %d/%m/%Y, at %l:%M%p") %> by <%= peep.user.username %></div>
```
danldb: There is a lot of logic here that should not really be in the view. There are several options for extracting this away - I would lean towards a helper method that takes an argument. You could extract this right back to the model, but as this is very much display logic, its probably better to keep this close to the front end.

spec/features/peeps_spec.rb
```diff
@@ -0,0 +1,25 @@
+feature 'Peeps' do
+  scenario 'be able to send peeps' do
+    register
+    sign_in
+    peep
+    expect(page).to have_content("Hello world")
```
danldb: "Hello world" is a magic string - why are you expecting the page to have "hello world"? If your helpers accepted arguments, this would be more readable. For instance `peep("Hello World")` would make it really clear that your expectation is looking for the content of the Peep.

spec/features/sign_in_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'Sign in' do
+  scenario 'signing in' do
+    register
+    sign_in
+    expect(status_code).to eq 200
+    expect(page).to have_content('Welcome, andy')
+  end
+end
```
danldb: What happens if you can't sign in?

spec/web_helpers.rb
```diff
@@ -0,0 +1,28 @@
...
+  visit '/peeps/new'
+  fill_in(:new_peep, with: 'Hello world')
+  click_button('Submit')
+end
+
+def peep2
+  visit '/peeps/new'
+  fill_in(:new_peep, with: 'Does this work?')
+  click_button('Submit')
+end
```
danldb: These methods should accept arguments. That would also remove the duplication of peep vs peep2.

danldb:
:checkered_flag: 

Well don Andy!

PR #194, opened 2015-11-08
octavianrotari

OctavianRotari:


tansaku:
You had your chitter_challenge reviewed by **David Bebb**.
### The good points are:
* Gemfile uses test groups
* Used modular style
* Forms use correct markup

* Unnecessary Components should be Removed
* Use View Helpers

PR #242, opened 2016-02-13
Built initial file structure, gemfile and spec_helper

Itsindigo:


mechacoach:
You had your chitter-challenge reviewed by **Jonathan Gardiner**

### The good points are:

* Gemfile uses test groups
* DataMapper set up in separate configuration file, ensurespechelper.rbisconfiguredcorrectly: spec_helper.rb pulls in entire app through single point, Only test helpers (not view helpers) are pulled in to spec_helper.rb
* Asset routes set correctly
* Used modular style
* Forms use correct markup

* Ensure spec_helper.rb is Configured Correctly
* Use View Helpers


see https://github.com/makersacademy/chitter-challenge/blob/master/docs/review.md for more details

silvabox:
Hi Matt, obviously you've still got quite a bit to complete, but what you've done so far looks good.

If you get time to come back it, think about introducing a layout file.  Also think about your route names:  `post '/signupinfo'` could just be `post '/signup'` (or `sign_up`) - the `info` is superfluous and it doesn't match the assoicated `get` route.

PR #252, opened 2016-02-13
1st user story

ScottGledhill:


mechacoach:
You had your chitter-challenge reviewed by **Gareth McKenna**

### The good points are:

* All unnecessary components removed, Ruby style guide conventions followed, useflash.nowandflash.nextcorrectly: flash.now and flash.next used correctly
* Gemfile uses test groups
* DataMapper set up in separate configuration file, ensurespechelper.rbisconfiguredcorrectly: spec_helper.rb pulls in entire app through single point, Only test helpers (not view helpers) are pulled in to spec_helper.rb
* Used modular style
* Forms use correct markup
* View Helpers used appropriately

* Ensure spec_helper.rb is Configured Correctly
* Avoid RSpec Feature Scenarios organized like Unit Tests

### Additional comments:
Very good overall, could do with username being applied to each account instead of being its own parameter. 

see https://github.com/makersacademy/chitter-challenge/blob/master/docs/review.md for more details

PR #255, opened 2016-02-14
Initialised Sinatra and DataMapper,

marion-lv:
created first test (user count changes when new user signs up)


silvabox:
Hi Marion, this is a reasonable start and what you've done so far looks OK.  A few general comments:
- move your `models` folder to sit under `app`
- move your `data_mapper_setup` file out of `models`
- move you CSS into a separate file.  Avoid inline CSS wherever possible.

PR #264, opened 2016-02-14
Davids pull request

drjparry:


app/app.rb
```diff
@@ -0,0 +1,84 @@
+ENV["RACK_ENV"] ||= "development"
+
+require 'sinatra/base'
+require 'sinatra/flash'
+require_relative 'models/user'
+require_relative 'models/post'
+require_relative 'data_mapper_setup'
+
+
+class Chitter < Sinatra::Base
```
katie210: Good refactor step would be to split each function into separate controller files:
https://github.com/makersacademy/chitter-challenge/blob/master/docs/review.md#split-routes-into-separate-controller-files

app/views/sign_up.erb
```diff
@@ -0,0 +1,29 @@
+<% if flash[:errors] && !flash[:errors].empty? %>
```
katie210: Good thing to look at: 
https://github.com/makersacademy/chitter-challenge/blob/master/docs/review.md#use-partials-with-standard-html-conventions

app/views/layout.erb
```diff
@@ -0,0 +1,19 @@
...
+    <title>Chitter</title>
+  </head>
+
+  <body>
+    <h1> Chitter </h1>
+    <% if @current_user %>
+      Welcome, <%= @current_user.name %>
+    <% end %>
+     <% if flash[:notice] %>
+    <div id='notice'>
```
katie210: Prefer semantic divs rather than <div>
https://github.com/makersacademy/chitter-challenge/blob/master/docs/review.md#use-partials-with-standard-html-conventions

PR #319, opened 2016-05-06
Update README.md

llcclarke:


silvabox:
Think about REST in your routes (e.g. `post '/usercreate'` should be `post '/users'`.  And also how to relate peeps and users using associations.

PR #326, opened 2016-05-07
Initial commit

sitypop:


silvabox:
Nice work Sity!  Check your `User` class as you have some `DataMapper` cruft at the end which should be deleted.

PR #328, opened 2016-05-07
Chitter Challenge

OMGDuke:


silvabox:
Good work overall.  I had problems running your rake tasks as you've defaulted your environment to `production` instead of `development`.  Also, your route for creating peeps should post to `/peeps` rather than `/`

PR #338, opened 2016-05-08
HWW ChitterChatter

harrywynnwill:


silvabox:
Your code is pretty good, although you haven't quite followed the specification for user sign in (name and username), which is trivial here but would obviously matter in the real world!

Would be good to get the CI build passing.

PR #339, opened 2016-05-08
First commit

flimflamjim:


silvabox:
It's a reasonable start, but you've got your DataMapper config a bit messed up.  The `Datamapper.setup` and `.finalize` code should only be called once from a separate file, not in each model.  Plus, you need to interpolate the Rack env into the database url otherwise you'll always be connected to test!

PR #467, opened 2016-10-21
Glenn B submission for Chitter Challenge

glynester:


app/models/peep.rb
```diff
@@ -0,0 +1,14 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+
+class Peep
+  include DataMapper::Resource
+
+  has n, :comments, through: Resource
```
silvabox: what type of relationship does this define?  Is it correct?

Isn't there another relationship that `Peep` should have?

app/models/comment.rb
```diff
@@ -0,0 +1,10 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+
+class Comment
```
silvabox: shouldn't `Comment` have relationships?

app/models/user.rb
```diff
@@ -0,0 +1,34 @@
+require 'bcrypt'
+
+class User
```
silvabox: Doesn't `User` have any relationships?

PR #468, opened 2016-10-21
Frances Maxwell Chitter Submission

francesmx:


app/controllers/peeps.rb
```diff
@@ -0,0 +1,22 @@
+class Chitter < Sinatra::Base
+
+  get '/' do
+    @peeps = Peep.all(:order => [:created_at.desc])
+    erb :'index'
+  end
+
+  post '/' do
```
silvabox: this should be `post '/peeps'` - think about REST.

app/models/user.rb
```diff
@@ -0,0 +1,33 @@
...
+require 'securerandom'
+
+class User
+  include DataMapper::Resource
+
+  attr_reader :password
+  attr_accessor :password_confirmation
+  validates_confirmation_of :password
+
+  has n, :peeps, through: Resource
```
silvabox: is this the right relationship?

app/views/index.erb
```diff
@@ -0,0 +1,18 @@
+<% if current_user %>
+  <a href="/peeps/new">Post a peep</a>
+<% else %>
+  <a href="/sessions/new">Log-in</a>
+  <a href="/users/new">Sign-up</a>
+<% end %>
+<h1>Welcome to Chitter</h1>
+<ul>
+  <% @peeps.each do |peep|
+    user = User.get(peep.user_id)
```
silvabox: DataMapper already does this for you; it's just `peep.user`

app/controllers/peeps.rb
```diff
@@ -0,0 +1,22 @@
...
+    peep = Peep.create(peep_text: params[:peep_text], user_id: session[:user_id])
+    if peep.save
+      peep.save
```
silvabox: this is redundant - you've already saved it.

spec/features/posting_peeps_spec.rb
```diff
@@ -0,0 +1,43 @@
+feature 'Posting peeps' do
+
+  let!(:user) do
+    User.create(name: 'Bo Peep',
+                username: 'BoPeep789',
+                email: 'bo@peep.com',
+                password: 'sheepies',
+                password_confirmation: 'sheepies')
+    log_in(email: 'bo@peep.com', password: 'sheepies')
```
silvabox: this line should be in a `before` block.  The `let` statement is intended to define a test object (i.e. `user`).

PR #471, opened 2016-10-22
Making the World Chitter

darrenf54321:


app/models/peep.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Peep
+
+  include DataMapper::Resource
+
+  property :id,       Serial
+  property :title,    String
+  #property :time,     Time
+
+  #belongs_to :user
```
silvabox: why is this commented out?

app/models/user.rb
```diff
@@ -0,0 +1,36 @@
...
+    nil
+    end
+  end
+
+  def password=(password)
+    @password = password
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+#has n, :peeps
```
silvabox: why is this commented out?

PR #474, opened 2016-10-22
Malinna's Chitter

MalinnaLeach:


app.rb
```diff
@@ -0,0 +1,70 @@
...
+  end
+
+  get '/sign-up' do
+    erb :sign_up
+  end
+
+  post '/new-user' do
+    User.new_user_create(params[:name], params[:username], params[:email], params[:password])
+    session["username"] = params[:username] if User.redirect == '/home'
+    redirect User.redirect
```
maryrosecook: `User.redirect` is a cool idea.  It may cause errors if you have multiple people clicking around your site at the same time, because the `redirect` instance variable will be shared between them.

app.rb
```diff
@@ -0,0 +1,70 @@
...
+    redirect '/home'
+  end
+
+  get '/sign-up' do
+    erb :sign_up
+  end
+
+  post '/new-user' do
+    User.new_user_create(params[:name], params[:username], params[:email], params[:password])
+    session["username"] = params[:username] if User.redirect == '/home'
```
maryrosecook: It feels like the code in the User model that sets the value of User.redirect could probably be moved to the controller.  Deciding where to redirect the user is a messaging concern, rather than a data manipulation concern.

app.rb
```diff
@@ -0,0 +1,70 @@
...
+    erb :sorry_sign_up
+  end
+
+  post '/logout' do
+    session["username"] = nil
+    redirect '/home'
+  end
+
+  post '/peep' do
+    Peep.create(user: User.first(username: session["username"]), time: Time.new, text: params[:peep])
```
maryrosecook: Really love these short controller methods.

app.rb
```diff
@@ -0,0 +1,70 @@
...
+    redirect User.redirect
+  end
+
+  get '/login' do
+    erb :login
+  end
+
+  post '/user-login' do
+    username = User.login_check(params[:username], params[:password])
+    session["username"] = username
```
maryrosecook: Could the lines than set session username be refactored to use a helper function eg:

```
def log_in_user(username)
  session["username"] = username
end
```

app.rb
```diff
@@ -0,0 +1,70 @@
...
+
+  get '/re-signup' do
+    erb :re_signup
+  end
+
+  get '/re-login' do
+    erb :re_login
+  end
+
+  get '/psswrd-error' do
```
maryrosecook: Is there a reason password is missing the vowels?

models/datamapper_setup.rb
```diff
@@ -0,0 +1,17 @@
...
+require 'data_mapper'
+require 'dm-postgres-adapter'
+require "dm-validations"
+
+configure :development do
+  DataMapper.setup(:default, "postgres://localhost/chitter_dev")
+end
+
+configure :test do
+  DataMapper.setup(:default, "postgres://localhost/chitter_test")
```
maryrosecook: It would be even cleaner to get the database URLs from environment variables.  eg `ENV["production_database_url"]`.  This would make deploying easier in the future, and allow you to keep the URLs secret (having them public is a mild security leak).

PR #477, opened 2016-10-22
Chitter challenge

groundberry:


app.rb
```diff
@@ -0,0 +1,75 @@
...
+  get '/' do
+    erb :'sessions/new'
+  end
+
+  get '/users/new' do
+    erb :'users/new'
+  end
+
+  post '/users' do
+    user = User.create(
```
maryrosecook: I love this clear formatting of the code!

app.rb
```diff
@@ -0,0 +1,75 @@
...
+      email: params[:email],
+      password: params[:password]
+    )
+    session[:user_id] = user.id
```
maryrosecook: Could the lines than set session user_id be refactored to use a helper function? eg:

```
def log_in_user(user_id)
  session[:user_id] = user_id
end
```

app.rb
```diff
@@ -0,0 +1,75 @@
...
+      session[:user_id] = user.id
+      redirect to('/peeps')
+    else
+      flash.now[:error] = 'The email or password is incorrect'
+      erb :'sessions/new'
+    end
+  end
+
+  delete '/sessions' do
+    session[:user_id] = nil
```
maryrosecook: Could you define a `log_out_user` helper method to clarify this code?

app.rb
```diff
@@ -0,0 +1,75 @@
...
+  end
+
+  helpers do
+    def current_user
+      @current_user ||= User.get(session[:user_id])
+    end
+  end
+
+  run! if app_file == $0
+
```
maryrosecook: Watch out for extra empty lines.

app/models/user.rb
```diff
@@ -0,0 +1,24 @@
...
+
+  property :id, Serial
+  property :email, String
+  property :password_digest, Text
+
+  def password=(password)
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(email, password)
```
maryrosecook: Would it be possible to rewrite this method to use a guard clause that raises an error if the password is wrong? It might make the control flow even clearer.

spec/features/creating_peeps_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    expect(page).to have_content(body2)
+  end
+
+  scenario 'I can see all peeps in reverse chronological order' do
+    sign_in(email: email, password: password)
+    body1 = 'Message 1'
+    post_peep(body: body1)
+    body2 = 'Message 2'
+    post_peep(body: body2)
+    expect(page).to have_selector('ul li:nth-child(1)', text: body2)
```
maryrosecook: To get this test working, try manually setting up the conditions (by actually creating the peeps in the browser) and then using `document.querySelector("ul li:nth-child(1)")` in the browser dev console to see if the element you expect gets selected.  (If I'm not being clear, just grab me and we can talk it through!)

app/views/layout.erb
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="utf-8">
+    <title>Chitter</title>
+  </head>
+  <body>
```
maryrosecook: You asked about getting the CSS working with a div, but I didn't see any CSS in the pull request.  Grab me if you want to chat about it in person!

PR #482, opened 2016-10-23
Chitter Challenge

andygjenkins:


app/controllers/peeps.rb
```diff
@@ -0,0 +1,21 @@
+class Chitter < Sinatra::Base
+
+  get '/peeps/new' do
+    erb :'peeps/new'
+  end
+
+  post '/peeps/new' do
```
silvabox: this should be `post '/peeps'`

app/controllers/peeps.rb
```diff
@@ -0,0 +1,21 @@
...
+    time = Time.now
+    peep = Peep.create(
+      peep_text: params[:peep_text],
+      time: time,
+      user: current_user
+    )
+    redirect '/peeps_list'
+  end
+
+  get '/peeps_list' do
```
silvabox: this should be `get '/peeps'`

app/server.rb
```diff
@@ -0,0 +1,21 @@
...
+  enable :sessions
+  register Sinatra::Flash
+  register Sinatra::Partial
+  use Rack::MethodOverride
+  set :session_secret, 'super secret'
+  set :partial_template_engine, :erb
+
+  enable :partial_underscores
+
+  get '/' do
```
silvabox: this clashes with the route defined in app.rb

app/views/index.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Chitter!</h1>
+<form method='get' action='/sessions/new'>
```
silvabox: these don't need to be forms really - you could just use anchor tags, e.g.:
```html
<a href="/sessions/new">Sign in</a>
```

app/views/peeps/new.erb
```diff
@@ -0,0 +1,7 @@
+<h1> Let's Peep! </h1>
+<form action='/peeps/new' method='post'>
+  <label for='peep_text'>
+	Peep: <input name='peep_text' type='text'>
+	</label>
```
silvabox: watch your indentation.  Make sure your opening and closing tags align.

app/views/peeps_list.erb
```diff
@@ -0,0 +1,9 @@
+<ul id="peeps">
+  <% @peeps.each do |peep| %>
+  <li>
```
silvabox: this should be indented to show it's inside the above loop

PR #483, opened 2016-10-23
Chitter Submission

B-raw:


app/controllers/peeps.rb
```diff
@@ -0,0 +1,27 @@
+class Chitter < Sinatra::Base
+  get '/' do
+    erb(:index)
+  end
+
+  get '/peeps' do
+    erb(:'peeps')
```
silvabox: organize your view files into resource-centric sub folders.  You have a `app/views/peeps/` folder, so all peep related views should be in there.  The convention for the 'top level list' is to call it index.  So this would be `erb(:'peeps/index')`

app/controllers/peeps.rb
```diff
@@ -0,0 +1,27 @@
...
+      redirect '/sessions/new'
+    end
+  end
+
+  post '/peeps' do
+    time = Time.now
+    peep_content = Peep.new_line(params[:content])
+    current_user.peeps << Peep.create(content: peep_content,
+                                      created_at: time)
+    current_user.save
```
silvabox: I'm not sure you need to explicitly save here - I think the`<<` does it implicitly (but don't take my word on it).

app/models/peep.rb
```diff
@@ -0,0 +1,14 @@
+class Peep
+  include DataMapper::Resource
+
+  property :id,                Serial
+  property :content,           Text
+  property :created_at,        DateTime
+
+  belongs_to :user
+
+  def self.new_line(content)
```
silvabox: This is a presentation concern and should not be in the model.

app/controllers/peeps.rb
```diff
@@ -0,0 +1,27 @@
...
+      erb(:'peeps/new')
+    else
+      flash.keep[:errors] = ["You need to be logged in to post a peep"]
+      redirect '/sessions/new'
+    end
+  end
+
+  post '/peeps' do
+    time = Time.now
+    peep_content = Peep.new_line(params[:content])
```
silvabox: Nice idea, but _never do this_.  You've introduced presentation-layer cruft into your raw data.  What happens when you want to render your data in a different format (e.g. XML, JSON, PDF)?

app/views/_compose_peep.erb
```diff
@@ -0,0 +1,7 @@
+<div class="compose_button">
```
silvabox: this file should be `app/views/peeps/_compose.erb`

PR #484, opened 2016-10-23
As a Maker I can post  and login into Chitter

amidabrian51:
I can login into Chitter.


Gemfile
```diff
@@ -1,10 +1,18 @@
...
 ruby '2.2.3'
 
+gem 'sinatra'
+gem 'dm-postgres-adapter'
+gem 'data_mapper'
+gem 'database_cleaner'
+gem 'rake'
+gem 'sinatra-flash'
+gem 'sinatra-partial'
+gem 'capybara'
```
silvabox: this should be in the `test` group since you've defined one.

app.rb
```diff
@@ -0,0 +1,33 @@
+ENV["RACK_ENV"] ||= "development"
+require 'sinatra/base'
+require_relative 'data_mapper_setup'
+
+class Chitter < Sinatra::Base
+  enable :sessions
+  set :session_secret, 'super secret'
+
+  get '/users/new' do
+    User.new
```
silvabox: This line is redundant - the new `User` instance doesn't go anywhere.

PR #493, opened 2016-10-23
Week 4 Weekend Homework

LawrenceHunt:


Gemfile
```diff
@@ -1,10 +1,21 @@
 source 'https://rubygems.org'
 ruby '2.2.3'
 
+gem 'data_mapper'
+gem 'dm-postgres-adapter'
+gem 'sinatra'
+gem 'sinatra-partial'
+gem 'bcrypt'
+gem 'sinatra-flash'
+gem 'timecop'
```
silvabox: I think this should be in the `test` group

app/controllers/peeps.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  get '/peeps' do
+    @peeps = Peep.all_in_reverse_order
+    erb :'peeps/peeps'
+  end
+
+  post '/peeps' do
+      if current_user && !params[:peep_text].empty?
+          Peep.create(user: current_user,
+          text: params[:peep_text],
```
silvabox: indent wrapped lines so it's clear they are a continuation

app/controllers/peeps.rb
```diff
@@ -0,0 +1,38 @@
...
+          timestamp: Time.now)
+          redirect to('/peeps')
+      elsif params[:peep_text]
```
silvabox: I'm not sure I understand the logic here.

app/controllers/peeps.rb
```diff
@@ -0,0 +1,38 @@
...
+
+
+  post '/responses/:peep_id' do
+    if current_user && params[:response_text]
+      Response.create(text: params[:response_text],
+                      user_id: current_user.id,
+                      peep_id: params[:peep_id],
+                      timestamp: Time.now)
+      redirect to('/peeps')
+    elsif !params[:text]
```
silvabox: in general, try to avoid `elsif`.  Also, what is `params[:text]`?

app/controllers/peeps.rb
```diff
@@ -0,0 +1,38 @@
...
+          timestamp: Time.now)
+          redirect to('/peeps')
+      elsif params[:peep_text]
+        flash.keep[:errors] = "You need to write something."
+        redirect :'/peeps'
+      end
+  end
+
+
+  post '/responses/:peep_id' do
```
silvabox: this should probably be `post '/peeps/:id/responses'`

app/controllers/users.rb
```diff
@@ -0,0 +1,29 @@
...
+      session[:user_id] = @user.id
+      redirect to('/peeps')
+    else
+      flash.now[:errors] = @user.errors.full_messages
+      erb :'users/new'
+    end
+  end
+
+  get "/users/profile/:username" do
+   user = User.first(username: params[:username].downcase)
```
silvabox: You could encapsulate this search logic in `User`

app/controllers/peeps.rb
```diff
@@ -0,0 +1,38 @@
+class Chitter < Sinatra::Base
+
+  get '/' do
+    redirect '/peeps'
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all_in_reverse_order
+    erb :'peeps/peeps'
```
silvabox: conventionally, call the list page `index`

app/views/peeps/peeps.erb
```diff
@@ -0,0 +1,51 @@
...
+  <form method="post" action="/peeps">
+    <input type="text" name="peep_text" rows = "2" placeholder="What's peeping?" maxlength="140">
+    <input type="submit" value="Peep">
+  </form>
+  <% end %>
+</div>
+
+<div class="peeps">
+  <ul>
+    <% if @peeps %>
```
silvabox: will `@peeps` ever be falsey?

app/controllers/users.rb
```diff
@@ -0,0 +1,29 @@
...
+   peeps = user ? user.peeps(order: [:timestamp.desc]) : []
+   peep_list = partial(:"peep/peep", locals: { peeps: peeps })
```
silvabox: this is clumsy - the partial should be rendered in the view layer.

app/controllers/users.rb
```diff
@@ -0,0 +1,29 @@
...
+   erb(:"user/index", locals: { peep_formatted_list: peep_list, user: user })
```
silvabox: where is this view file?

PR #496, opened 2016-10-23
Chitter - where chit happens

tadasmajeris:


app/controllers/peeps.rb
```diff
@@ -0,0 +1,20 @@
...
+
+class Chitter < Sinatra::Base
+
+  get '/peeps' do
+    @peeps = Peep.all(:order => [:created_at.desc])
+    erb :'/peeps/index'
+  end
+
+  post '/peeps' do
+    if current_user && !params[:peep_text].empty?
```
silvabox: the validation of the peep text belongs in the model.

app/controllers/comments.rb
```diff
@@ -0,0 +1,17 @@
+require 'sinatra/base'
+
+class Chitter < Sinatra::Base
+
+  post '/comments/:peep_id' do
```
silvabox: close - it should probably be `post '/peeps/:id/comments'`

app/controllers/comments.rb
```diff
@@ -0,0 +1,17 @@
...
+    if current_user && !params[:comment_text].empty?
```
silvabox: comment text validation should be in the model

app/controllers/peeps.rb
```diff
@@ -0,0 +1,20 @@
+require 'sinatra/base'
+
+class Chitter < Sinatra::Base
+
+  get '/peeps' do
+    @peeps = Peep.all(:order => [:created_at.desc])
```
silvabox: It would be preferable to push this logic into the model for better encapsulation.  Then you could have something like `@peeps = Peep.in_reverse_order`

app/views/peeps/index.erb
```diff
@@ -0,0 +1,61 @@
...
+        <div class="peeper"><b><%=peep.user.username%></b> <i> - <%=peep.user.name%></i></div>
+        <div class="peep_date"><%=peep.created_at.strftime('%Y-%m-%d %H:%M')%></div>
+        <p><%=peep.peep_text%></p>
+
+        <div class="comments">
+
+          <% if !peep.comments.empty? %>
+            <% peep.comments.each do |comment| %>
+              <div class="comment">
+                <div class="peeper"><b><%=comment.user.username%></b> <i> - <%=comment.user.name%></i></div>
```
silvabox: I wouldn't use the `peeper` class here as this is a comment, not a peep.

PR #497, opened 2016-10-23
Weekend Challenge

Swinston88:


app/controllers/peeps.rb
```diff
@@ -0,0 +1,13 @@
+class Chitter < Sinatra::Base
+  get '/peeps' do
+    @peeps = Peep.all.reverse
+    erb :'peeps/index'
+  end
+
+  post '/peeps/new' do
```
silvabox: this should be just `post '/peeps'`

PR #502, opened 2016-10-23
Chitter Pull Request

RachaelMahon:
Seems to be working in local host - some issues with heroku I'm currently debugging

### DIFF
--- a/app/app.rb
+++ b/app/app.rb
@@ -1,10 +1,13 @@
 ENV["RACK_ENV"] ||=  "development"
 require 'sinatra/base'
+require "sinatra/flash"
 require_relative 'data_mapper_setup'
+require_relative "models/peep"
+require_relative "models/user"
 
 class Chitter < Sinatra::Base
   use Rack::MethodOverride
-
+  register Sinatra::Flash
   enable :sessions
   set :session_secret, 'super secret'
 
@@ -13,11 +16,15 @@ class Chitter < Sinatra::Base
   end
 
   post '/log_in' do
-    user = User.get(username: params[:username], password: params[:password])
-    current_user
-    session[:user_id] = current_user.id
-    redirect '/chitter'
-  end
+    user = User.authenticate(params[:username], params[:password])
+   if user
+     session[:user_id] = user.id
+     redirect '/chitter'
+   else
+     flash.now[:errors] = ['Your email or password is incorrect - try sign up!']
+     erb :index
+   end
+ end
 
   get '/sign_up' do
     erb :sign_up
@@ -31,8 +38,8 @@ class Chitter < Sinatra::Base
 
   get '/chitter' do
     p Peep.all
-    peeps = Peep.all
-    p peeps
+    @peeps = Peep.all
+    @peeps = @peeps.reverse
     erb :chitter
   end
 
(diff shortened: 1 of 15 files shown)


app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+  enable :sessions
+  set :session_secret, 'super secret'
+
+  get '/' do
+    erb :index
+  end
+
+  post '/log_in' do
+    user = User.authenticate(params[:username], params[:password])
+   if user
```
maryrosecook: Watch out for this indentation! Have a look at a Ruby style guide for idiomatic Ruby indentation.

app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+   end
+ end
+
+  get '/sign_up' do
+    erb :sign_up
+  end
+
+  post '/add_user_data' do
+    user = User.create(username: params[:username], password: params[:password])
+    session[:user_id] = user.id
```
maryrosecook: Could the lines than set session user_id be refactored to use a helper function? eg:

```
def log_in_user(user_id)
  session[:user_id] = user_id
end
```

app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+    redirect '/chitter'
+  end
+
+  get '/chitter' do
+    @peeps = Peep.all
```
maryrosecook: Could you do this line and the next one in one line for clarity and succinctness?

app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+  end
+
+  delete '/log_out' do
+    session[:user_id] = nil
+    flash.now[:notice] = 'goodbye!'
+    redirect to '/'
+  end
+
+  helpers do
+    def current_user
```
maryrosecook: Love this `current_user` method!

app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  get '/chitter' do
+    @peeps = Peep.all
+    @peeps = @peeps.reverse
+    erb :chitter
+  end
+
+  post '/add_peep' do
+    peep = Peep.create(content: params[:content])
+    session[:user_id]
```
maryrosecook: Is this line required?

app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+    peep.user = current_user
```
maryrosecook: Can you pass `current_user` into your call of `create` a few lines up? This would move this model logic out of the controller, and you'd no longer need to call `save` on the line below.

app/app.rb
```diff
@@ -0,0 +1,66 @@
...
+    peep.save
+    redirect '/chitter'
+  end
+
+  delete '/log_out' do
+    session[:user_id] = nil
```
maryrosecook: Could you define a `log_out_user` helper method to clarify this code?

app/models/user.rb
```diff
@@ -0,0 +1,32 @@
...
+  attr_reader :password
+  attr_accessor :password_confirmation
+  validates_confirmation_of :password, message:"Password and confirmation password do not match"
+  validates_presence_of :password
+
+  def password=(password)
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(username, password)
```
maryrosecook: Would it be possible to rewrite this method to use a guard clause that raises an error if the password is wrong? It might make the control flow even clearer.

PR #503, opened 2016-10-23
CHITTER

wirsindpapst:


wirsindpapst:
Hi Ben,

I know you're being super strict on what you spend time reviewing, and totally understand the rationale for this, but if I can make a desperate plea.... 

My build is failing because of a DB issue, one of the command in the Travis.yml file is failing, and despite much research I haven't been able to rectify it. Locally however, all tests are passing and coverage is high (see below screenshot), so if I can make a case for it to be reviewed anyway it'd be much appreciated.

screen shot 2016-10-24 at 08 50 45

In addition (getting my excuses in early this time...), there is another config issue in that I couldn't get spec to run with sinatra-partial (a gem conflict this time). As such, whilst this pull request does not use partials, and therefore is not refactored, there is a separate branch 'refactor' on my repo, which does do this. 

Cheers,

Liam

Gemfile
```diff
@@ -1,10 +1,22 @@
 source 'https://rubygems.org'
 ruby '2.2.3'
 
+gem 'data_mapper'
+gem 'dm-postgres-adapter'
+gem "rspec-sinatra"
```
silvabox: you don't need `rspec-sinatra`, `capybara` or `database_cleaner` in production.

app/chitter_app.rb
```diff
@@ -0,0 +1,76 @@
...
+  get '/users/register' do
+    @user = User.new
+    erb :'users/register'
+  end
+
+  get '/sessions/sign_in' do
+    erb :'sessions/sign_in'
+  end
+
+  post '/users/new_user_authenticate' do
```
silvabox: this route name is incorrect, it should just be `post '/users'`.  You are not authenticating the user at this stage.

app/chitter_app.rb
```diff
@@ -0,0 +1,76 @@
...
+    if user
+      session[:user_id] = user.id
+      redirect '/peeps/peepdeck'
+    else
+      flash.now[:errors] = ['The email or password is incorrect']
+      erb :'sessions/sign_in'
+    end
+  end
+
+  get '/peeps/peepdeck' do
```
silvabox: `peepdeck` is superfluous.

PR #506, opened 2016-10-24
Chitter app

amaalali:
Chitter social app


app/models/peep.rb
```diff
@@ -0,0 +1,11 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+
+class Peep
+  include DataMapper::Resource
+
```
silvabox: should this not be related to `User` somehow?

app/models/user.rb
```diff
@@ -0,0 +1,31 @@
+require 'bcrypt'
+
+class User
+  include DataMapper::Resource
+
```
silvabox: should this not be related to `Peep` somehow?

spec/features/register_spec.rb
```diff
@@ -0,0 +1,73 @@
+feature 'Users can sign up for the service' do
+
+  it 'with valid and correct entries' do
```
silvabox: word your `it` tests according to the expected behaviour; not just the context.  it _does what_ with valid and correct entries?  Alternatively, use `scenario` instead of `it`.

PR #510, opened 2016-10-24
working but no CSS

James-SteelX:


app/app.rb
```diff
@@ -0,0 +1,90 @@
...
+    if user
+      session[:user_id] = user.id
+      redirect to('/')
+    else
+      flash.now[:errors] = ['The email or password is incorrect']
+      erb :'/sign_in'
+    end
+  end
+
+  get '/new_peep' do
```
silvabox: should be `get '/peeps/new'`

app/app.rb
```diff
@@ -0,0 +1,90 @@
...
+    erb :'/new_peep'
+  end
+
+  post '/new_peep' do
```
silvabox: should be `post '/peeps'`

app/models/peep.rb
```diff
@@ -0,0 +1,15 @@
...
+require 'dm-postgres-adapter'
+require 'dm-migrations'
+
+class Peep
+
+  include DataMapper::Resource
+
+  property :id, Serial
+  property :author, String, required: true
+  property :user_name, String, required: true
```
silvabox: what happens if the user changes their name?  Shouldn't the peep be directly related to the user?

app/models/user.rb
```diff
@@ -0,0 +1,34 @@
+require 'bcrypt'
+
+class User
+  include DataMapper::Resource
+
+  has n, :peeps, through: Resource
```
silvabox: is this the correct relationship?

app/views/_welcome.erb
```diff
@@ -0,0 +1,20 @@
...
+    <input type="hidden" name="_method" value="delete">
+    <input type="submit" value="Sign Out">
+  </form>
+
+  <form action="/new_peep" method="get">
+    <input type="submit" value="New Peep!">
+  </form>
+
+<% else %>
+  <% if !request.path_info.include?('/sessions') || request.path_info == 'sessions/delete' %>
```
silvabox: The view should not be concerned with the request object - that's the controller's job.

app/views/index.erb
```diff
@@ -0,0 +1,17 @@
+<h1>Recent Peeps!</h1>
+
+<ul id = 'peeps'>
+  <% @peeps.reverse_each do |peep| %>
+  <li>
+    Time: <%= peep.peep_time %>
+    <br>
```
silvabox: generally avoid `<br>` tags as they imply layout, which is not the job of HTML.  HTML should only be concerned with semantic structure.

app/app.rb
```diff
@@ -0,0 +1,90 @@
...
+   if current_user.nil?
+     flash.keep[:errors] = ['Please sign in to post a peep']
+     redirect('/')
+   else
+    peep = Peep.create(author: current_user.name, user_name: current_user.user_name, peep_time: Time.now.strftime("%d/%m/%Y %H:%M"), text: params[:new_peep])
```
silvabox: what if the peep is invalid?

silvabox:
Hi James, you are not committing to git frequently enough unless you rebased this before sending the PR?

You should commit your code at every completed stage, which is generally one completed feature.

PR #511, opened 2016-10-24
Chitter challenge - Irene Oppo

souljuse:


app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    end
+  end
+
+  get '/' do
+    @users = User.all
+    @peeps = Peep.all
+    erb :index
+  end
+
+  post '/new_peep' do
```
silvabox: this should be `post '/peeps'`

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    peep = Peep.new(title: params[:title], content: params[:content], created_at: Time.now, user_id: current_user.id)
+    peep.save
+    redirect('/')
+  end
+
+  get '/tags/:name' do
```
silvabox: erm...?

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+  enable :partial_underscores
+
+  helpers do
+    def current_user
+      @current_user ||= User.get(session[:user_id])
+    end
+  end
+
+  get '/' do
+    @users = User.all
```
silvabox: Do you need all users at this stage?

app/views/index.erb
```diff
@@ -0,0 +1,37 @@
...
+		</label>
+		<button>Create a new Peep</button>
+	</form>
+<% else %>
+	<p>
+		Log in to post peeps
+	</p>
+<% end %>
+
+<% if !!@peeps %>
```
silvabox: this is redundant as `!!@peeps` will always be true, since `@peeps` will always be something (even if it's empty)

PR #512, opened 2016-10-24
Elizabeth Venner - Chitter Challenge

elizabethvenner:


app/app.rb
```diff
@@ -0,0 +1,73 @@
...
+require_relative 'data_mapper_setup'
+
+class ChitterChatter < Sinatra::Base
+  use Rack::MethodOverride
+  register Sinatra::Flash
+  enable :sessions
+  set :session_secret, 'super secret'
+
+  get '/' do
+    @peeps = Peep.all.reverse!
```
maryrosecook: No need for the !

app/app.rb
```diff
@@ -0,0 +1,73 @@
...
+
+  helpers do
+    def current_user
+      @current_user ||= User.get(session[:user_id])
+    end
+  end
+
+  post '/users' do
+    @user = User.create(email: params[:email],
+    password: params[:password],
```
maryrosecook: Can this indentation make it clearer that these are arguments to the create method?

app/app.rb
```diff
@@ -0,0 +1,73 @@
...
+    password_confirmation: params[:password_confirmation],
+    name: params[:name],
+    username: params[:username])
+    if @user.save
+      session[:user_id] = @user.id
+      redirect to '/'
+    else
+      @user.errors.each do |e|
```
maryrosecook: It feels like this won't show all the errors on the user if there is more than one.

app/app.rb
```diff
@@ -0,0 +1,73 @@
...
+  end
+
+  post '/users' do
+    @user = User.create(email: params[:email],
+    password: params[:password],
+    password_confirmation: params[:password_confirmation],
+    name: params[:name],
+    username: params[:username])
+    if @user.save
+      session[:user_id] = @user.id
```
maryrosecook: Could the lines than set session user_id be refactored to use a helper function eg:

```
def log_in_user(user_id)
  session["user_id"] = user_id
end
```

app/app.rb
```diff
@@ -0,0 +1,73 @@
...
+      erb :sign_up
+    end
+  end
+
+  get '/sessions/new' do
+    erb :'sessions/new'
+  end
+
+  post '/sessions' do
+    user = User.authenticate(params[:email], params[:password])
```
maryrosecook: Can this line and the next one be put onto one line using a Ruby if-assignment?

app/models/user.rb
```diff
@@ -0,0 +1,49 @@
...
+             :presence => "Please fill in your email address",
+             :is_unique => "This email address has already been registered, please use a different one",
+             :format => "Please fill in a valid email address"
+           }
+  property :password_digest, Text, :required => true
+  property :name, String, :required => true,
+           :messages => {
+             :presence => "Plesae fill in your name"
+           }
+  property :username, String, :required => true, :unique => true,
```
maryrosecook: Love this clear formatting!

PR #515, opened 2016-10-24
Chitter-challenge homework

powderham:


app/app.rb
```diff
@@ -0,0 +1,49 @@
...
+require "sinatra/flash"
+require "sinatra/partial"
+require_relative "models/peep"
+require_relative "models/user"
+require_relative "models/data_mapper_setup"
+require_relative "controllers/peeps"
+require_relative "controllers/users"
+require_relative "controllers/sessions"
+
+data_mapper_configure
```
maryrosecook: Can you put this method in a module in `data_mapper_setup` so you can call it with `DataMapperSetup.data_mapper_configure`.  This namespaces the method to make it clear what section of the app it belongs to.

app/controllers/sessions.rb
```diff
@@ -0,0 +1,27 @@
+class Chitter < Sinatra::Base
```
maryrosecook: Love that you split things into separate controllers!

app/controllers/sessions.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  get '/sessions/new' do
+    erb :'sessions/sign_in'
+  end
+
+  post '/sessions' do
+    user = User.authenticate(params[:email], params[:password])
+    p user
+    if user
+      session[:user_id] = user.id
```
maryrosecook: Can you write a login helper to make it clearer what this is doing?

app/controllers/sessions.rb
```diff
@@ -0,0 +1,27 @@
...
+    else
+      flash.now[:errors] = ['The email or password is incorrect']
+      erb :'sessions/sign_in'
+    end
+  end
+
+  delete '/sessions' do
+    @peeps = Peep.all
+    @users = User.all
+    session.clear
```
maryrosecook: Maybe add a log_out helper to make it clearer what this is doing.

maryrosecook: Watch out for clearing all the session data.  If you start using the session for something else, this line will blow that data away, too.

app/controllers/users.rb
```diff
@@ -0,0 +1,24 @@
+class Chitter < Sinatra::Base
+
+  get '/users/new' do
+    @user = User.new
+    erb :'users/new'
+  end
+
+  post '/users' do
+    @user = User.new(email: params[:email],
+                username: params[:username],
```
maryrosecook: Try to align all args that are split by newlines.

app/models/user.rb
```diff
@@ -0,0 +1,38 @@
...
+require "dm-timestamps"
+require "bcrypt"
+
+class User
+  include DataMapper::Resource
+  attr_accessor :password_confirmation
+  attr_reader :password
+
+  property :id, Serial
+  property :email, String, required: true,
```
maryrosecook: Love this formatting! Really clear.

app/models/user.rb
```diff
@@ -0,0 +1,38 @@
...
+           }
+  property :username, String
+  property :name, String
+  property :password_digest, Text
+  validates_confirmation_of :password,
+           :message => 'The two passwords submitted did not match'
+
+  has n, :peeps
+
+  def self.authenticate(email, password)
```
maryrosecook: Can you rewrite this to use a guard clause it make its flow a bit clearer?

PR #540, opened 2016-12-04
completing chitter challenge

aabolade:


spec/spec_helper.rb
```diff
@@ -6,3 +7,125 @@
...
+# See http://rubydoc.info/gems/rspec-core/RSpec/Core/Configuration
+RSpec.configure do |config|
+
+  config.before(:suite) do
+    DatabaseCleaner.clean_with(:truncation)
+  end
+
+  config.before(:each) do
+    DatabaseCleaner.strategy = :transaction
+  end
```
roidriscoll: You have a `config.before(:each)` block below too, so you're duplicating the effort, although I would expect to see `DatabaseCleaner.strategy = :transaction` in the`config.before(:suite)` above.

PR #564, opened 2017-01-29
Rick Clark pull request

rkclark:


rkclark:
Self Review:

Was happy with the challenge over all and enjoyed it. Less Hound issues this time (and I did fix some of the things it told me about!).

Only bit I got stuck on was getting Travis to work, which is annoying as I did land on the correct solution but didn't implement it properly until Stefan gave me a nudge in the right direction.

I think my bootstrap structure is simpler this time and the use of partials breaks things up nicely.

Areas for improvement based on code review:
- Rakefile works but is not DRY at all.
- My DB Cleaner is using truncation throughout, not transaction
- I moved the public folder into the app file to avoid having to change the public_folder config, I guess it should be in the project root?
- I think bits of my feature tests could be consolidated but I don't think they are too 'unit-testy'
- I have three spec helper modules, not one. I like it like that but not sure whether it was necessary for this size of project.
- I put view helpers inside the relevant controller than than in a separate module

app/views/_nav.erb
```diff
@@ -0,0 +1,30 @@
...
+        <a class="nav-link btn" href="/sessions/new" id="sign_in">Sign In</a>
+      </li>
+      <li class="nav-item">
+        <a class="nav-link btn" href="/users/new" id="sign_up">Register</a>
+      </li>
+      <% else %>
+      <li class="nav-item new-peep-button">
+        <a class="nav-link btn" href="/peeps/new" id="new_peep">New Peep</a>
+      </li>
+      <% end %>
```
neoeno: I'd consider indenting this control flow stuff in ERB

app/views/index.erb
```diff
@@ -0,0 +1 @@
+<%= partial :peeps %>
```
neoeno: Good instinct here — makes it much more readable!

sass/main.scss
```diff
@@ -0,0 +1,133 @@
...
+  left:0;
+  width: 100%;
+}
+
+.small-text {
+  font-size: 0.6em;
+}
+
+.main-font-color {
+  color: $main-font-color !important;
```
neoeno: I agree with Hound!

spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,44 @@
+feature "FEATURE: post peep" do
+
+  let(:name) { 'J T Kirk' }
+  let(:email) { 'JTK@test.com' }
+  let(:user_name) { 'TheShat' }
+  let(:password) { 'password' }
+  let(:password_confirmation) { 'password' }
```
neoeno: Neat to have these extracted. I might be tempted to make them constants too.

spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+    sign_in(email, password)
+    post_peep("")
+    expect(page).to have_content "Peep text must not be blank"
+  end
+
+  scenario "Cannot add a peep over 250 characters" do
+    sign_in(email, password)
+    post_peep("This is the Captain of the Enterprise. Our respect for other life forms requires that we give you this... warning. There is one critical item of information that has never been incorporated into the memory banks of any Earth ship. Since the early years of space exploration, Earth vessels have had incorporated into them a substance known as... corbomite. It is a material and a device which prevents attack on us. If any destructive energy touches our vessel, a reverse reaction of equal strength is created, destroying -...")
+    expect(page).to have_content "Peep text must be at most 250 characters long"
+  end
```
neoeno: Might be worth checking some boundary conditions on this one (e.g. 249, 250, 251)

spec/features/sign_up_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    visit('/')
+    click_link('sign_up')
+    expect(page.current_path).to eq "/users/new"
+  end
+
+  let(:name) { 'J T Kirk' }
+  let(:email) { 'JTK@test.com' }
+  let(:user_name) { 'TheShat' }
+  let(:password) { 'password' }
+  let(:password_confirmation) { 'password' }
```
neoeno: These have been repeated a few times — *might* be worth extracting them someplace common, though some might say that makes them a [mystery guest](https://robots.thoughtbot.com/mystery-guest).

spec/features/sign_up_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+    end
+
+    scenario "cannot click to sign up" do
+      sign_in(email, password)
+      expect(page).not_to have_selector("#sign_up")
+    end
+  end
+
+  scenario 'Can enter user details and create new user' do
+    sign_up('James T Kirk', 'JTK@test.com', 'TheShat', 'password', 'password')
```
neoeno: Might be an opportunity for keyword arguments to make it clearer what all these things are.

neoeno:
Oh — and on the `public` folder thing — I wouldn't worry about it too much. There's probably a right answer but as long as the directory structure does its job of revealing intent then it's good to me :+1:

spec/features/sign_up_spec.rb
rkclark: Yes I thought about this for a while, and figured that if I extracted them then it would be much harder to understand the tests, so decided to leave them :)

app/controllers/user.rb
```diff
@@ -0,0 +1,23 @@
...
+  post '/users' do
+    user = User.create(
+      name: params[:name],
+      email: params[:email],
+      user_name: params[:user_name],
+      password: params[:password],
+      password_confirmation: params[:password_confirmation])
+
+      session[:user_id] = user.id
+    if user.save
```
rkclark: Good point!

PR #639, opened 2017-05-07
Chitter Challenge

BertZZ:


spec/web_helpers.rb
```diff
@@ -0,0 +1,45 @@
...
+  click_button 'Sign Up'
+  fill_in :name, with: "Bertie"
+  fill_in :email, with: "invalid@email"
+  fill_in :username, with: "BertZ"
+  fill_in :password, with: "Password1234"
+  fill_in :password_confirmation, with: "Password1234"
+  click_button "Create Account"
+end
+
+def sign_in(username:, password:)
```
neoeno: Hound is right!

app/data_manager_setup.rb
```diff
@@ -0,0 +1,9 @@
+require 'data_mapper'
+require 'dm-postgres-adapter'
+
+require_relative 'models/user'
+require_relative 'models/peeps'
+
+DataMapper.setup(:default, ENV['DATABASE_URL'] || "postgres://localhost/chitter_#{ENV['RACK_ENV']}")
+DataMapper.finalize
+DataMapper.auto_upgrade!
```
neoeno: This file is pretty great — but it's not required anywhere, right?

PR #642, opened 2017-05-07
Not yet finished but pulling in early just in case.

motri:


Rakefile
```diff
@@ -1,7 +1,25 @@
...
+  end
+end
+
+
+  desc "Destructive upgrade"
+  task :auto_migrate do
+    DataMapper.auto_migrate!
+    puts "Auto-migrate complete (data was lost)"
+  end
+end
```
neoeno: Running rubocop locally would have informed you of a syntax error in this file!

PR #695, opened 2017-07-02
Chitter weekend challenge

vhonebon1:


app/app.rb
```diff
@@ -0,0 +1,76 @@
...
+    @peeps = Peep.all
+    erb :'links/index'
+  end
+
+  post '/peeps/new' do
+    Peep.create(content: params[:new_peep], creator: params[:username])
+    redirect to '/peeps'
+  end
+
+  get '/peeps/new' do
```
maryrosecook: It's often helpful to name templates after the relevant part of the URL.  Using homogenous language for the same concept really aids code readability.

app/app.rb
```diff
@@ -0,0 +1,76 @@
...
+    erb :'links/add_peep'
+  end
+
+  get '/users/new' do
+    @user = User.new
+    erb :'users/new'
+  end
+
+  post '/users' do
+    @user = User.create(email: params[:email],
```
maryrosecook: Is there a way to avoid needing to pass each piece of user data in separately (eg email, username etc)?

app/app.rb
```diff
@@ -0,0 +1,76 @@
...
+                       name: params[:name],
+                       username: params[:username],
+                       password: params[:password],
+                       password_confirmation: params[:password_confirmation])
+    if @user.save
+      session[:user_id] = @user.id
+      redirect to '/peeps/new'
+    else
+      flash.now[:notice] = "Password and confirmation password do not match"
```
maryrosecook: Can this line be made more declarative of what it does? At the moment, it seems to reveal its implementation, rather than its intention.

app/app.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  post '/users' do
+    @user = User.create(email: params[:email],
+                       name: params[:name],
+                       username: params[:username],
+                       password: params[:password],
+                       password_confirmation: params[:password_confirmation])
+    if @user.save
+      session[:user_id] = @user.id
```
maryrosecook: Can this line be made more declarative of what it does? At the moment, it seems to reveal its implementation, rather than its intention.

app/app.rb
```diff
@@ -0,0 +1,76 @@
+ENV["RACK_ENV"] ||= "development"
+
+require 'sinatra/base'
+require 'sinatra/flash'
+require_relative './models/peep.rb'
+require_relative './models/user.rb'
+require_relative 'data_mapper_setup'
+
+class Chitter < Sinatra::Base
```
maryrosecook: Could the separate resources be handled by different controllers to improve the SRPness of this code?

app/views/layout.erb
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <title>Chitter</title>
+  </head>
+
+  <body>
+    <% if current_user %>
```
maryrosecook: It feels like there are two responsibilities in this template.  Could you use partials to delegate these responsibilities?

spec/features/view_peeps_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+    Peep.create(content: 'This is my second peep!')
+    visit('/peeps')
+    within 'ul#peeps' do
+      time = Time.now
+      time = time.strftime("%H:%M")
+      expect(page).to have_content 'This is my second peep!'
+      expect(page).to have_content time
+    end
+  end
+  scenario 'Username is dislayed with peep' do
```
maryrosecook: Try to use whitespace idiomatically.  The scenarios should be separated by a blank line.

PR #703, opened 2017-07-03
Chitter Challenge attempt

alistairkung:


README.md
```diff
@@ -65,48 +42,20 @@ Notes on functionality:
...
------
-
-If you have time you can implement the following:
 
-* In order to start a conversation as a maker I want to reply to a peep from another maker.
+Mini code review:
+-------------
 
-And/Or:
+Overall I feel that I have managed to improve on many of the concepts that were introduced to us last week:
```
maryrosecook: Great!

app/controllers/peeps.rb
```diff
@@ -0,0 +1,23 @@
+
+class Chitter < Sinatra::Base
+
+  get '/peeps' do
```
maryrosecook: Really love that there is one controller per resource!

app/controllers/peeps.rb
```diff
@@ -0,0 +1,23 @@
...
+  get '/peeps/new' do
+    erb :'peeps/new'
+  end
+
+  post '/peeps' do
+    if current_user
+      peep = Peep.create(message: params[:peep], user_id: session[:user_id])
+      redirect '/peeps'
+    else
+      flash.now[:errors] = ['You must be signed in to post a peep']
```
maryrosecook: Can this code be made more declarative of what it does? At the moment, it seems to reveal its implementation, rather than its intention.

app/controllers/sessions.rb
```diff
@@ -0,0 +1,29 @@
...
+class Chitter < Sinatra::Base
+
+  get '/sessions/new' do
+    erb :'sessions/new'
+  end
+
+  post '/sessions' do
+    user = User.authenticate(params[:email], params[:password])
+    if user
+      session[:user_id] = user.id
```
maryrosecook: Can this code be made more declarative of what it does? At the moment, it seems to reveal its implementation, rather than its intention.

app/controllers/users.rb
```diff
@@ -0,0 +1,25 @@
+
+class Chitter < Sinatra::Base
+
+  get '/users/new' do
+    @user = User.new
+    erb :'users/new'
+  end
+
+  post '/users' do
+    @user = User.create(email: params[:email],
```
maryrosecook: Is there a way to avoid having to each piece of data (email, username etc) explicity?

app/models/peep.rb
```diff
@@ -0,0 +1,14 @@
+class Peep
+  include DataMapper::Resource
+
+  belongs_to :user, required: true
+
+  property :id, Serial
+  property :message, String
+  property :created_at, DateTime
+
+  def peeper_username
```
maryrosecook: Nice helper!

app/models/user.rb
```diff
@@ -0,0 +1,33 @@
+require 'bcrypt'
+
+class User
+  include DataMapper::Resource
+
+  has n, :peeps
+
+  property :id, Serial
+  property :email, String, required: true, unique: true
+  property :firstname, String, required: true
```
maryrosecook: What is the convention for table column names with multiple words in them (eg first name)?

app/views/layout.erb
```diff
@@ -0,0 +1,29 @@
...
+    <% end %>
+    <% if flash[:errors] && !flash[:errors].empty? %>
+      Please refer to the following errors below:
+      <ul id='errors'>
+        <% flash[:errors].each do |error| %>
+          <li><%= error %></li>
+        <% end %>
+      </ul>
+    <% end %>
+    <% if current_user %>
```
maryrosecook: Yes, like you said - definitely worth using partials to better delegate the different parts of the responsibility.

PR #779, opened 2017-11-19
Submission

Eustaquio122:


app/controllers/user_controller.rb
```diff
@@ -0,0 +1,17 @@
+class Chitter < Sinatra::Base
+  get '/sign_up' do
+    erb :'user/new'
+  end
+
+  post '/register_user' do
+    user = create_user(params)
+    create_session_data(user)
+    redirect '/sign_up' unless user.id
```
mdsomerfield: I think using 'unless' here makes it feel like user.id is the exception case, rather than expected case. I'd rather see redirect `"user/..." unless user.id == nil `etc, it makes it clearer that 'normally we do this, unless it has failed, then we do this'

app/helpers/tag_helpers.rb
```diff
@@ -0,0 +1,11 @@
+class Chitter < Sinatra::Base
+  def create_tag(user)
+    Tag.create(tag_name: "@#{user.username}", user_id: user.id)
+  end
+
+  def check_for_tags(content)
+    tags = content.scan(/[$|\s](@[^\s]*)+/)
```
mdsomerfield: It feels like this method is doing two things - scanning the content for tags and then adding them to the database. It also contains some logic around that process (check_for_mentions) which isn't particularly clear - I don't get a good sense of intent from the code or the method names.

Worth thinking about side effects here - if you call a method 'check_for_tags', do you expect one of the side effects to be that they get added to the database?

PR #785, opened 2017-11-19
Chitter

tmerrr:
First draft complete.


app/app.rb
```diff
@@ -0,0 +1,104 @@
...
+
+  get '/goto/users/sign_in' do
+    redirect '/users/sign_in'
+  end
+
+  post '/users/sign_in' do
+    login_name = params[:username_or_email]
+    password = params[:password]
+    user = find_user(login_name)
+    if user
```
mdsomerfield: Whenever I see nested 'if' statements, I think that there must be a way to refactor the method. As a rule of thumb, if you have multiple levels of nesting, some of the logic probably belongs elsewhere.

app/app.rb
```diff
@@ -0,0 +1,104 @@
...
+    @user = User.new
+    erb :sign_up
+  end
+
+  get '/goto/users/new' do
+    redirect 'users/new'
+  end
+
+  post '/users/new' do
+    name = params[:name]
```
mdsomerfield: I'm not sure you gain anything by assigning these params to variables here - it makes the method a bit more cluttered and harder to read as a result.

app/app_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+end
+
+def find_user(username_or_email)
+  user = User.first(username: username_or_email)
+  user = User.first(email: username_or_email) if user.nil?
+  user
+end
+
+def sign_up_error_message(username, email, password, password_confirmation)
+  flash.now[:alert] = if User.all(username: username).any?
```
mdsomerfield: I feel like this logic is in the wrong place. You're attempting to create a user, and if the user isn't successfully created, you're analysing the inputs to suggest what might have gone wrong.

A more normal model is that you validate the inputs before attempting to create the user, and creating the user fails, you analyse the exception to extract a suitable error message, even if that message is just "Unexpected error"

app/app_helper.rb
```diff
@@ -0,0 +1,22 @@
+def current_user
+  @current_user ||= User.get(session[:user_id])
+end
+
+def find_user(username_or_email)
+  user = User.first(username: username_or_email)
+  user = User.first(email: username_or_email) if user.nil?
```
mdsomerfield: I think conditional clauses at the end of line statements are a bit abused in ruby, the above code is harder to read because when I first see it I think 'what, he's searching the user twice', then I scan right and I realise 'oh, right, only if the first user is nil', but it's very easy not to scan right and if I wasn't paying special attention to the code I might miss this.

I would prefer to see something like
```
user = User.first(username: username_or_email) || User.first(email: username_or_email) 
```
This code is equivalent to your lines 6 and 7, but makes the intent clearer.

spec/features/new_peep_spec.rb
```diff
@@ -0,0 +1,25 @@
+feature 'Create Peep:' do
+  scenario 'A user can create a new Peep message' do
+    create_account('Test', 'mrtest', 'test@test.uk', 'Testing')
+    write_new_peep('hello world')
+    expect(Peep.all.length).to eq(1)
+    expect(page.current_url.ends_in?('/peeps')).to be(true)
+    expect(page).to have_content('hello world')
+    expect(page).to have_content('mrtest')
```
mdsomerfield: Feels like you are testing multiple things in the same test here. A test should usually only have one reason to fail.

spec/features/user_signup_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  scenario "can't use an existing username or email address" do
+    create_account('Tom Moir', 'tmerrr', 'tom@mail.me', 'MyPassword')
+    # Same username:
+
+    expect { create_account('Tim Muir', 'tmerrr', 'tim@gmail.com', 'MyPassword') }
+      .not_to change { User.all.length }
+    expect(page).to have_content('Username already exists')
+
+    # Same email:
```
mdsomerfield: As per above, I think this test is testing two cases - rather than breaking them up with comments, we should break them up into different tests.

app/views/peeps.erb
```diff
@@ -0,0 +1,51 @@
...
+  <head>
+    <meta charset="utf-8">
+    <link rel="stylesheet" type="text/css" href="stylesheet.css">
+    <title> Chitter </title>
+  </head>
+
+  <div>
+      <h1> Latest Peeps </h1>
+
+    <% if @current_user %>
```
mdsomerfield: Rather than a view which had two different states, I would expect this to be two different views (and possibly two different routes)

PR #792, opened 2017-11-19
Gabriela - Chitter challenge submission

gabrielabud:


app/models/user.rb
```diff
@@ -0,0 +1,34 @@
...
+  has n, :peeps
+
+  def password=(password)
+    @password = password
+    self.password_digest = BCrypt::Password.create(password)
+  end
+
+  def self.authenticate(email, password)
+    user = first(email: email)
+    if user && BCrypt::Password.new(user.password_digest) == password
```
mdsomerfield: This feels like the wrong way round. Shouldn't this be:
`BCrypt::Password.new(password) == user.password_digest`
Normally password encryption is one way. I would expect that we would encrypt the password the user supplies and compare that to the digest.

app/views/layout.erb
```diff
@@ -0,0 +1,40 @@
...
+    <title>Chitter</title>
+    <link href="https://fonts.googleapis.com/css?family=Lato:400,700" rel="stylesheet">
+    <link href="/normalize.css" rel="stylesheet">
+    <link href="/style.css" rel="stylesheet">
+  </head>
+
+  <body>
+    <div class="everything">
+      <header>
+      <% if current_user %>
```
mdsomerfield: I would probably make this two different views (and return a different view from the controller depending omn the sign in state) rather than an if/else in the view. It makes it explicit earlier in the flow that there is a signed in/signed out state (which helps with separation of logical concerns) and would also make the views easier to read.

PR #819, opened 2017-12-18
Chitter

Le5tes:


Le5tes:
Still need to add in the mailgun API key as an environment variable and feature test

Le5tes:
Turns out it does successfully send emails. 

Also the environment file is not committed to GitHub, to preserve the privacy of my Mailgun api key, which is why it's failing.

PR #827, opened 2018-01-28
chitter-challenge

GiadaSimonetti:


README.md
```diff
@@ -55,58 +40,6 @@ I want to receive an email if I am tagged in a Peep
...
-
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
```
EdwardAndress: It would be helpful to have a bit more guidance here in terms of how to set up the application and databases. Maybe you left it out because you're underestimating the expertise you now have (and assume everyone else would 'just know' how to do these things)? :)

app/app.rb
```diff
@@ -0,0 +1,51 @@
+ENV['RACK_ENV'] ||= 'development'
+require 'sinatra/base'
+require 'sinatra/flash'
+require './app/models/data_mapper_setup'
+require './app/models/peep'
+require './app/models/user'
+
```
EdwardAndress: Did you get onto the part of the challenge (during the week) where routes for different resources are split out into separate controllers?  This is called the modular pattern and, if you've not already read about it and tried it out, I recommend that you give it a shot.  We can talk about this if you like.

Also, have a quick read about what REST means and how to make your routes RESTful.  For example:

```ruby
# this
get 'posts/new' do
    <code>
end

# rather than this
get '/new' do
    <code>
end
```

app/app.rb
```diff
@@ -0,0 +1,51 @@
...
+    @peeps = Peep.all(:order => [:created_at.desc])
+    erb :index
+  end
+
+  get '/new' do
+    erb :new
+  end
+
+  post '/posts' do
+    Peep.first_or_create(post_text: params[:post_text], created_at: params[:created_at])
```
EdwardAndress: I'd expect a `post` route to always create something (if the params are valid), rather than try to find something in the DB.

```ruby
# this
Peep.create(<params>)
# rather than this
Peep.first_or_create(<params>)
```

spec/models/peep_spec.rb
```diff
@@ -0,0 +1,14 @@
+# require "./app/models/peep"
```
EdwardAndress: Is this commented out because you had problems with TimeCop?  If so, let me know on Slack and I'll provide some tips.

spec/features/posting_and_seeing_peep_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+  scenario 'See a message (peep) on chitter' do
+    sign_up(user_name: 'Homer', email: 'homerj@gmail.com', password: 'donuts', password_confirmation: 'donuts')
+    create_peep('Hello world!')
+    expect(page.status_code).to eq 200
+    within 'ul#posts' do
+    expect(page).to have_content('Hello world!')
+    end
+  end
+
+
```
EdwardAndress: I'd separate creating peeps from viewing the list of peeps here.  In order to do this, you can create two peeps directly, without using the UI (`Peep.create(<params>)`), then navigate to the page where the Peeps should be listed and ensure that the list is correct.

spec/features/posting_and_seeing_peep_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+    scenario "The peeps are in reverse chronological order" do
+      sign_up(user_name: 'Homer', email: 'homerj@gmail.com', password: 'donuts', password_confirmation: 'donuts')
+      create_peep('Hi!')
+      create_peep('Hello world!')
+      expect(page.status_code).to eq 200
+      within 'ul#posts' do
+        expect("Hi!").to appear_before('Hello world!')
```
EdwardAndress: Nice use of the orderly Gem!  I've not seen it before so I learned something really useful by reading your code :)

app/models/user.rb
```diff
@@ -0,0 +1,21 @@
+require 'bcrypt'
```
EdwardAndress: Why is there no user_spec.rb ?  Were you unsure about what to test?

app/models/peep.rb
```diff
@@ -0,0 +1,11 @@
+class Peep
+  include DataMapper::Resource
+
```
EdwardAndress: I'd include some validations here for things that must be in every Peep (and test for them in peep_spec.rb)

EdwardAndress:
Hi Giada,

This is generally a good attempt and I hope you enjoyed the challenge.

I think the main things to focus on from my comments are:

1.  RESTful routes
2. The modular pattern (splitting up routes into resource-specific controllers)
3. Try to test-drive everything :)

app/models/data_mapper_setup.rb
```diff
@@ -0,0 +1,22 @@
+# require 'data_mapper'
+# require 'dm-postgres-adapter'
+#
+# # require_relative './models/'
+# # require_relative './models/'
+# # require_relative './models/'
+#
+# DataMapper.setup(:default, ENV['DATABASE_URL'] || "postgres://localhost/bookmark_manager_#{ENV['RACK_ENV']}")
+# DataMapper.finalize
```
soph-g: Commented out code can make it hard to read later - consider removing code if no longer needed before opening a pull request, or checking out a different branch if you are experimenting and want to come back to it later.

soph-g:
Hi Giada,

You get two reviews for the price of one today :)  In addition to Eddie's comments I have left a small comment on the code. 

Also, I noticed that you asked a question about testing the CSS file in your review request, which I'm going to respond to with another question - does the CSS file need to be tested? Maybe have a google and see if you can find out what the standard practice is. I'm happy to chat about different types of testing and testing scope further if you'd like.

Sophie

README.md
GiadaSimonetti: I always underestimate the importance of README... :/

PR #847, opened 2018-02-18
Pull request - see read me for personal code review

marcusfgardiner:


app.rb
```diff
@@ -0,0 +1,42 @@
...
+  post '/cheets/new' do
+    Cheet.add_cheet(params[:title], params[:body])
+    redirect '/cheets'
+  end
+
+  get '/user/new' do
+    erb(:sign_up)
+  end
+
+  post '/user/new' do
```
soph-g: No need for the '/new' when posting a new user.

app.rb
```diff
@@ -0,0 +1,42 @@
...
+  get '/' do
+    redirect '/cheets'
+  end
+
+  get '/cheets' do
+    @cheets = Cheet.reverse_chronological_cheets
+    erb(:cheets)
+  end
+
+  post '/cheets/new' do
```
soph-g: No need for the '/new' when posting a new cheet

soph-g:
Hi Marcus,

I've had a read through your code and your README - there are too many questions for me to cover off here are my general notes, and I'll answer some of your less-googe-able questions below.

```
General Notes
* Really liked that you attempted DataMapper, getting relations to work would be a good learning exercise
* Couldn't see a clean script for the DB, but I might have missed it
* There's quite a bit of commented code in your repo, it's usually best to remove this or keep work in progress on a separate branch. For example it looked like you'd done some set up with the PG gem, which was then confusing when reading through the code.
* Generally your RESTful routing was good, I made a couple of comments where you'd used '/new' on a post that wasn't required.
* Good feature tests - for me this would be enough for integration testing as you are saving and retrieving from the database, but on a more complex project then an integration test might be recommended.
```
Comments on other questions:
- I wouldn't normally expect a Rakefile to be tested - it's usually possible to exclude files that wont be tested from the coverage reports.
- Sessions are generally used for managing logged in users. However, this is different to displaying user information within an app. For example, if you wanted to display the username with a Peep, then the Peep would be related to a User within the database, and this is where the data would come from.
- 'Partials with standard HTML conventions' means just make sure that any partials are being rendered in line with the normal HTML approach e.g. included within body tags.

Hope this is all clear, happy to discuss further if required.

Thanks,

Sophie

PR #862, opened 2018-02-18
Chitter Challenge and Personal Review - Josué Estévez

Jestfer:


README.md
```diff
@@ -1,3 +1,72 @@
...
+=================
+
+My approach
+------
+Pre-code:
+* First, I read through the User Stories and drew the Domain Models in a table in order to have a clear idea about the main objects and messages of the app.
+* Secondly, I drew a relational diagram focusing focusing on what happens after every step in the app. For example: the user opens the website, then peeps are displayed (no login required). The user can choose to sign-up and/or log-in to post new peeps.
+* Finally, I drew a quick mock-up of the pages, to have a better understanding on where to place the different sections, etc.
+
+Regarding the development process, I tried to stick by this flow as much as possible:
```
EdwardAndress: 👍

README.md
```diff
@@ -1,3 +1,72 @@
...
+------
+
+* I keep having trouble being led by TDD, especially when it comes to Unit Tests (I haven not managed to finish all of them in the way I wanted to). I tried to implement the one line of code at a time without breaking old tests when refactoring or adding new code, and for a while it seemed to help, but then I got lost in my code once again and I had to separate from the optimal flow for a bit.
+* I had not many difficulties with Capybara and Feature Tests though. I found myself testing a lot on the browser and changing many things in the Model. I also spent quite a few hours to get familiar with DataMapper, which is the first ORM I have come across so far.
+* Regarding databases, I am happy that I understand how relations between tables work and what kind of relations there are.
+* Finally, I have observed that in many occasions I struggle to access specific data in my code, or to differentiate among them, especially when dealing with hashes in objects, symbols... I feel like I rush a bit sometimes and start typing and testing the different options rather quickly instead of thinking about what I really want to do, which would save me some time definitely.
```
EdwardAndress: 👍

app.rb
```diff
@@ -0,0 +1,65 @@
...
+  get '/users/new' do
+    erb(:'users/new')
+  end
+
+  post '/users' do
+    user = User.create(
+      peeper: params['name'],
+      username: params['username'],
+      email: params['email'],
+      password: BCrypt::Password.create(params['password'])
```
EdwardAndress: Nice effort with the password encryption.  I'd probably refactor this to the User class as part of the create method though.

app.rb
```diff
@@ -0,0 +1,65 @@
...
+    )
+
+    if user.errors.size.positive?
```
EdwardAndress: does `if user.errors` work?

EdwardAndress: another pattern, which you'll see used quite often, is to separate instantiation from entering a new record in the DB (`create` combines both).  Something like this (example taken from a Rails controller):
```ruby
def create
    @client = Client.new(params[:client])
    if @client.save
      redirect_to @client
    else
      # This line overrides the default rendering behavior, which
      # would have been to render the "create" view.
      render "new"
    end
  end
```

README.md
EdwardAndress: Nice reflection

Rakefile
```diff
@@ -1,7 +1,63 @@
...
+
 if ENV['RACK_ENV'] != 'production'
   require 'rspec/core/rake_task'
   
   RSpec::Core::RakeTask.new :spec
   
   task default: [:spec]
 end
+
+task :setup do
```
EdwardAndress: This all looks great.  One way to improve it slightly would be to separate database creation from table creation.  Then you could have tasks like `create_db`, `drop_db`, `create_tables` and `drop_tables` which give more precise control. For example, you could drop the tables without having to drop and recreate the whole db.  You could also have some safeguards on `drop_db` so that you never accidentally drop a db which has tables.

app.rb
```diff
@@ -0,0 +1,65 @@
...
+      flash.next[:errors] = user.errors.first # Next view flash is taken!
+      redirect '/users/new'
+    end
+
+    session[:id] = user.id
+    redirect '/'
+  end
+
+  get '/sessions/new' do
```
EdwardAndress: Nice work making RESTful routes for sessions!

spec/database_connection_spec.rb
```diff
@@ -0,0 +1,17 @@
+require 'datamapper'
+
+describe DataMapper do
+  describe '.setup' do
+    it 'sets up connection to test db' do
+      expect(DataMapper).to receive(:setup).with(:default, 'postgres://localhost/chitter_test')
+
+      DataMapper.setup(:default, 'postgres://localhost/chitter_test')
```
EdwardAndress: This is not really testing anything and you don't need to test the DataMapper methods.  If you have some other methods that delegate to DataMapper then you could use the same expectation there and a different method call.

If this is only ever used in your Rake task, you can either try to test the Rake task directly or assume that it all works if your other tests pass (and the whole app doesn't "blow up")

spec/features/sign_up_spec.rb
```diff
@@ -0,0 +1,13 @@
+feature 'signing up' do
+  scenario 'user can sign up' do
+    sign_up
+
+    expect(page).to have_content 'Welcome to Chitter, MarcoAsensio!'
+  end
+
+  scenario 'same email cannot be used twice' do
```
EdwardAndress: Does this rely on the other scenario being run first?  This is normally to be avoided.

spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'posting peeps' do
```
EdwardAndress: This scenario doesn't involve the user signing in but you're testing that "a registered user can post peeps".

spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,10 @@
...
+  scenario 'registered users can post peeps' do
+    visit '/peeps/new'
+
+    fill_in('new_peep', with: 'Betis - Real Madrid today!')
+    click_button 'Publish peep'
+
+    expect(page).to have_content 'Betis - Real Madrid today!'
+  end
+end
```
EdwardAndress: The counterpoint to this is that "an unregistered / signed out user cannot post peeps".  This is also worth testing.

spec/peep_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+  describe '.all' do
+    it 'returns all peeps' do
+      peeps = Peep.all
+
+      expect(peeps.count).to eq 2
+      expect(peeps.first['peep']).to eq '¡Día de partido! ¡Vamos Real, hasta el final!'
+      expect(peeps[1]['peep']).to eq 'El bicho scores a hattrick!'
+    end
+  end
+end
```
EdwardAndress: It would be clearer to make the set up part of this file.  You could, for example, have a before block which inserts entries into the db.  Then do `expect(Peep.all).to match_array ["el coche es antiguo", "donde esta Steve McManaman"]`.  The great thing about `match_array` is that it doesn't matter what order the items are in.

spec/user_spec.rb
```diff
@@ -0,0 +1,24 @@
+describe User do
+  describe '.create' do
+    it 'creates a new user' do
+      user = User.create(peeper: 'Lupita', username: 'Lupi', email: 'lupitareyes@example.es', password: 'lupitrona')
+
+      expect(user.id).to eq 3
```
EdwardAndress: This feels brittle (it will break easily) because you're testing for a specific ID.  Instead you could do something like
```ruby
expect { User.create(params) }.to change { User.all.count }.by 1
```

This would test that something is entered into the DB and not be effected by what's already in the DB

spec/user_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    end
+
+    it 'secures the password with BCrypt' do
+      pass = BCrypt::Password.create('lumber')
+
+      u = User.new(peeper: 'Lumberman', username: 'Lumbi', email: 'lumberman@example.es', password: pass)
+
+      expect(u.save).to eq true
+      expect(User.get(u.id).password).to eq pass
```
EdwardAndress: Given that you're passing an encrypted password to `User#new` in the controller, this is more of an integration test, than part of the `user_spec`.  If, however, you were to move encryption into the `User` class this test would stay here (implemented slightly differently).

EdwardAndress:
Hey Josu!  Nice work on the whole.  It's cool that you used DataMapper even though we didn't use it during the week and also that you got to the user sign in / sign up stuff.  Really great stuff.

There's a few comments about testing and use of BCrypt.  I think those are the main things to focus on here.  Let me know if you have any questions.

app.rb
Jestfer: Hi @EdwardAndress ! Thank you for your thorough code review, finally I have some time to go over things more focused. 

In order:
1) Yes, exactly, `user.errors` works fine.

2) Do you think we can discuss your second comment on this section at some point next week? I understand the difference between **Create** and **New** (which requires save to store it into the DB apart from creating a new instance of an object).

I would like to address especially the advantage of separating these cases in this specific example. Thanks in advance!

Rakefile
Jestfer: Great advice, Eddie, thanks!

spec/features/sign_up_spec.rb
Jestfer: So, I imagine then the right way to test it would be to sign-up again (twice) with different data, right? Either that or to truncate the table in the test DB after running each test, right? Any preferences on better practices here? Thanks in advance!

spec/features/post_peep_spec.rb
Jestfer: Oh, you're right here, Eddie. Actually this surprises me... I tried to post a peep manually without being authenticated, and it does not allow me (expected) as it depends on the User object to do it... 
My guess is that after populating the table with a User (when running the setup_test_db Rake task), it authenticates the user automatically! And that is why it allows the test to post that peep and sees the content on the page. Does it make sense? I am actually quite surprise with this case, happy to go over it together next week too! Thanks.

spec/peep_spec.rb
Jestfer: Awesome matcher! I did not know it existed!

What about this?
```
describe Peep do
  before(:all) do
    @peeps = []
    peep_objects = Peep.all
    
    peep_objects.each { |peep| @peeps << peep.peep }
  end

  describe '.all' do
    it 'returns all peeps' do
      expect(@peeps).to match_array ['El bicho scores a hattrick!', '¡Día de partido! ¡Vamos Real, hasta el final!']
    end
  end
end
```

spec/user_spec.rb
Jestfer: Hey Eddie! Thanks for this comment, I would like to review this point with you as well.

I have implemented the encryption within the model (User class), like this:

```
class User
  include DataMapper::Resource
  validates_uniqueness_of :email, :username

  has n, :peeps

  property :id,         Serial
  property :peeper,     String
  property :username,   String
  property :email,      String
  property :password,   String, :length => 60

  before :save do
    BCrypt::Password.create(password)
  end
end
```

Now my question is, in the mentioned test I was checking whether or not after encrypting the password it was still the same in the expectation. Now, how should I test a dependancy as a Unit Test being BCrypt the dependancy? The only refactoring I can think of is just checking the same, but removing the BCrypt part, like:

```
    it 'secures the password with BCrypt' do
      u = User.new(peeper: 'Lumberman', username: 'Lumbi', email: 'lumberman@example.es', password: 'lumber')

      expect(u.save).to eq true
      expect(User.get(u.id).password).to eq 'lumber'
    end
```

This is one of the points I am more confused about my tests I believe. Thanks again in advance!

PR #879, opened 2018-03-18
Chitter Challenge - Julian Ng - "Phoenix"

julesnuggy:


neoeno:
Hi Julian,

You've reached pretty far here — which is really brilliant. There are a few things that are holding this back from being a good codebase.

The two most important:

## Testing

The unit tests are quite unreliable as they often rely on hard-coded record IDs which are unlikely to be the same on anyone else's computer (indeed, they were not on mine), just due to the order things have been put into the database.

You are also relying on `inspect` and string comparison very heavily, when you could have requested the specific record and extracted the field. This would definitely raise a few eyebrows from a sr developer and I'd suggest dropping the habit.

That said, your feature tests look pretty good, aside from relying on those record `id`s in the field names. Happy to see those there.

## Linting

There are many linting warnings here. Based on the work you've done, they are definitely within your capability to resolve! These things are important for a developer, as they are part of having professional consideration for your team. I would recommend fixing them all before every commit.

There are a couple of other things that stem from this. For instance — your rake task currently fails due to your using `Db_Connection`. By using the idiomatic notation, you cut yourself off from mistakes of this kind. I'd suggest `DatabaseConnection`.

All this said — you've clearly experimented and driven pretty hard with this one. That's a great thing that will take you a long way. The reason I'm direct about it now is these bad habits can be career-limiting and it would be a real shame for your skill and drive not to take you as far as it could. Make it your highest priority!

PR #930, opened 2018-06-10
Chitter challenge

simone-smith:


Rakefile
```diff
@@ -1,7 +1,54 @@
+require 'pg'
```
samjones1001: Just to confirm what I think we found out yesterday morning - this all looks fine.  Other people should be able to use your Rake tasks to set up their DB environment and run your app.

Rakefile
```diff
@@ -1,7 +1,54 @@
...
   require 'rspec/core/rake_task'
-  
+
   RSpec::Core::RakeTask.new :spec
-  
+
   task default: [:spec]
 end
+
+task :setup do
```
samjones1001: You mentioned in your self-review that you would like to have set up your database tables so that there was a one-to-many relationship between users and peeps.  Based on what you've already done here, its actually a fairly simple change.

I'd suggest you take a look at this to see what this would look like that see if you can research what the appropriate addition to your SQL statement would look like:

http://www.databaseprimer.com/pages/relationship_1tox/

I'm probably also going to run a session on table relationships at some point this week,  so it might be interesting for you to come along to that.

app.rb
```diff
@@ -0,0 +1,46 @@
...
+class Chitter < Sinatra::Base
+  register Sinatra::Flash
+  enable :sessions
+
+  get '/' do
+    @user = User.find(session[:user_id])
+    erb :index
+  end
+
+  post '/' do
```
samjones1001: I'm not clear what the purpose of this route it - same for the post /users/new route on line 30

lib/peep.rb
```diff
@@ -0,0 +1,47 @@
...
+  attr_reader :id, :peep, :user, :created_at
+
+  def initialize(id, peep, user, created_at)
+    @id = id
+    @peep = peep
+    @user = user
+    @created_at = created_at
+  end
+
+  def self.post(options)
```
samjones1001: This is a minor issue, but conventionally the method for adding resources to a db would be called create and the method for displaying all instances of a resource (on line 23) would be called all.

This is just convention, and the names you've given in place of these are perfectly clear, but generally speaking following convention helps to reveal your intentions which is always a good thing.

lib/user.rb
```diff
@@ -0,0 +1,50 @@
...
+
+  def self.find(id)
+    return nil unless id
+    connection = database_connection
+    result = connection.exec("SELECT * FROM users WHERE id = '#{id}'")
+    User.new(result[0]['id'], result[0]['email'], result[0]['password'],\
+    result[0]['name'], result[0]['username'])
+  end
+
+  def self.database_connection
```
samjones1001: Given that this method is repeated in both your user and peep model files, you might want to refactor it out into a separate script.

lib/peep.rb
```diff
@@ -0,0 +1,47 @@
+require 'pg'
```
samjones1001: Are there any tests for this class?  I can't see any 😬

PR #936, opened 2018-06-10
Chitter Web App

CLTPayne:


lib/peep_message.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM peeps")
+    result.to_a.reverse.map do |peep|
+      PeepMessage.new(peep['id'], peep['message'], peep['name'],
+                      peep['username'], peep['time'])
+    end
+  end
+
+  def self.create(message, name, username, time)
```
samjones1001: What was the thinking behind splitting your create and insert_into_database methods into two.  I don't think it's wrong as such, but conventionally I would expect the create method to be responsible for writing to the database.  By using the SQL RETURNING keyword (as you do on line 43), you can achieve the same result of returning a Ruby object.

Some thing like:

```ruby
def self.create(message, name, username, time)
  result = DatabaseConnection.query(
+      "INSERT INTO peeps (message, name, username, time)
+      VALUES ('#{message}', '#{name}', '#{username}', '#{time}')
+      RETURNING peep_id, message, name, username, time"
+ )
   PeepMessage.new(result.first['id'], peep_message = result.first['message'], peep_name = result.first['name'], peep_username = result.first['username'], peep_time = result.first['time'])
end
```

lib/sign_up.rb
```diff
@@ -0,0 +1,56 @@
+require 'bcrypt'
+require_relative './database_connection'
+
+class ChitterSignUp
```
samjones1001: I feel like this is really a User class.  It seems to deal with the creation and saving of users, so I think that would be a clearer name.

lib/sign_up.rb
```diff
@@ -0,0 +1,56 @@
...
+    @username = username
+    @email = email
+    @password = password
+  end
+
+  def ==(other)
+    @user_id == other.user_id
+  end
+
+  def self.new_user(name, username, email, password)
```
samjones1001: This probably ought to be self.create - again, this is only a convention, but generally speaking conventions should be followed!

PR #945, opened 2018-07-09
Chitter challenge - Cornelis Heuperman

heuperman:


spec/units/user_spec.rb
```diff
@@ -0,0 +1,50 @@
+require 'user'
+
+describe User do
+  describe '.create' do
+    it 'creates a new user' do
+      user = create_new_user
```
mdsomerfield: Fine to use a test helper 'create_new_user' when you are just creating a user to support testing of some other code - but if the thing that I am testing is the code in the test helper, I would want the code for that to appear in the test itself, otherwise I have to look in another file to see what I am testing.

Also, it feels like this is testing state not behaviour. Have you come across that idea yet? Basically, there is no point in testing the initial state of an object because you are implicitly testing this by testing behaviour elsewhere. For example, the test 'returns a user given a correct username and password' will fail if either the id or the password is not initialised correctly, which makes this test redundant.

lib/blab.rb
```diff
@@ -0,0 +1,36 @@
...
+      Blab.new(blab['id'], blab['timestamp'], blab['content'], blab['user_id'])
+    end
+  end
+
+  def self.create(content, user_id)
+    result = DatabaseConnection.query("INSERT INTO blabs (content, user_id)" +
+      " VALUES ('#{content}', '#{user_id}')" +
+      " RETURNING id, timestamp, content, user_id;")
+    Blab.new(
+      result.first['id'],
```
mdsomerfield: Efficiency nitpick - if you do:

```
blab_data = result.first

Blab.new(
  blab_data['id'], 
  blab_data['timestamp'] 
  ... etc 
)
```

It both reads better (as in, the 'result' is named more clearly) and is more efficient (because every time you call result.first you are doing a lookup, and even though there is probably only one result there is still a minor cost to this). The efficiency thing doesn't matter in this case, but it's still a good habit to be aware of that as there are cases when it does matter.

lib/blab.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM blabs;")
+    result.map do |blab|
+      Blab.new(blab['id'], blab['timestamp'], blab['content'], blab['user_id'])
+    end
+  end
+
+  def self.create(content, user_id)
+    result = DatabaseConnection.query("INSERT INTO blabs (content, user_id)" +
```
mdsomerfield: You have similar SQL (for this and other methods) in your user class. Something a good ORM does is abstract this away. Think about whether you could write a 'DatabaseQueries' class such that here you can call:

```
blabs_data = DatabaseQueries.insert('blabs', { 'content' => content, 'user_id' => user_id })
```
Or, more verbose but arguably better:
```
insert_query = InsertQuery.new('blab');
insert_query.add_field('content', content);
insert_query.add_field('user_id', user_id);
blab_data = insert_query.execute();
```
Why is the second one arguably better? Because as well as abstracting away any need to know SQL syntax the user has a clear and instructive interface. Either version though makes your code more DRY and better maintains single responsibility principle (I would argue that the purpose of this class is to model a Blab, not to build SQL).

PR #960, opened 2018-09-02
Esam's Chitter Challenge

EsamAl-Dabagh:


lib/peep.rb
```diff
@@ -0,0 +1,45 @@
...
+    select_database
+
+    peeps = DatabaseConnection.query("SELECT * FROM peeps ORDER BY id DESC")
+
+    peeps.map do |peep|
+      Peep.new(peep["id"], peep["body"], peep["posted_at"], peep["user_id"])
+    end
+  end
+
+  def self.add(peep, user_id)
```
mdsomerfield: Minor: I would use 'body' or 'text' or 'peep_text' (etc etc) here, instead of just peep. My expectation when I see the variable peep here is that it is an instead of Peep, not just the body.

Similarly, in the method above, I might be tempted to use peeps_data and peeps_data.map do |peep_data|, just to disambiguate the data from the model.

lib/peep.rb
```diff
@@ -0,0 +1,45 @@
...
+    select_database
+
+    result = DatabaseConnection.query(
+                              "INSERT INTO peeps (body, user_id) 
+                              VALUES ('#{peep}', '#{user_id}')
```
mdsomerfield: In the group code review, we talked about escaping values - injecting values directly into your SQL leaves you open to SQL injection attacks. What is somebody peeped:

```
'); DROP TABLE Peeps;
```

lib/user.rb
```diff
@@ -0,0 +1,64 @@
+require "./lib/database_connection"
+require "./lib/select_database"
+require "bcrypt"
+
+class User
+
+  attr_reader :id, :email, :username, :name, :password
```
mdsomerfield: I don't think you want to make password accessible outside the user object. I don't think you're using it anywhere for a start (I don't see any authentication) but usually we would want to keep this private and have an `authenticate` method which validates a users password without exposing the information.

In general, try not to expose information until you need it - attr_readers should only be added where other classes actually need to read state, rather than automatically when you create a new class.

lib/user.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    id = result[0]["id"]
+
+    User.new(id, email, username, name, password)  
+  end
+
+  def self.find(id)
+    return nil unless id
+
+    result = DatabaseConnection.query("SELECT * FROM users WHERE id = '#{id}';")
```
mdsomerfield: What if we don't find the result? How can we handle that error case?

PR #962, opened 2018-09-02
Mathilde - Chitter challenge (partial)

ChocolatineMathou:


lib/peep.rb
```diff
@@ -0,0 +1,37 @@
+require 'pg'
+
+class Peep
+
+  def self.all
+    if ENV['RACK_ENV'] == 'test'
```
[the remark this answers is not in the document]
mdsomerfield: Also, single responsibility principle: I feel like this should be a separate method get_connection, or event a separate class, e.g.:
``` 
connection = Connection.new_connection();
```
or
``` 
connection = Database.connect();
```
I notice that you've repeated this code below, which suggests that it could be refactored to avoid duplication.

spec/profile_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+  describe '.all' do
+    it 'retrieves the usernames from the database' do
+      profiles = Profile.all
+      expect(profiles).to include('chocolatine')
+    end
+  end
+
+  describe '.create' do
+    it 'creates a new user' do
+      Profile.create(username: 'chocolatine', name: 'mathilde', email: 'mathilde@email.com', password: '1234')
```
mdsomerfield: There's no expectation here

spec/profile_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'profile'
+
+describe Profile do
+
+  describe '.all' do
+    it 'retrieves the usernames from the database' do
+      profiles = Profile.all
+      expect(profiles).to include('chocolatine')
```
mdsomerfield: Where does chocolatine come from? The test below? Your tests should be independent of each other.

PR #963, opened 2018-09-02
Chitter Challenge!

LondonJim:


lib/users.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def self.create(name, user_name, email, password)
+    database_connect.exec("INSERT INTO users (name, user_name, email, password) VALUES('#{name}', '#{user_name}', '#{email}', '#{password}') RETURNING id, name, user_name;")
+  end
+
+  def self.select_user(user_name)
+    result = database_connect.exec("SELECT id, name, user_name FROM users")
+    result.map do |user|
+      return Users.new(id: user['id'], name: user['name'], user_name: user['user_name']) if user.values[2] == user_name
```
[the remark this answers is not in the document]
mdsomerfield: It's worth paying attention to this comments - and trying to understand why they are made! Here's a discussion on line length, I'll leave you to research the others...

https://news.ycombinator.com/item?id=5436993

lib/email_notice.rb
```diff
@@ -0,0 +1,36 @@
...
+
+class EmailNotice
+
+  extend DatabaseConnect
+
+  def self.check_peep(peep, current_user_name)
+    result = database_connect.exec("SELECT user_name, email FROM users;")
+    result.each do |user|
+      if peep.include? "##{user.values[0]}"
+        return send_email_alert(user.values[1], current_user_name, peep) # tagged users email and the user name of the tagger
```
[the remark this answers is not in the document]
mdsomerfield: It's worth paying attention to this comments and trying to understand why they are made. Here's a link to a discussion on 80 char limit: https://news.ycombinator.com/item?id=5436993

lib/email_notice.rb
```diff
@@ -0,0 +1,36 @@
...
+      end
+    end
+    nil
+  end
+
+  def self.send_email_alert(user_email, current_user_name, peep)
```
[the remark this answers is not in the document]
mdsomerfield: Ok, re: previous comment about houndci, there are exceptions :)

lib/email_notice.rb
```diff
@@ -0,0 +1,36 @@
+require "dotenv"
+require "pony"
+require_relative './database_connect'
+
+class EmailNotice
+
+  extend DatabaseConnect
+
+  def self.check_peep(peep, current_user_name)
```
mdsomerfield: It's not really clear what you're checking here. I had to read this method a couple of times to work out what it's doing; I'd rather know from the method name.

lib/email_notice.rb
```diff
@@ -0,0 +1,36 @@
...
+    result = database_connect.exec("SELECT user_name, email FROM users;")
+    result.each do |user|
```
mdsomerfield: Rather than each, since you return on the first match, you might consider using 'any?' rather than 'each' - it makes it clearer what you're trying to achieve

app.rb
```diff
@@ -0,0 +1,80 @@
...
+    @current_user = session[:current_user]
+    @peeps = Peeps.all
+    erb :peep_feed
+  end
+
+  get '/post_peep' do
+    erb :post_peep
+  end
+
+  post '/post_peep' do
```
mdsomerfield: Why is your route called 'post_peep'? Surely it should be just be post 'peep'? You already have post in the verb, it doesn't need to be in the route.

lib/input_checkers.rb
```diff
@@ -0,0 +1,26 @@
...
+
+class InputCheckers
+
+  extend DatabaseConnect
+  include BCrypt
+
+  def self.unique_input_checker(option, checking_key)
+    result = database_connect.exec("SELECT #{checking_key} FROM users;")
+    result.map do |items|
+      return true if items.values[0] == option
```
mdsomerfield: Again, you could use 'any?' here. The use of 'map' and 'return' feels a bit awkward and its harder to discern intent.

lib/peep_manager.rb
```diff
@@ -0,0 +1,32 @@
+require_relative './database_connect'
+require 'pg'
+
+class Peeps
+
+  extend DatabaseConnect
```
mdsomerfield: I like that you've pushed your DatabaseConnect code into a class and then included it into your model classes in this way.

spec/input_checkers_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+require 'pg'
+
+include BCrypt
+
+describe InputCheckers do
+
+  before(:each) do
+    connection = PG.connect(dbname: 'chitter_test')
+    password = Password.create("testpassword")
+    connection.exec("INSERT INTO users (name, user_name, email, password) VALUES('test name', 'test user name', 'test@email.com', $txt$#{password}$txt$);")
```
mdsomerfield: You're creating this user before each test... I don't see anywhere where you're deleting it. Or is that part of a helper somewhere?

PR #970, opened 2018-09-03
Pull request for Chitter weekend challenge

freddie1221:
* This meets all the 'straight up' requirements
* All tests are green, test coverage is 100%
* This uses a PostgreSQL database and Sinatra web app


Lib/messages.rb
```diff
@@ -0,0 +1,42 @@
+require 'pg'
```
samjones1001: You mentioned in your self reflection that you felt as though this class had more than one responsibility (inserting into a DB, retrieving from a DB etc.), but I'd argue that it's actually fine.

I'd describe the responsibility of the class as managing Messages. It's sensible to keep all of the behaviour for Messages in one place. The alternative would presumably be to have object responsible for the saving of any type of resource, an object for the retrieval of any type of resource etc, which feels like a strange way of organising things.

Lib/messages.rb
```diff
@@ -0,0 +1,42 @@
...
+    @created_at = created_at
+    @name = name
+  end
+
+  def self.add(content, name)
+    database_connection.exec("INSERT INTO messages(content,name) 
+      VALUES('#{content}','#{name}')")
+  end
+
+  def self.show
```
samjones1001: conventionally, a method like this which retrieves every example of a particular resource would be called .all

spec/features/messages_feature_spec.rb
```diff
@@ -0,0 +1,29 @@
+feature 'messages' do
```
samjones1001: There is probably a small refactor you could do to these tests to keep them DRY. Some of the set up steps are very similar across them and could be extracted to a helper method.  It's a very small thing though.

PR #976, opened 2018-09-30
chitter-challenge 

m-rcd:


app/app.rb
```diff
@@ -0,0 +1,27 @@
...
+require_relative './helpers'
+
+class Chitter < Sinatra::Base
+  enable :sessions, :method_override
+  set :session_secret, 'super secret'
+  register Sinatra::Flash
+  register Sinatra::Partial
+
+  require_relative './models/send_tag_alert'
+  require_relative 'controllers/login'
```
mdsomerfield: It's great that you've broken things out into controllers; I'm not sure you need quite so many controllers however. I think login/logout/signup are all related to auth & accounts, so I'd be tempted to put those in a single AccountController, while peeping and replying are both related to Peeps, so I would put those in a PeepsController. Controllers in general should be organised around concerns, rather than individual routes.

app/controllers/peeping.rb
```diff
@@ -0,0 +1,15 @@
+class Chitter < Sinatra::Base
+  get '/' do
+    @peeps = Peep.all(:order => [:created_at.desc])
+    erb :index
+  end
+
+  post '/peeps' do
+    @user = current_user
+    Peep.create(message: params[:peep], user_id: @user.id)
+    User.all.each do |user|
```
mdsomerfield: The way I read this code, you are pulling all users from the database, and then if the text of the peep contains their username they are being sent an alert - is that correct?

This would probably be a problem if you had lots of users in your database, because you'd have to get each record, transform it into a Ruby instance, and then do a fuzzy text match. Better would be if you could identify user names from the text, e.g. with an @ symbol, and then do a User.find(:name => username) for each identified name.

app/models/user.rb
```diff
@@ -0,0 +1,31 @@
+require 'bcrypt'
+require_relative './peep'
+require_relative './reply'
+
+class User
+
+  attr_reader :password
```
mdsomerfield: Why do you need attribute readers/accessors for password and password_confirmation?

PR #979, opened 2018-09-30
Chitter Challenge

NadiaAiraf:


app.rb
```diff
@@ -0,0 +1,52 @@
...
+                user_id: User.first("user" => "#{session[:user]}")[:id])
+    redirect '/'
+  end
+
+  get '/register' do
+    erb :register
+  end
+
+  post '/register' do
+    if Login.user_exists?(params[:username])
```
mdsomerfield: Better here would be to have a User.register class method which tries to create a user and throws an exception if the user already exists, which is caught and handled here. Why is that better? Because it prevents anybody from misusing it. With this setup, another developer trying to write registration code would have to remember to check if the user_exists first, and so there's potential to introduce a bug. If you force them to use a method that validates and fails if validation fails then that eliminates misunderstanding or oversight.

mdsomerfield: This is also an example of the 'tell, don't ask' principle.

PR #980, opened 2018-09-30
Chitter-challenge

FlorenceHapgood:


lib/users.rb
```diff
@@ -0,0 +1,26 @@
...
+  include DataMapper::Resource
+
+  property :id,         Serial
+  property :name,       String
+  property :username,   String
+  property :email,      String
+  property :password,   String
+end
+
+DataMapper.finalize
```
mdsomerfield: It feels like this code and the lines below belong in a more generic place, app.rb or a separate db_setup.rb - particularly as it is duplicated above in your chitter.rb.

lib/users.rb
```diff
@@ -0,0 +1,26 @@
+require 'rubygems'
+require "data_mapper"
+
+if ENV['ENVIRONMENT'] == 'test'
```
mdsomerfield: As per comment below (for some reason I am reading upwards), this is duplicate code from your chitter.rb, so probably belongs in a separate connection class.

app.rb
```diff
@@ -0,0 +1,59 @@
...
+  enable :sessions
+
+  register Sinatra::Flash
+
+  get '/' do
+    erb :sign_in
+  end
+
+  post '/sign_in' do
+    session[:username] = params[:username]
```
mdsomerfield: This feels like quite a long method - worth thinking about ways you can refactor it to make it more readable, either by moving code into named methods or thinking about whether some of the code actually belongs in another class.  For example, checking whether the user is duplicate or not is probably not the responsibility of this class, you could put that in a Users.register class method, and as part of that validate whether the information is unique.

spec/chitter_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+describe Chitter do
+
+  describe "#all" do
+    it "Returns a chitter object, which we can extract a message instance from" do
+      @test = Chitter.new(:message => "Test 1", :posted_by => "Florence", :created_at => Time.now)
+      @test.save
+      array = Chitter.all
+      chitter_object = array[@test.id - 1]
+      expect(chitter_object.message).to eq("Test 1")
```
mdsomerfield: So, I totally understand the urge to write these tests - however, they're not really necessary! You can adequately test the database component of your app with feature tests, writing unit tests like this you're really just testing DataMapper itself, which isn't your code.

PR #992, opened 2018-10-01
Weekend Chitter Challenge (ruby version 2.5.0)

JonathanAndrews:


app.rb
```diff
@@ -0,0 +1,76 @@
...
+  enable :sessions
+  register Sinatra::Flash
+
+  get '/' do
+    'Hello world'
+  end
+
+  post '/users' do
+    string = "Sorry, that email or username has already been used."
+    unless (user = User.make(email: params["email"], password: params["password"], username: params["username"], name: params["name"]))
```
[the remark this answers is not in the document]
mdsomerfield: It's worth paying attention to these comments. Line length of 80 or 100 characters max are conventions designed to make code easier to read. For example, as I review this, I can't actually read this whole line without scrolling right.

app.rb
```diff
@@ -0,0 +1,76 @@
...
+  get '/users/new' do
+    erb :sign_up
+  end
+
+  get '/users/sign_in' do
+    erb :sign_in
+  end
+
+  post '/users/sign_in' do
+    string = "Sorry, either your username or password were incorrect"
```
mdsomerfield: Avoid names like 'string'. The name of a variable should reflect the intent - the intent here is error_message or validation_error or something like that. 

That said, do you even need to assign this to a variable in the first place? `flash[:error] = "Sorry, either your username or password were incorrect"` below is fine.

app.rb
```diff
@@ -0,0 +1,76 @@
...
+  end
+
+  get '/users/:id' do
+    @user = User.get(params[:id])
+    erb :users
+  end
+
+  get '/peeps' do
+    @current_user = User.current_user
+    @user_class = User
```
mdsomerfield: I'm not sure why you need line 60.

mdsomerfield: Ah, I see now - you are calling `@user_class.get(peep.user_id).username` in your view. This is generally bad practice, for two reasons:

1) It's a lot clearer when you define all the vars for view in advance of rendering the view, as it means that another developer can see exactly what information is needed by the view in one place
2) Fetching the data from User actually involves a call to the database. You are making multiple calls to user_class.get(peep.user_id) in your view, which means multiple calls to the database, which adds load to your system. Instead, if you have a relationship, you should be able to do peep.user.username & peep.user.name. This will result in fewer calls to the database. To understand this fully, it's worth reading about lazy loading, eager loading, and how Datamapper uses strategic eager loading to reduce database load.

lib/user.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def self.sign_out
+    @current_user = nil
+  end
+
+  def self.current_user
+    @current_user
+  end
+
+  def self.new_email?(email)
```
mdsomerfield: I feel like some of these methods should be private. Also, though, do you need them all? Each one makes another call to the database. If you just try to find a user with the username and matching encrypted password, then either there is one or there isn't, so you have one call to the database which results in a user being generated or an error if there isn't a match.

views/chitter_homepage.erb
```diff
@@ -0,0 +1,24 @@
+<h1>Chitter Homepage</h1>
+<% @peeps.each do |peep| %>
+  <ul>
+    <li><%= @user_class.get(peep.user_id).username %></li>
+    <li><%= @user_class.get(peep.user_id).name %></li>
```
mdsomerfield: As per comment above, avoid making database class or including complex logic in your views

PR #994, opened 2018-10-01
Chitter challenge

djstephenson7:


lib/cheets.rb
```diff
@@ -0,0 +1,20 @@
...
+DataMapper.setup(:default, 'postgres://localhost/chitter')
+
+class Cheets
+
+  include DataMapper::Resource
+  property :id,         Serial
+  property :body,       String
+  property :created_at, DateTime
+
+  DataMapper.finalize
```
mdsomerfield: As discussed in the group session, this code probably belongs in a separate class referenced from your app.rb, or possibly in a rake file.

app.rb
```diff
@@ -0,0 +1,18 @@
+require 'sinatra/base'
+require './lib/cheets'
+
+class Chitter < Sinatra::Base
+  enable :sessions
```
mdsomerfield: You've enabled sessions but you're not using them anywhere as far as I can see.

lib/cheets.rb
```diff
@@ -0,0 +1,20 @@
...
+  DataMapper.auto_migrate!
+  DataMapper.auto_upgrade!
+
+  Cheets.create body: 'Test Cheet One', created_at: Time.now
```
mdsomerfield: This feels like test code, and should really be in a spec or a spec_helper - test and production code should be kept separately.

spec/app_spec.rb
```diff
@@ -0,0 +1,23 @@
+require './app'
+
+describe Chitter do
+  feature 'Test' do
+    scenario "stores a list of test cheets" do
+      visit('/')
+
+      expect(page).to have_content('Test Cheet One')
```
mdsomerfield: Related to the comment about the location of test code, if I read this test in isolation I have no idea where Test Cheet One etc come from. It's much clearer to another dev if the test setup is located in the same place as the test, as then they can see precisely what's going on.

spec/cheets_spec.rb
```diff
@@ -0,0 +1,18 @@
+require './lib/cheets'
+
+describe Cheets do
+
+  it { should have_property   :id         }
```
mdsomerfield: As we discussed in the group session, you probably don't need to be testing these things. When we're writing declarative code without any behaviour there's nothing really there for a unit test; and the model itself should be covered by a feature test so you still have the coverage.

PR #999, opened 2018-10-01
Pull request

afaraone:


afaraone:
I have edited travis.yml to create my user and database, and now it is no longer carry out a test. Have I done this correctly?

PR #1006, opened 2018-10-28
Chitter Challenge PR

MHUS25:


lib/peep.rb
```diff
@@ -0,0 +1,26 @@
...
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM peeps")
+    result.map do |peep|
+      Peep.new(id: peep['id'], message: peep['message'], peep_timestamp: peep['peep_timestamp'])
+    end
+  end
+
+  def self.create(message:, peep_timestamp:)
+    result = DatabaseConnection.query("INSERT INTO peeps (message, peep_timestamp) VALUES('#{message}', '#{peep_timestamp.strftime("%Y-%m-%d %k:%M")}') RETURNING id, message, peep_timestamp;")
```
toddpla: another option is to let PostgreSQL deal with the timestamp in the back end

ALTER TABLE Peeps ADD COLUMN created_at TIMESTAMP DEFAULT NOW()

README.md
```diff
@@ -1,21 +1,20 @@
-Chitter Challenge
-=================
+# Chitter
```
soph-g: Good detailed README

PR #1010, opened 2018-10-28
Chitter Challenge - Aimee Craig

aimeesophia:


app.rb
```diff
@@ -0,0 +1,46 @@
+require 'sinatra/base'
+require './lib/peep'
+require './lib/user'
+require './database_connection_setup'
+
+class Chitter < Sinatra::Base
```
samjones1001: Generally your controller looks good.  It's not doing any logical heavy lifting and all your routes look pretty concise.

One thing you might want to take a look at is RESTful routing - it's a set of conventions for how we should name routes in a web application and is a really easy wind in terms of making your code look of a professional standard.

lib/database_connection.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
+
+class DatabaseConnection
```
samjones1001: Good that you've extracted this out into its own class

lib/peep.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM peeps;")
+    result.map { |peep|
+      Peep.new(id: peep['id'], message: peep['message'], time: peep['time'])
+    }
+  end
+
+  def self.post(message)
```
samjones1001: This is a small point, but conventionally this method would be called create rather than post.

PR #1017, opened 2018-10-28
Chitter Challenge

melissasedgwick:


lib/database_connection.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
+
+class DatabaseConnection
```
samjones1001: Good that you extracted this out into a separate object.

lib/message.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'database_connection'
+
+class Message
+
+  attr_reader :id, :content, :time, :user_name, :user_handle
+
+  def initialize(id:, content:, time:, user_name:, user_handle:)
```
samjones1001: I like the use of named parameters here.  IT's generally a good idea for any method with a long list of arguments.

spec/features/sign_in_spec.rb
```diff
@@ -0,0 +1,45 @@
+feature 'Sign In' do
```
samjones1001: All of the tests in this file begin in the same way - feels like a good use case for extracting out into a helper method.

PR #1018, opened 2018-10-28
Chitter Challenge

runticle:


app.rb
```diff
@@ -0,0 +1,79 @@
...
+    end
+  end
+
+  get '/profile' do
+    @cheets = Cheet.all.reverse
+    @current_user = session[:current_user]
+    erb :profile
+  end
+
+  get '/logout' do
```
soph-g: Would use 'delete' as the HTTP verb as you are deleting the session.

lib/user.rb
```diff
@@ -0,0 +1,55 @@
+class User
+
+  attr_reader :id, :email, :name, :user, :pass
+
+  def initialize(id, email, name, user, pass, confirm = pass)
```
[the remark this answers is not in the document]
soph-g: Why has confirm been set to `pass` by default?

spec/features/sign_in_spec.rb
```diff
@@ -0,0 +1,38 @@
+
+feature 'As a user:' do
+  scenario 'I can create an account' do
+    create_account('jones')
+    expect(page).to have_content "Welcome, John"
```
soph-g: Names don't match in this test, would expect it to fail

spec/user_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'user'
+
+describe User do
```
soph-g: Make sure you are also testing the class methods interact with the database as expected.

PR #1025, opened 2018-10-28
Chitter challenge

Br0ckers:
All straight up user stories completed.

All tests passing except 2 unit tests, feature tests for same items pass.


lib/message.rb
```diff
@@ -0,0 +1,36 @@
+require 'pg'
+require 'Date'
+require './lib/user'
+
+class Message
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
samjones1001: This block of code for connecting to the correct database appears in lots of the methods in your models - could be a good candidate for extraction to it's own method in order to avoid this repetition

PR #1031, opened 2018-10-29
Submission

sacullezzar:
Trying to introduce account features.  Seems to be unable to find users table in database.  Basic styling because CSS.


app.rb
```diff
@@ -0,0 +1,30 @@
+require 'sinatra/base'
+require './lib/message.rb'
+require './lib/account.rb'
+require 'date'
+require './lib/database_connection_setup.rb'
+
+class Chitter < Sinatra::Base
```
samjones1001: Think about taking a look at RESTful routing. It's a convention for the way we name routes in a web application and a pretty east win for making your code look of a professional standard.

samjones1001: Other than that, this controller looks good.  Your routes are pretty concise and aren't responsible for any heavy logical lifting.

spec/features/features_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'pg'
+
+feature "Posting a message" do
+  scenario "Visiting the index page" do
+    visit('/')
+    expect(page).to have_content "Chitter"
+  end
+  scenario "User posts a message and sees it added to the list" do
+
+    connection = PG.connect(dbname: 'chitter_challenge_test')
```
samjones1001: I'd probably move this db connection stuff out of my tests into a helper method

PR #1032, opened 2018-10-29
Chitter Challenge pull request

NaomiBid:


spec/features/adding_peep_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature 'Adding a new peep' do
+  scenario 'A user can add a peep to Chitter Manager' do
+    visit('/peeps/new')
+    fill_in('peep text', with: 'Test Peep')
+    fill_in('name', with: 'Test Name')
+    click_button('Submit')
+
+    expect(page).to have_content('Test Name', 'Test Peep')
+  end
+end
```
hnryjmes: good feature test

spec/setup_test_database.rb
```diff
@@ -0,0 +1,10 @@
+require 'pg'
+
+def setup_test_database
+  p "Setting up test database..."
+
+  connection = PG.connect(dbname: 'peep_manager_test')
+
+  # Clear the bookmarks table
+  connection.exec("TRUNCATE peeps;")
+end
```
hnryjmes: good idea to move code into a helper file, even if it wasn't used in this most recent commit

views/peeps/peeps.erb
```diff
@@ -0,0 +1,5 @@
+<ul>
+  <% @peeps.reverse.each do |peep| %>
```
hnryjmes: good choice to use reverse, however it might be good to check that @peeps is definitely keeping the peeps in chronological order to begin with

PR #1034, opened 2018-10-29
chitter challenge - David Best

davidbest1882:


lib/user.rb
```diff
@@ -0,0 +1,41 @@
...
+    # encrypt the plantext password
+    encrypted_password = BCrypt::Password.create(password)
+
+    # insert the encrypted password into the database, instead of the plaintext one
+    result = connection.exec("INSERT INTO users (name, email, username, password) VALUES('#{name}', '#{email}', '#{username}', '#{encrypted_password}') RETURNING id, name, email, username;")
+    User.new(id: result[0]['id'], name: result[0]['name'], email: result[0]['email'], username: result[0]['username'])
+  end
+
+  def self.find(id)
+    return nil unless id
```
samjones1001: Not sure that this line is doing anything.  there's always going to be an id - if there isn't then the method will throw an argument error.

spec/message_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      #Add the test data
+      message = Message.create(message: "This is my first peep", name: "Joe Bloggs")
+      Message.create(message: "This is my second peep", name: "Joe Bloggs")
+      Message.create(message: "This is my third peep", name: "Joe Bloggs")
+      # connection.exec("INSERT INTO chitter (message) VALUES('This is my first peep');")
+      # connection.exec("INSERT INTO chitter (message) VALUES('This is my second peep');")
+
+      messages = Message.all
+
+      expect(messages.length).to eq 3
```
samjones1001: as a rule I'd suggest only having one expectation for each of your unit tests.  The reason for this being that if you later make some change that causes this test to break, you won't be able to tell straight away which expectation is causing the failure.

PR #1035, opened 2018-10-29
Friday challenge

hurc87:
The user can sign in, but is not able to sign out just yet, the formatting is all over the place at the moment as I got stuck on one method. The rest of the program would not run all the time this was broken, we managed to fix it this morning but now do not have time to fix the rest of the program.


lib/user.rb
```diff
@@ -0,0 +1,34 @@
...
+  def self.create(email:, name:, username:, password:)
+    encrypted_password = BCrypt::Password.create(password)
+
+    result = DatabaseConnection.query("INSERT INTO users (email, name, username, password) VALUES ('#{email}', '#{name}', '#{username}', '#{encrypted_password}') RETURNING id, email, name, username;")
+    User.new(id: result[0]['id'], email: result[0]['email'], name: result[0]['name'], username: result[0]['username'])
+  end
+
+  def self.find(id)
+    return nil unless id
+    result = DatabaseConnection.query("SELECT * FROM users WHERE id = #{id};")
```
samjones1001: Judging from the error messages on the couple of tests which you have failing, this is the line that's stopping your application from working.

There's a bit of an error in the syntax for your SQL query - can you spot it? 🤔

PR #1040, opened 2018-10-29
Letting go....awaiting your review comments!

PreetiSekhon2:


app.rb
```diff
@@ -0,0 +1,101 @@
+require "sinatra/base"
+require "sinatra/flash"
+require "./lib/user.rb"
+require "./lib/message.rb"
+require_relative "./database_connection_setup.rb"
+
+class Chitter < Sinatra::Base
```
samjones1001: Your controller this week looks much improved over your RPS implementation. It's great to see you using RESTful routing conventions and keeping your routes concise and free of logical heavy lifting.

app.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  get "/user/logout" do
+    session.clear
+    flash[:usernotice] = "Logout successful"
+    redirect "/message"
+  end
+
+  post "/user" do
+    user = User.login(name: params[:name],password: params[:password])
+    if user.is_a?(User)
```
samjones1001: The idea you're applying here of determining which route an application should follow based on the result of a db query is a very common one.

More conventionally, you would simply have your login method return false or nil if it fails allowing you to do something like

```ruby
user = User.login(name: params[:name],password: params[:password])
if user 
  //routing logic
else 
  //other routing logic
end
```

lib/communicationmanager.rb
```diff
@@ -0,0 +1,35 @@
+require "./lib/user.rb"
+require "./lib/databaseconnection.rb"
+
+
+class Communicationmanager
```
samjones1001: This feels like a sensible extraction

lib/user.rb
```diff
@@ -0,0 +1,56 @@
+require "./lib/databaseconnection.rb"
+require "bcrypt"
+
+class User
+
+attr_reader :name,:password, :email, :id
+
+  def self.create(name:, password:, email:, receive_email:)
```
samjones1001: A few of the methods in this class feel a bit busy and there's some repeated code across them - potentially a candidate for some small refactors?

migrations/db_scripts
```diff
@@ -0,0 +1,27 @@
+create database chitter
```
samjones1001: Extracting these scripts out feels like a good step in the direction of automating db setup.  If you want to continue down this path take a look at what you can do with a Rakefile

PR #1062, opened 2018-11-26
My Chitter Challenge submission

willmcb:
Other than what is in the README todo section I think the tests could be drier, but I didn't get round to looking at this. 

Furthermore, there is some error handling that I didn't get around to, for instance if a second user tries to add a username that is already taken, also the peeps need to be escaped currently the app errors if you use single quotes in the peep content.

Would love to get some feedback on this.

Thanks 👍


lib/models/peep.rb
```diff
@@ -0,0 +1,45 @@
...
+                             user_id) VALUES ('#{content}', '#{user_id}');")
+  end
+
+  def self.find(id:)
+    result = DatabaseConnection.query("SELECT * FROM peeps WHERE id = #{id};")
+    result.map { |record| peep_builder(record) }.first
+  end
+
+  def self.update(id:, content:)
+    DatabaseConnection.query("UPDATE peeps SET content
```
alicelieutier: This is not the best place for a new line. You may want to write this instead:
```
DatabaseConnection.query(
    "UPDATE peeps SET content = '#{content}'
     WHERE id = #{id};"
)
```

lib/models/user.rb
```diff
@@ -0,0 +1,50 @@
...
+    return nil unless id
+    result = DatabaseConnection.query("SELECT * FROM users
+                                       WHERE id = '#{id}'").first
+    build_user(result)
+  end
+
+  def self.authenticate(email:, password:)
+    result = DatabaseConnection.query("SELECT * FROM users
+                                      WHERE email = '#{email}'").first
+    if result && BCrypt::Password.new(result['password']) == password
```
alicelieutier: This is encryption of password but not hashing.

lib/chitter.rb
```diff
@@ -0,0 +1,79 @@
+require './database_connection_setup'
```
alicelieutier: Having code execute on 'require' seems like an antipattern to me. Instead, this code should be in a function that you can call at the beginning of your app (just after line 11)

PR #1063, opened 2018-11-26
Chitter challenge

AgaDufrat:


README.md
```diff
@@ -62,46 +44,43 @@ Notes on functionality:
...
+To set up the appropriate tables, connect to each database in `psql` and run the SQL scripts in the `db/migrations` folder in the given order.
 
-Reviewers will potentially be using this [code review rubric](docs/review.md).  Referring to this rubric in advance may make the challenge somewhat easier.  You should be the judge of how much challenge you want this weekend.
+### To run the Chitter app:
 
-Notes on test coverage
-----------------------
+```
+rackup -p 9292
+```
```
willmcb: I think the default port is 9292 so you can just run rackup - unless you want to run it on a non default port

README.md
```diff
@@ -62,46 +44,43 @@ Notes on functionality:
...
 ```
+rspec
+```
+
+### To run linting:
 
+```
+rubocop
+```
```
willmcb: Nice clear README.

app.rb
```diff
@@ -0,0 +1,32 @@
...
+require_relative './lib/message_manager'
+require_relative './lib/user'
+require './database_connection_setup'
+
+class ChitterApp < Sinatra::Base
+  enable :sessions
+
+  get '/chitter' do
+    @user = User.find(id: session[:user_id])
+    @chitter = Message.show_peeps
```
willmcb: Why not have your program objects have the same name as you DB tables? 

i.e. Peeps table -> Peep object
      Users table -> Users object

You can then have clearer method names like:

`Peeps.all`

app.rb
```diff
@@ -0,0 +1,32 @@
...
+  post '/message' do
+    Message.post_peep(content: params[:content])
+    redirect '/chitter'
+  end
+
+  get '/users/new' do
+    erb :"users/new"
+  end
+
+  post '/users/new' do
```
willmcb: Nice use of RESTful routes!

database_connection_setup.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/database_connection'
+
+if ENV['ENVIRONMENT'] == 'test'
+  DatabaseConnection.setup('chitter_test')
+else
+  DatabaseConnection.setup('chitter')
```
willmcb: Good names for your databases, I had `chitter_prod` and `chitter_test` - thinking back I think the `_prod` part was redundant.

lib/message_manager.rb
```diff
@@ -0,0 +1,29 @@
...
+      ORDER BY timestamp DESC")
+    result.map do |peep|
+      Message.new(id: peep['id'], content: peep['content'], \
+        timestamp: peep['timestamp'])
+    end
+  end
+
+  attr_reader :id, :content, :timestamp
+
+  def initialize(id:, content:, timestamp:)
```
willmcb: You can have the database automatically apply timestamps rather than passing them in with you program objects - when creating a table you can use:

```
CREATE TABLE peeps (
  id SERIAL PRIMARY KEY,
  date_added timestamp DEFAULT CURRENT_TIMESTAMP,
  content varchar(240)
);
```
CURRENT_TIMESTAMP is a PostgreSQL constant that holds the current time.

lib/user.rb
```diff
@@ -0,0 +1,32 @@
...
+require_relative './database_connection'
+
+class User
+  def self.create(name:, username:, email:, password:)
+    encrypted_password = BCrypt::Password.create(password)
+
+    result = DatabaseConnection.query("INSERT INTO users \
+      (name, username, email, password) \
+      VALUES('#{name}', '#{username}', '#{email}', '#{encrypted_password}') \
+      RETURNING id, name, username, email;")
```
willmcb: Well done to split these over multiple lines - I think it would be easier to read if they were justified though.

lib/user.rb
```diff
@@ -0,0 +1,32 @@
...
+    User.new(id: result[0]['id'], name: result[0]['name'],\
+      username: result[0]['username'], email: result[0]['email'])
```
willmcb: You can remove the multiple calls to `[0]` if you call `.first` on the result set returned from the database.

spec/database_connection_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'database_connection'
+
+describe DatabaseConnection do
+  describe '.setup' do
+    it 'sets up a connection' do
+      expect(PG).to receive(:connect).with(dbname: 'chitter_test')
+
```
willmcb: Nice use of mocking.

spec/message_manager_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      expect(peep).to be_a Message
+      expect(peep.id).to eq persisted_data['id']
+      expect(peep.content).to eq 'My new peep'
+    end
+  end
+
+  describe '.show_peeps' do
+    it 'returns a list of peeps' do
+      time_now = Time.parse("2018-11-22 12:00:00")
+      allow(Time).to receive(:now) { time_now }
```
willmcb: This is a robust way of checking the time. Buy hard coding it you are making sure that the test runs the same each time. 

You might want to think about testing multiple different times though so that you catch edge cases. 

For example differences like `09:01` and `9:01` where you have leading zeros etc.

views/index.erb
```diff
@@ -0,0 +1,24 @@
+
+<% if @user %>
```
willmcb: Well done for making your pages dynamic like this.

views/index.erb
```diff
@@ -0,0 +1,24 @@
...
+<% end %>
+
+<p>Previous messages:</p>
+<ul>
+  <% @chitter.each do |peep| %>
+    <li class="peep" id="peep-<%= peep.id %>">
+        <%= peep.content %> (<%= peep.timestamp %>)
+    </li>
+  <% end %>
+</ul>
```
willmcb: You might want to think about using separate views for different parts of your functionality as well as layout files. This way you reduce the amount of logic you need in your views while still providing good user interaction.

More info here: 
http://ruby-log.blogspot.com/2013/02/sinatra-layouts.html

bd/migrations/01_create_messeges_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE messages (id SERIAL PRIMARY KEY, content TEXT, timestamp TIME );
```
willmcb: You can have the database automatically apply timestamps rather than passing them in with you program objects - when creating a table you can use:

```
CREATE TABLE peeps (
  id SERIAL PRIMARY KEY,
  date_added timestamp DEFAULT CURRENT_TIMESTAMP,
  content varchar(240)
);
```
CURRENT_TIMESTAMP is a PostgreSQL constant that holds the current time.

PR #1066, opened 2018-11-26
Chitter challenge pull

joecoker:


lib/peep.rb
```diff
@@ -0,0 +1,34 @@
...
+  attr_reader :peep_id, :content, :timestamp
+
+  def initialize(peep_id, content, timestamp)
+    @peep_id = peep_id
+    @content = content
+    @timestamp = timestamp
+  end
+
+  def self.list
+    if ENV['ENVIRONMENT'] == 'test'
```
alicelieutier: These lines of code are a good candidate for DRYing.

PR #1073, opened 2018-11-26
Chitter challenge 

SamJamCul:


lib/cheet.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM cheets;")
+    unsorted = result.map do |cheet|
+      Cheet.new(id: cheet['id'],
+        content: cheet['content'],
+        timestamp: cheet['timestamp'])
+    end
+    sorted = unsorted.sort_by { |cheet| cheet.timestamp }
```
alicelieutier: You could sort in your SQL query.

PR #1076, opened 2018-11-26
All basic features, bar uniqueness of email and username.  Tagging feature is unfinished

MichaelSlevin:


lib/peeps.rb
```diff
@@ -0,0 +1,96 @@
...
+                    new_peep["replying_to_id"]
+                    )
+          new_peep.update_tags
+          new_peep
+  end
+
+  def update_tags
+    tags = self.message.scan(/@\w+/)
+    tags.each do |tag|
+      p tag
```
alicelieutier: remove p statements. You can leave the ones that will help in debugging the server later on, like "#{username} just signed in", but not ones used for debugging while coding.

lib/peeps.rb
```diff
@@ -0,0 +1,96 @@
...
+                    new_peep["time"],
+                    new_peep["replying_to_id"]
+                    )
+          new_peep.update_tags
+          new_peep
+  end
+
+  def update_tags
+    tags = self.message.scan(/@\w+/)
+    tags.each do |tag|
```
alicelieutier: What happens if the same user is tagged twice. Can you add a test for it?

lib/peeps.rb
```diff
@@ -0,0 +1,96 @@
...
+      p tag
+      tag_stripped = tag[1..tag.size]
```
alicelieutier: `username` seems like a better variable name here.
```ruby
username = tag[/\w+/]
```

lib/peeps.rb
```diff
@@ -0,0 +1,96 @@
...
+
+
+  def self.is_a_user?(username)
+    Users.user_with_name(username)
+  end
+
+
+  private
+
+  def self.sql(query)
```
alicelieutier: DRY out in other component.

lib/peeps.rb
```diff
@@ -0,0 +1,96 @@
...
+  end
+
+  def self.all
+    peeps = sql('
+                      SELECT p.id, user_id, message, time, name, username
+                      FROM peeps p
+                      left join users u
+                      on u.id = user_id
+                      order by time desc;')
+    peeps.map { |peep| Peeps.new(peep['id'],
```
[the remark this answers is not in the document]
MichaelSlevin: thanks for providing the example

PR #1096, opened 2019-01-14
An chitter challenge

anhuiyang:


Gemfile
```diff
@@ -2,11 +2,22 @@ source 'https://rubygems.org'
 
 ruby '2.5.0'
 
+gem 'sinatra'
+gem "data_mapper", '1.2.0'
+gem "dm-postgres-adapter", "~> 1.2"
 gem 'rake'
+
```
CelineKaslin: I think Rubocop gem have to be in the group test

app.rb
```diff
@@ -0,0 +1,75 @@
...
+    erb :login
+  end
+
+  post '/login' do
+    user = User.auth(params[:username], params[:password])
+    session[:userid] = user.id if user
+    redirect '/'
+  end
+
+  get '/logout' do
```
CelineKaslin: Maybe it can be an other approach to do a delete '/sessions' route for the log out, which gonna "kill" the session of the user.
also, you can redirect your user to the homepage instead of having a view for log out.

app.rb
```diff
@@ -0,0 +1,75 @@
...
+    current_user
+    if logged_in?
+      erb :logged_in
+    else
+      erb :index
+    end
+  end
+
+  post '/post_msg' do
+    Message.create(
```
CelineKaslin: If I understand well anyone is able to post a message a being logged in is not mandatory ?

config/data_mapper.rb
```diff
@@ -0,0 +1,5 @@
+require './lib/message'
+require './lib/user'
+DataMapper.setup(:default, "postgres://localhost/chitter_#{ENV['RACK_ENV']}")
+DataMapper.finalize
+DataMapper.auto_migrate!
```
CelineKaslin: Might be good to put this task in the rake file

spec/features/user_post_msg_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'pry'
+feature 'post a message' do
+  scenario 'to see the message with timestamp' do
+    visit '/'
+    fill_in :msg, with: '1st test message'
+    click_on 'send'
+    time = Time.now.strftime("%v %R")
```
CelineKaslin: Don't get why you had the time stamp in here. Might be my misunderstanding, I ll search about it

PR #1107, opened 2019-01-14
Final chitter without tags

michaelnguyen974:


app.rb
```diff
@@ -0,0 +1,70 @@
...
+    @peeps = Peep.all
+    erb :index
+  end 
+
+  post '/peeps' do 
+    Peep.create(content: params[:content])
+    redirect '/'
+  end
+
+  get '/peeps/:id' do
```
Sindex42: Does this route work?

app.rb
```diff
@@ -0,0 +1,70 @@
...
+    @peep = Peep.get(params[:id])
+    erb :peeps
+  end
+
+  get '/signup' do 
+    erb :signup
+  end
+
+  post '/signup' do 
+    user = User.create(name: params[:name], username: params[:username], email: params[:email], password: params[:password])
```
Sindex42: Could this have been split into multiple lines?

spec/feature/user_spec.rb
```diff
@@ -0,0 +1,86 @@
+require './lib/user.rb'
+require 'web_helper.rb'
+
```
Sindex42: Would it be better if the feature tests were less granular?

PR #1109, opened 2019-01-14
updated

GLMill:


GLMill:
Ive updated this! actually understand datamapper now - kinda, had written singleton not module corrected this lol, everythings cleaner - i took out the js form hiding (i included in the first place just to see it working probs should have just made it say hello lol )

PR #1127, opened 2019-02-03
Completed STRAIGHT UP & ADVANCED challenges

nkhil:
Completed `STRAIGHT UP` and `ADVANCED` user stories. 

**Goal: I can refactor anything**

Please take a look at the following commits for this project:

- refactor to setup DatabaseConnection object (e6d451e)
- refactor user.rb to use DatabaseConnection (e6d451e)

I've tried to DRY out my code as much as possible by creating a separate `DatabaseConnection` class to abstract out this responsibility. 

I have also refactored the code to make sure it doesn't exceed the 80 character limit by adding line breaks in the PSQL queries making it easier to read and comprehend. 

In addition, also note the `formatted_date` and `formatted_time` instance methods in `Peep` that were refactored to display a more semantic time.


app/lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
+require 'pg'
+
+class DatabaseConnection 
+  def self.setup(db)
+    @connection = PG.connect(dbname: db)
```
JPProgrammingTime: Could you not have made this method do the environment variable 'test' checks and pass on the appropriate connection to be used in your other classes like Peeps and User? Might have saved you the trouble of having to re-write the syntax for that in methods such as the 'self.all', 'self.create' etc methods. This way you can just connect to the appropriate database as a single line in such methods like databaseconnect = DatabaseConnection.setup then do whatever you need like databaseconnect.exec(----).

db/migrations/create_chitter_table.sql
```diff
@@ -1,3 +1,4 @@
 CREATE DATABASE chitter;
 CREATE TABLE peeps(id SERIAL PRIMARY KEY, peeps VARCHAR(60));
--- ALTER TABLE peeps ADD COLUMN title VARCHAR(60);
\ No newline at end of file
+ALTER TABLE peeps ADD COLUMN timestamp timestamp with time zone;
```
JPProgrammingTime: Nice, I did not think about the time zone. Mine is just the time of post, did not know you could do this.

db/migrations/create_chitter_table.sql
```diff
@@ -1,3 +1,4 @@
...
+CREATE TABLE users (id SERIAL PRIMARY KEY, email VARCHAR(60), password VARCHAR(140));
```
JPProgrammingTime: Might be missing something here but how did you manage to link users to their peeps without a foreign key?

spec/setup_test_database.rb
```diff
@@ -0,0 +1,7 @@
+require 'pg'
+
+def setup_test_database
+  p "Setting up Chitter's test database..."
+  connection = PG.connect(dbname: 'chitter_test')
+  connection.exec("TRUNCATE peeps;")
```
JPProgrammingTime: So I decided to also test my user's sign-up process using a test database so I had to truncate my users table as well. I am assuming you added default test users to your test database and just truncated your peeps table?

app/lib/peep.rb
```diff
@@ -12,26 +13,14 @@ def initialize(peep:, id:, timestamp:)
...
-    if ENV['ENVIRONMENT'] == 'test'
-      connection = PG.connect(dbname: 'chitter_test')
-    else
-      connection = PG.connect(dbname: 'chitter')
-    end
-
-    result = connection.exec("SELECT * FROM peeps ORDER BY id DESC;")
+    result = DatabaseConnection.query("SELECT * FROM peeps ORDER BY id DESC;")
     result.map { |peep| 
       Peep.new(peep: peep['peep'], id: peep['id'], timestamp: peep['timestamp'])
```
JPProgrammingTime: Really like how you used your wrapping to return peeps here. Would make testing, refactoring and adding features in the future much easier I think.

app/views/index.erb
```diff
@@ -10,7 +10,13 @@
 <body>
   <div class="main__container">
     <h1>Chitter</h1>
-
+    
+      <% if @user %>
```
JPProgrammingTime: Would it be a good idea to allow non-users to see peeps or better to only allow users to see peeps?

PR #1133, opened 2019-02-04
Chitter Challenge

Estevenson1994:
Completed attempt of the Chitter Challenge excluding the advanced user story. 

All tests pass locally, travis is failing to find database. 

Goals I would like feedback on:

1. I can learn anything by myself:

Throughout this project, and the previous week it has been required to learn how to create and interact with databases to a basic level. This is something I had limited exposure to before. Evidence for this can be found in the model files found in the 'lib' directory. 

This project also required using BCrypyt, a password manager class so that the passwords stored in the database are maintained as encrypted. Evidence for the use of this can be found in the lib/user.rb file.


lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
+require 'pg'
+
+class DatabaseConnection
+  def self.setup(dbname)
+    @connection = PG.connect(dbname: dbname)
+  end
+
+  def self.connection
```
thielsen: Attr_reader might be better here

lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
+require 'pg'
```
thielsen: Nice clean class..

lib/peeps.rb
```diff
@@ -0,0 +1,31 @@
+require_relative 'database_connection'
+
+class Peep
+  
+  attr_reader :id, :peep, :date, :time, :username
+  def initialize(id:, peep:, date:, time:, username:)
```
thielsen: When you get to methods with a larger number of arguments like this it can be better to take in a single hash argument and read from that. Means argument order doesn't matter and allows easy add and subtract of arguments as things change without directly breaking existing connectivity.

lib/user.rb
```diff
@@ -0,0 +1,37 @@
+require_relative 'database_connection'
+require 'bcrypt'
+
+class User
+  include BCrypt
+  attr_reader :id, :name, :username, :email, :password
+
+  def initialize(id:, name:, username:, email:, password:)
```
thielsen: Same comm ent regarding maybe taking in a hash

lib/user.rb
```diff
@@ -0,0 +1,37 @@
...
+    @id = id
+    @name = name
+    @username = username
+    @email = email
+    @password = password
+  end
+ 
+  def self.create(name:, username:, email:, password:)
+    hidden_password = BCrypt::Password.create(password)
```
thielsen: Could wrap the password class in your own class - that way if you decide to use a different password hash function in the future no need to change everything in the code

spec/features/peeps_spec.rb
```diff
@@ -0,0 +1,53 @@
+require_relative '../web_helper.rb'
+require 'timecop'
+
+feature 'Viewing peeps' do 
+  include Helpers
```
thielsen: Could just add this into spec_helper as a require without it being a class - would DRY things up a bit

spec/features/registration_spec.rb
```diff
@@ -0,0 +1,12 @@
+feature 'registration' do
```
thielsen: All the tests look thorough and look to test behaviour rather than state

lib/user.rb
```diff
@@ -0,0 +1,37 @@
...
+    return nil unless id
+    result = DatabaseConnection.query("SELECT * FROM users WHERE id = '#{id}';")
+    User.new(id: result.first['id'], name: result.first['name'], username: result.first['username'], email: result.first['email'], password: result.first['password'])
+  end
+
+  def self.authenticate(username:, password:)
+    result = DatabaseConnection.query("SELECT * FROM users WHERE username = '#{username}';")
+    if result.first.nil?
+      nil
+    elsif BCrypt::Password.new(result.first['password']) == password
```
thielsen: Could tidy this up with a 'return nil if result.first.nil?, and then lose the elsif/end

app.rb
```diff
@@ -0,0 +1,71 @@
+require 'sinatra/base'
```
thielsen: Good use of encapsulation in most of these for Users/Peeps

spec/features/peeps_spec.rb
thielsen: Though now I read the review notes they use this module Helpers - not sure it needs to be called every time though (or maybe I'm wrong hence why you've done it..)

PR #1135, opened 2019-02-04
Partial completion

Shendidy:


JamesRudrum:
Sherif Code Review

Step 1
no logout
no replying
some css styling

Step 2
No setup instructions
No screenshots
gem file uses test groups
Spec helper is configured correctly(cleans database and connects feature testing with app)

Feature Tests

need second expect in posting_peeps_spec 
i.e. expect(page).to have_content(‘This is a second peep’)

need another test to test for reverse order
i.e. once have created pee.ps and visit page where they are posted should test reverse order of list using>>
expect(page.find('li:nth-child(1)')).to have_content 'Test chit 03'
expect(page.find('li:nth-child(2)')).to have_content 'Test chit 02'
expect(page.find('li:nth-child(3)')).to have_content 'Test chit 01'

need test for checking if time posted with peep
i.e.
Post.create(chit: "Test chit 01")
visit('/posts')
expect(page).to have_content 'Peep:'
expect(page).to have_content 'Test chit 01'
expect(page).to have_content 'Posted On:'
expect(page).to have_content(Time.now.hour)
expect(page).to have_content(Time.now.min)

Feature test for sign up
visit('/')
fill_in('username', with: 'JPRudrum')
fill_in('password', with: 'AFC4LYF')
fill_in('email', with: 'jr@gmail.com')
click_button('Sign Up!')
expect(page).to have_current_path "/posts"

Feature test for log in

Need unit tests

Maybe use more routes. But yours works great so maybe not needed, however routes need to be RESTful.

Use <article> instead of <div> in html files as is recommended for HTML5

Maybe use helpers for other things in app such as user.

COMMIT MORE helps prove TDD and other stuff

PR #1139, opened 2019-02-04
James Rudrum 4 User Stories

JamesRudrum:
How do I do CSS
I can TDD anything - from my commits can tell that test than pass test


Shendidy:
- Beautiful user interface :)
- 1st page is the posts/signup, but there’s no option to login if you already have an account.
- I think to read peeps we don’t need to be signed up.
- Every time a user class is called it checks which database to use, is it possible to use database connection method when loading the page, then just submitting a query whenever needed rather than checking for the database (test or not) every time?
- Instead of using .reverse, would it be better to load the information from the database in the order we will be using it for display using ORDER BY ‘field’ ASC or DESC?
- Once we’re in there’s no button to take us to a page where we can place a post, we need to type in the URL ourselves, can we add a button or a link to do that?

PR #1140, opened 2019-02-04
Weekend 4: Chitter Challenge

aidengleave:
#I write code that is easy to change

I feel that this challenge shows that I have come on a great deal in my ability to write clear and readable code that would be easy to work with.

I have made sure to extract repeated functions to their own methods so that they are easily changed.

I still feel that I need to focus more on using agile development processes and looking after my emotional well-being.


db/migrations/create_peeps_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps(user_id SERIAL PRIMARY KEY, user_name VARCHAR(60), peep VARCHAR(140), time TIMESTAMP default current_timestamp);
```
dippedincode: Great use of inherent SQL time function

PR #1147, opened 2019-03-02
Chitter - tell the world how good your toast is

acodeguy:


spec/db_helpers.rb
```diff
@@ -0,0 +1,6 @@
+def setup_test_database
+  connection = DatabaseConnection.setup('chitter_test')
+  connection.query('truncate users, peeps')
+  connection.query('alter sequence users_id_seq restart with 1')
```
limavadakara: This is good; hadn't realised that you could reset the sequence to 1.

PR #1153, opened 2019-03-03
Chitter by Jules

Jimjule:


lib/chitter.rb
```diff
@@ -0,0 +1,35 @@
...
+  def self.post(text, datetime = DateTime.now)
+    if ENV['ENVIRONMENT'] == 'test'
+      peeps = PG.connect( dbname: 'chitter_peeps_test')
+    else
+      peeps = PG.connect( dbname: 'chitter_peeps')
+    end
+    peeps.exec("INSERT INTO peeps (text, datetime) VALUES ('#{text}', '#{datetime}');")
+  end
+
+  def initialize(id, text, datetime = DateTime.now)
```
Brendao1: You can use Time.now() if Rubocop doesn't like DateTime

app.rb
```diff
@@ -0,0 +1,19 @@
+require "sinatra"
```
Brendao1: It's really nice that you have your interface combined in 1 view. This makes the app user friendly.

lib/chitter.rb
```diff
@@ -0,0 +1,35 @@
+require "pg"
+
+class Chitter
+
+  attr_reader :peeps, :id, :text, :datetime
+
+  def self.list
+    @peeps = Array.new
+    if ENV['ENVIRONMENT'] == 'test'
```
Brendao1: Perhaps you could refactor this code with a ternary?

views/index.erb
```diff
@@ -0,0 +1,37 @@
...
+  <meta charset="utf-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+
+  <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+  <link rel="stylesheet" type="text/css" href="/style.css">
+
+</head>
+
+<body>
+
```
Brendao1: The blank lines between elements in the body makes reading the code easier. I'm not sure what the Rubocop convention is, but some of the lines at the top and at the end could be removed.

PR #1158, opened 2019-03-04
Was ill... did as much as could

dafuloth:
There's still lots to explore. Looking forward to revisiting this project


README.md
```diff
@@ -1,133 +1,126 @@
...
-Challenge:
--------
+## Getting started
 
-As usual please start by forking this repo.
+_PostgreSQL should be installed_
 
-We are going to write a small Twitter clone that will allow the users to post messages to a public stream.
+```sh
+git clone git@github.com:dafuloth/chitter-challenge.git
```
dory1318: The link does not work, managed to clone using this one: git clone https://github.com/dafuloth/chitter-challenge.git

README.md
```diff
@@ -1,133 +1,126 @@
...
------
+## Challenges
+- I want the `timestamp` to be set automatically, based on when the peep is made. Determined that an easy way to achieve this is with default value `DEFAULT now()`.
+
+## Features
+
+My approach is to tackle the "_seeing peeps_" features first since they do not require signing up or posting (which requires signing up).
 
-This week you integrated a database into Bookmark Manager using the `PG` gem and `SQL` queries. You can continue to use this approach when building Chitter Challenge.
+### Design choices: The Single-Page Application
```
dory1318: As  a personal opinion, I prefer single page app too, for the reasons mentioned by you!

lib/peep.rb
```diff
@@ -0,0 +1,27 @@
...
+      DatabaseConnection.setup('chitter')
+    end
+
+    raw_peeps = DatabaseConnection.query('SELECT * FROM peeps')
+
+    peeps = raw_peeps.map do |item|
+      { timestamp: item['timestamp'][0, 19], message: item['peep'] }
+    end
+
+    peeps.reverse
```
dory1318: good solution! but a different approach; sql lets you make ascending/descending order, so you can select it like this: 'SELECT * FROM peeps ORDER BY timestamp DESC;' - so no need the extra line to reverse it :)

public/css/style.css
```diff
@@ -0,0 +1,52 @@
+body {
```
dory1318: nice look!

lib/database_connection.rb
```diff
@@ -0,0 +1,12 @@
+require 'pg'
+
+class DatabaseConnection
+  def self.setup(dbname)
+    @connection = PG.connect(dbname: dbname)
+  end
+
+  def self.query(sql)
+    @connection.exec(sql)
```
dory1318: I get an error message when I try to post something containing this character: '
" PG::SyntaxError at /
ERROR: syntax error at or near "t" LINE 1: ...ERT INTO peeps VALUES (DEFAULT, DEFAULT, 'why can't post pee... ^
file: database_connection.rb location: async_exec line: 9 "

I coldn't find the problem yet

app.rb
```diff
@@ -0,0 +1,29 @@
...
+    message = params['message']
+
+    @zero_length_msg = message.length.zero?
+
+    unless @zero_length_msg
+      DatabaseConnection.query("INSERT INTO peeps VALUES (DEFAULT, DEFAULT, '#{message}')")
+      redirect '/'
+    end
+
+    erb :index
```
dory1318: not necessary to put the erb :index line in 'post' methods

app.rb
```diff
@@ -0,0 +1,29 @@
...
+    @peeps = Peep.all
+
+    erb :index
+  end
+
+  post '/' do
+    @peeps = Peep.all
+    message = params['message']
+
+    @zero_length_msg = message.length.zero?
```
dory1318: this maybe could have been in a ./lib/?.rb file as a separate method to keep the app.rb clean from logic. It's ok with shorter codes but it would be easier to understand longer codes in the future.

README.md
dafuloth: Glad at least one way worked for you :) I think for some both will work, or just one of them

lib/peep.rb
dafuloth: Good to know! I did not much look into SQL so that Ruby way was the only way I knew how to do it

public/css/style.css
dafuloth: Haha, thank you. Not a professional dev yet but I will at least try to make it presentable for you to look at ^^

lib/database_connection.rb
dafuloth: Good catch. I did not take any measures to sanitise user input, which I should have done. Adding it to the list of possible improvements!

dafuloth: What you have demonstrated is an example of SQL injection, similar to this: 
https://www.netsparker.com/blog/web-security/sql-injection-cheat-sheet/#ByPassingLoginScreens

app.rb
dafuloth: Agreed! I really should make time for refactoring in the weekend challenges...

app.rb
dafuloth: True, and yet it was the best solution I could come up with at the time... I need it to load the `index.erb` template so that it can respond to the true/false state of `@zero_length_msg`. This was in order to achieve the effect of `Sinatra/Flash` which I was unable to get working (with `enable :sessions`)

dafuloth: Assuming the "happy path", program flow would go to `redirect '\'` as we would expect the `post` to do. I think maybe if I could get Sinatra/Flash working then I wouldn't have use `erb :index` in this `post` method.

PR #1164, opened 2019-03-04
Chitter Challenge

steve-f-kahn:
I did up to and including user login and registration.


lib/peep.rb
```diff
@@ -0,0 +1,12 @@
+require_relative 'database_connection'
+
+class Peep
+
+  def self.all
+    Database_Connection.querry("SELECT * FROM peeps ORDER BY time DESC")
```
Harrison25000: Nice, very short and precise, use of a separate database connector helper file was v. smart, simplifies everything greatly.

lib/database_connection.rb
```diff
@@ -0,0 +1,16 @@
+require 'pg'
+
+class Database_Connection
```
Harrison25000: Very wise to use a separate file for the connecting features, makes the rest of your code (peeps & users) much more simplified and clear.

PR #1170, opened 2019-03-31
James Palmer Chitter app

jpalmerr:
Completed first two user stories.
Was short for time this weekend.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -5,7 +5,6 @@ require_relative 'lib/peep'
 class Chitter < Sinatra::Base
   get '/' do
     @peepdeck = Peep.all
-    puts @peepdeck
     erb :index
   end
 
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,14 @@
+require 'sinatra/base'
+
+# controller
+class Chitter < Sinatra::Base
+  get '/' do
+    erb :index
+  end
+
+  get '/message/new' do
+    erb :compose
+  end
+
+  run! if app_file == $0
+end
--- a/app.rb
+++ b/app.rb
@@ -5,6 +5,7 @@ require_relative 'lib/peep'
 class Chitter < Sinatra::Base
   get '/' do
     @peepdeck = Peep.all
+    puts @peepdeck
     erb :index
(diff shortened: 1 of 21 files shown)


app.rb
```diff
@@ -0,0 +1,22 @@
+require 'sinatra/base'
+require_relative 'lib/peep'
+
+# controller
+class Chitter < Sinatra::Base
+  get '/' do
+    @peepdeck = Peep.all
+    puts @peepdeck
```
micosmin: I think at the end this line can be taken out of the project

lib/peep.rb
```diff
@@ -0,0 +1,38 @@
+require 'pg'
+
+class Peep
+
+  def self.all
+
+    if ENV['ENVIRONMENT'] == 'test'
```
micosmin: I would suggest to move the if else statement in a script that runs when the app loads

Create a file, require the pg gem, and add this if statement.
Require the file at the top of your controller
Delete the if statements

spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+
+# feature 'Viewing home page' do
+#   scenario 'Shows peepdeck' do
+#     visit('/')
+#     expect(page).to have_content 'My first peep!'
+#   end
+# end
+
+feature 'Viewing home page' do
+  scenario 'Shows peepdeck' do
```
micosmin: You can place this scenario inside of the first feature, as it is testing the same feature

spec/peep_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'peep'
+
+describe Peep do
+  describe '.all' do
+    it 'returns Peepdeck' do
+      PG.connect(dbname: 'chitter_app_test')
```
micosmin: I think this should you the database set in the env variable. Not sure this line is needed here anymore. Worth testing

app.rb
jpalmerr: agreed, was there for visibility but forgot to take it out

PR #1178, opened 2019-03-31
Chitter challenge Evelyn

EvelynAleLeo:


Gemfile
```diff
@@ -2,8 +2,11 @@ source 'https://rubygems.org'
 
 ruby '2.5.0'
 
+gem 'capybara'
```
arthurashman: This is a testing gem and can be in the group :test

Gemfile
```diff
@@ -2,8 +2,11 @@ source 'https://rubygems.org'
...
+gem 'pg'
 gem 'rake'
 gem 'rubocop', '0.56.0'
```
arthurashman: This is a testing gem and can be in the group :test

README.md
```diff
@@ -36,98 +25,17 @@ As a Maker
...
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
-
-You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+Ruby
+Capybara
+Sinatra
+PostgreSQL
```
arthurashman: Consider adding some info about how to download and use this code/program

db/migrations/DB.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE "peeps" (
```
arthurashman: This does not set up the TABLE as you expect it to. The "" are not needed, and capitalisation needs checking: 

CREATE TABLE peeps(id SERIAL PRIMARY KEY ...

lib/chitter_model.rb
```diff
@@ -0,0 +1,36 @@
...
+class Chitter
+
+  def self.get_peeps
+
+    if ENV['ENVIRONMENT'] == 'test'
+      con = PG.connect :dbname => 'Chitter_test', :user => 'evelynblazquez'
+    else
+      con = PG.connect :dbname => 'Chitter', :user => 'evelynblazquez'
+    end
+    rs = con.exec "SELECT * FROM peeps ORDER BY datetime"
```
arthurashman: rs is not well named - I do not know from reading it what it is for

app.rb
```diff
@@ -0,0 +1,49 @@
+require 'sinatra/base'
+require './lib/chitter_model'
+
+class Peeps < Sinatra::Base
+
+  enable :sessions
+
```
arthurashman: These routes are confusing and do not seem to be in the right order to gather the params you are expecting to be be gathering

lib/chitter_model.rb
```diff
@@ -0,0 +1,36 @@
+require 'pg'
+
+class Chitter
+
+  def self.get_peeps
+
+    if ENV['ENVIRONMENT'] == 'test'
+      con = PG.connect :dbname => 'Chitter_test', :user => 'evelynblazquez'
```
arthurashman: Not sure if the :user part is necessary. Makes it difficult to run on a machine that's not yours

PR #1185, opened 2019-04-01
Chitter Challenge - Not finished/ slightly abysmal 

LaurenQurashi:


lib/peep.rb
```diff
@@ -0,0 +1,26 @@
+
+require 'pg'
+
+class Peep
+
+  def self.all
+    if ENV['ENVIRONMENT'] = 'test'
```
Hives: you need a double '==' here!

lib/peep.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  def self.create(content:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    connection.exec("INSERT INTO peeps (content) VALUES('#{content}')")
```
Hives: this line fails if the peep `content` contains an apostrophe, beacause it interprets the apostrophe as ending the VALUES string. see if you can work out a way to fix this :)

app.rb
```diff
@@ -0,0 +1,25 @@
...
+  get '/' do
+    erb :homepage
+  end
+
+  get '/feed' do
+    @peeps = Peep.all
+    erb :index
+  end
+
+  get '/feed/create' do
```
Hives: the RESTful way to name this route would be something like `get '/peep/create'`.

the format for the URI is '/resource/action', and what you're doing here is creating a new peep, not a new feed. so it should be '/peep/create'.

app.rb
```diff
@@ -0,0 +1,25 @@
...
+    erb :"feed/create"
+  end
+
+  post '/feed' do
```
Hives: likewise here, it should be `post '/peep'` for the same reason

app.rb
```diff
@@ -0,0 +1,25 @@
...
+  Peep.create(content: params[:content])
+  redirect '/feed'
+end
```
Hives: your indentation has got messed up here. you can run rubocop regularly to pick up things like this

views/index.erb
```diff
@@ -0,0 +1,33 @@
...
+  <head>
+    <link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.3.1/css/bootstrap.min.css" integrity="sha384-ggOyR0iXCbMQv3Xipma34MD+dH/1fQ784/j6cY/iJTQUOhcWr7x9JvoRxT2MZw1T" crossorigin="anonymous">
+    <title>Feed</title>
+    <meta charset="utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
+  </head>
+
+  <body>
+    <div class="container">
+      <div class="alert alert-info" role="alert">
```
Hives: you've got all this stuff duplicated in your index.erb and homepage.erb. not very DRY!

you can create a `layout.erb` file with a `<%= yield %>` command in it, and Sinatra will wrap every page in your app with it. you can use this to extract the stuff that's the same on every page into a single place. i found this lesson very helpful to understand this: <https://learn.co/lessons/sinatra-yield-readme>

README.md
```diff
@@ -131,3 +131,94 @@ SimpleCov.start
...
+[Getting started](#getting-started) | [To Use](#To-Use) | [Task](#Task) | [Approach](#Approach) |
+[Code Status](#Code-Status)
+
+This weekend challenge is for submission to Makers Academy. The code creates a web app that mocks the basic functionality of Twitter.
+
+### Getting Started
+1. Clone to your local machine using
+  `git@github.com:LaurenQurashi/chitter-challenge.git
+2. Run `gem install bundle`
+3. Run `bundle`
```
Hives: you should add something here about how to get the database(s) set up

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+feature 'Viewing Peeps' do
+  scenario 'visiting the index page' do
+    visit ('/')
+    expect(page).to have_content "Chitter"
+  end
+
+  scenario 'A user can see bookmarks' do
+    Peep.create(content: "Peep 1")
+    Peep.create(content: "Peep 2")
+    Peep.create(content: "Peep 3")
```
Hives: feature tests are more about simulating stuff the user can do. so rather than doing `Peep.create` you should use Capybara to simulate entering a peep via the front end, e.g. something like:
```ruby
```

Hives: On second thoughts I'm not totally sure about this... maybe it's fine how you've got it...

Hives: according to [the rubric](https://github.com/makersacademy/chitter-challenge/blob/master/docs/review.md#ensure-correct-location-of-featureacceptance-and-unit-tests), in a feature test:

>  It's acceptable to manipulate the database directly to set things up, but note their are risks associated here, i.e. that you will get your database in a state that it couldn't get into via the web interface, then when the tests pass it may not reflect precisely the user experience of using the site.

lib/peep.rb
Hives: this is **setting** `ENV['ENVIRONMENT']` to `"test"` - that's why your app is connecting to the wrong database.

views/index.erb
```diff
@@ -0,0 +1,33 @@
...
+        <h1>What's Everyone saying?</h1><br>
+      </div>
+      <div>
+        <p class="text-left">
```
Hives: `<p>` tags are to indicate paragraphs of text. i don't think you should use them to contain a `<ul>` element.

views/index.erb
```diff
@@ -0,0 +1,33 @@
+<!DOCTYPE html>
+<html>
+  <center>
```
Hives: `<center>` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/center). see if you can find a way to do the same thing using CSS

spec/spec_helper.rb
```diff
@@ -1,5 +1,24 @@
 require 'simplecov'
 require 'simplecov-console'
+require 'capybara'
+require 'capybara/rspec'
+require 'rspec'
+require_relative './setup_test_database'
+
+ENV['RACK_ENV'] = 'test'
+ENV['ENVIRONMENT'] = 'test'
```
Hives: consider just using the `RACK_ENV` variable to indicate that you're in the test environment. do you need an extra `ENVIRONMENT` variable too?

spec/peep_spec.rb
```diff
@@ -0,0 +1,27 @@
+
+require 'peep'
```
Hives: tests look pretty decent in general

lib/peep.rb
LaurenQurashi: you're an angel, that was ruining my day

PR #1195, opened 2019-04-26
Set up spec_helper and controller plus an index page test

stuartpet:


public/CSS/peep_style.css
```diff
@@ -0,0 +1,53 @@
+.hero-image {
```
jordantroberts: Loving the CSS! :)

spec/features/peep_spec.rb
```diff
@@ -0,0 +1,29 @@
+require 'peeps'
```
jordantroberts: Code has been thoroughly feature tested 👍

PR #1204, opened 2019-04-28
Chitter challenge - first 3 user stories complete 

jordantroberts:
Have yet to implement the registration feature but the first three user stories are complete :)


jordantroberts:
@stuartpet Thanks for the CSS tips! 😃

PR #1215, opened 2019-04-28
Chitter-challenge - Tara

tsankhalpara:
Please see the readme to show the areas of the challenge I have struggled with and have not been able to complete.


views/chitter/index.erb
```diff
@@ -1,8 +1,17 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <title></title>
+    <style>
```
NapperJLG: To keep your view files a bit cleaner, you can create a seperate CSS file and at a link to it in the html. Dom sent me this for the RPS game last week which helped a bit: https://github.com/webdevjeffus/css-for-sinatra#putting-it-in-place

spec/features/sign_up_spec.rb
```diff
@@ -5,4 +5,10 @@
     visit('/')
     expect(page).to have_button 'Sign up'
   end
+
+  scenario 'Allow user to enter details' do
```
NapperJLG: These two feature test probably could have been merged into one. In this test, capybara is clicking the signup button in order to let the user enter details. If the page didn't have the sign up button (which the first feature test is testing for), then the test would still fail, so you're still covering yourself.

PR #1220, opened 2019-04-29
Partial solution - Calum Darroch

CalumDarroch:


spec/units/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'peep'
+
+describe Peep do
+
+  describe '.all' do
+    it 'returns a full list of peeps' do
```
ayelishaamoah: Have you considered adding the peeps you are testing for within the test itself

Calums_README.md
```diff
@@ -0,0 +1,16 @@
+## APPROACH
+
+In this challenge, I have followed a similar approach to that which we undertook in the weekly challenge of building a bookmark manager app.
+
+
+
+
+### To set up the database
+
+ Connect to `psql` and create the `peeps` database:
```
ayelishaamoah: I noticed the database name is the same as the table name - should this be 'chitter'?

lib/peep.rb
```diff
@@ -0,0 +1,16 @@
+require 'pg'
+
+class Peep
+
+  def self.new(content)
```
ayelishaamoah: Add the option to choose tests database

PR #1221, opened 2019-05-26
Chitter Challenge

EdinaBMakers:


spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,22 @@
+feature 'Viewing peeps' do
```
LaurenceTaylor: Good feature tests, matches the rubric (i.e. not too similar to unit tests)

app.rb
```diff
@@ -0,0 +1,43 @@
...
+    erb(:'peeps/new')
+  end
+
+  post '/peeps/new' do
+    Peep.send(message: params[:message])
+
+    redirect('/peeps')
+  end
+
+  get '/users/new' do
```
LaurenceTaylor: Could split routes into separate controller files?

public/chitter.css
```diff
@@ -0,0 +1,54 @@
+.topbar {
```
LaurenceTaylor: Good idea to keep the css in a different file

views/users/new.erb
```diff
@@ -0,0 +1,19 @@
+<form action='/users' method='post'>
```
LaurenceTaylor: 'clean' HTML with correct labels

lib/peep.rb
```diff
@@ -0,0 +1,52 @@
...
+  def eql?(other_peep)
+    return false if other_peep.nil?
+    return false unless other_peep.is_a?(Peep)
+
+    @id == other_peep.id \
+    && @message = other_peep.message \
+    && @sent_time = other_peep.sent_time
+  end
+
+  def ==(other_peep)
```
LaurenceTaylor: What do these two methods do?

PR #1230, opened 2019-05-28
Partial Solution to Chitter Challenge

samkitchen94:


spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,16 @@
+feature "viewing peeps" do
+  scenario "it allows you to view a list of peeps" do
+    peep_1 = Peep.create(message: "First peep")
+    peep_2 = Peep.create(message: "Second peep")
+    peep_3 = Peep.create(message: "Third peep")
+
+    visit '/chitter'
+
+    expect(page).to have_content("Third peep")
```
hughbric: There might be a way to test if this is the first element on the web page. This would test if the peeps are in reverse order with most recent first. Something like, `expect(first('.box')).to have_content("Third peep)`

lib/peep.rb
```diff
@@ -0,0 +1,37 @@
...
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+  peeps = connection.exec("SELECT * FROM peeps ORDER BY sent_time DESC;")
+  peeps.map { |peep| to_peep(peep) }
+  end
+
+  def self.create(message:)
+    if ENV['ENVIRONMENT'] == 'test'
```
hughbric: You can refactor these into separate files so you DRY out the connections.

PR #1241, opened 2019-06-07
Completed chitter challenge

kks110:


app.rb
```diff
@@ -0,0 +1,80 @@
...
+    end
+    redirect('/')
+  end
+
+  get '/sign_up' do
+    erb :sign_up
+  end
+
+  post '/sign_up' do
+    if params[:name].length > 24
```
lindseydew: These if statements could be abstracted to a `validation` module

lib/login_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+    if User.user_exists?(username)
+      user = User.find_by_username(username)
+      if correct_password?(password, user)
+        true
+      else
+        false
+      end
+    else
+      false
+    end
```
lindseydew: This method could be simplified to:
``` 
     user = User.find_by_username(username)
     user && correct_password?(password, user)
```

PR #1242, opened 2019-06-07
Not finished but one to many relations is set up

AnnaWeakclaw:


lib/user.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  def self.find(email)
+    find_user = DbConnection.query("SELECT * FROM users WHERE email = '#{email}';")
+    User.new(email: find_user[0]["email"], id: find_user[0]["id"], name: find_user[0]["username"]) 
+  end
+
+  def self.find_by_id(id)
+    return nil unless !id.nil?
+    find_user = DbConnection.query("SELECT * FROM users WHERE id = #{id};")
```
lindseydew: Might want to check the DB was able to retrieve a record for this `id` before serializing into your model

PR #1243, opened 2019-06-08
🐦

peterjcole:
It's on Heroku here: https://peters-peeps.herokuapp.com

See [my readme](https://github.com/peterjcole/chitter-challenge) for details.

Might work on this some more over the weekend, but wanted to see if Travis will work.


lib/peep.rb
```diff
@@ -0,0 +1,15 @@
+require 'data_mapper'
+
+class Peep
+  include DataMapper::Resource
```
lindseydew: Nice use of an ORM well done

spec/features/chitter_spec.rb
```diff
@@ -0,0 +1,157 @@
+feature 'chitter' do
+  before(:each) do
+    Peep.destroy
+    User.destroy
+    create_users
```
lindseydew: Good use of recreating state

PR #1245, opened 2019-06-08
chitter challenge submission

LaurenOM:


lib/user.rb
```diff
@@ -0,0 +1,54 @@
...
+    @username = username
+  end 
+
+  def self.add(name:,username:, email:, password:)
+    DatabaseConnection.query("INSERT INTO users (name,username,email,password) 
+                      VALUES ('#{name}','#{username}','#{email}','#{password}')")
+  end 
+
+  def self.unique_username?(username)
+    result = DatabaseConnection.query("SELECT * FROM users WHERE username = '#{username}'").map { |row| row }
```
lindseydew: Dont think you need to use a `map` here. 
Also, boolean logic can be simplified to `result.isEmpty?`

PR #1265, opened 2019-06-23
chitter challenge - work in progress

petraartep:
please check readme


spec/setup_test_database.rb
```diff
@@ -0,0 +1,7 @@
+require 'pg'
+
+def setup_test_database
+  connection = PG.connect(dbname: 'chitter_test')
```
Danielandro: In README, test database is called `chitter_db_test` - might be why this isn't working

app.rb
```diff
@@ -0,0 +1,37 @@
+require 'sinatra/base'
+require 'sinatra/flash'
+require_relative './lib/peep'
+require_relative '.lib/user'
```
Danielandro: Should be `'./lib/user'`

PR #1281, opened 2019-07-21
Josh Hemsley - Chitter Challenge

hemser1:


app/app.rb
```diff
@@ -0,0 +1,64 @@
...
+    flash[:notice] = 'You have signed out.'
+    redirect '/'
+  end
+
+  get '/signup' do
+    erb :signup
+  end
+
+ post '/users' do
+  user = User.create(email: params['email'], password: params['password'])
```
fvarnals: Is there any benefit to using a symbol instead of a string in the params? In the bookmarks walkthrough they used symbols - i.e. params[:email]

lib/chitter.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  def self.create(nickname, message, post_time)
+    result = DatabaseConnection.query(
+      "INSERT INTO messages
+      (nickname, message, post_time)
+      VALUES('#{nickname}','#{message}','#{post_time}')
+      RETURNING id, nickname, message, post_time;"
+    )
+    Chitter.new(
+      id: result[0]['id'],
```
fvarnals: maybe it would make your code DRYer if you used .first on result above, rather than having to use [0] here?

lib/user.rb
```diff
@@ -0,0 +1,37 @@
...
+    User.new(
+      id: result[0]['id'],
+      email: result[0]['email']
+    )
+  end
+
+  def self.find(id:)
+    return nil unless id
+    result = DatabaseConnection.query("SELECT * FROM users WHERE id = #{id}")
+    User.new(
```
fvarnals: Again could use .first on result here, + above and below?

PR #1283, opened 2019-07-21
Will's basic chitter program

willhowes:


README.md
```diff
@@ -1,25 +1,59 @@
...
+03_add_user_column.sql
+
+TO CREATE TEST DATABASE
 
+1. Connect to psql
+2. Create the database using the psql command CREATE DATABASE chitter_test;
+3. Connect to the database using the pqsl command \c chitter_test;
+4. Run the queries saved in the following files under ```db/migrations```: 01_create_peeps_table.sql
+02_create_users_table.sql
+03_add_user_column.sql
```
riannemcc: Big fan of your detailed README! Inspo for next weekend...

db/migrations/01_create_peeps_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, peep VARCHAR(280), time TIMESTAMP);
```
riannemcc: This TIMESTAMP business seems pretty handy. Is it a straightforward way of getting the time on the peeps?

spec/features/log_in_spec.rb
```diff
@@ -0,0 +1,28 @@
+# As a Maker
+# So that only I can post messages on Chitter as me
+# I want to log in to Chitter
+
+feature 'log in' do
+  scenario 'a user can log in' do
+    User.create(email: "test@example.com", password: "password123", username: "test_user")
+
+    visit('/sessions/new')
```
riannemcc: You could use a helper for the log in process to avoid repeating yourself in these tests

db/migrations/02_create_users_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE users (id SERIAL PRIMARY KEY, email VARCHAR(60), password VARCHAR(140)), username VARCHAR(60);
```
riannemcc: There's a bracket in the wrong place here which prevents you from setting up database correctly. 
But I managed it and the site looks great!

PR #1288, opened 2019-07-21
Freddie - Chitter

fvarnals:


spec/Features/order_peeps_by_time_spec.rb
```diff
@@ -0,0 +1,14 @@
+# As a maker
+# So that I can see what others are saying
+# I want to see all peeps in reverse chronological order
+feature "View messages in reverse chronological order" do
```
hemser1: Really like this test, clever way of doing it!

spec/setup_test_database.rb
```diff
@@ -0,0 +1,9 @@
+require 'pg'
+
+p 'Setting up database...'
+
+def setup_test_database
+  connection = PG.connect(dbname: 'chitter_test')
```
hemser1: Could we have created a DatabaseConnection class and then call the class each time so you don't have to repeat the PG.connect process?

app/models/peep.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def self.list
+    result = connect_to_database.exec('SELECT message_text FROM peeps;')
+    result.map do |peep|
+      Peep.new(id: peep['id'], text: peep['message_text'])
+    end
+  end
+
+  def self.connect_to_database
```
hemser1: Could we have pulled this into a new file? Just inline with SRP :D

spec/units/peep_spec.rb
```diff
@@ -0,0 +1,15 @@
+require './app/models/peep.rb'
+require 'database_helpers'
+
+describe "#post" do
+  it "Should allow user to post a peep" do
+    message = 'This is my first peep'
+    peep = Peep.post(message: 'This is my first peep')
+    connection = PG.connect(dbname: 'chitter_test')
+    p persisted_data = persisted_data(id: peep.id)
```
hemser1: might want to remove the P

PR #1290, opened 2019-07-22
it posts to a feed at least

Kaymo1990:
Able to post and see posts, didn't attempt beyond that.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,23 @@
+require 'sinatra/base'
+require './lib/cheets.rb'
+
+class Chitter < Sinatra::Base
+
+  enable :sessions
+
+  get ('/') do
+    @cheets = Cheets.all
+    erb :index
+  end
+
+  get ('/newpost') do
+  end
+
+  post ('/newpost') do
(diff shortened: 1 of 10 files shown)


lib/cheets.rb
```diff
@@ -0,0 +1,34 @@
...
+end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'cheet_base_test')
+    else
+      connection = PG.connect(dbname: 'cheet_base')
+    end
+    result = connection.exec("SELECT * FROM chitter")
+    result.map { |cheet| cheet['cheet'] }
```
ckettell: Could you add time in here and format?

app.rb
```diff
@@ -0,0 +1,23 @@
...
+class Chitter < Sinatra::Base
+
+  enable :sessions
+
+  get ('/') do
+    @cheets = Cheets.all
+    erb :index
+  end
+
+  get ('/newpost') do
```
ckettell: Could lose this route?

PR #1293, opened 2019-07-22
Rianne's Chitter Challenge (partial)

riannemcc:
Only managed to reach user story 2 due to being away for 4 days. One test currently failing due to the formatting of the Time, should be rectified easily.


docs/user_stories.md
```diff
@@ -0,0 +1,40 @@
+STRAIGHT UP
+
+As a Maker
+So that I can let people know what I am doing  
+I want to post a message (peep) to chitter
+
+* visiting page /peeps/new
```
willhowes: I like this - nice and clear what to do to post a peep.

public/style.css
```diff
@@ -0,0 +1,68 @@
+body {
```
willhowes: good work with the css and putting it into a separate style sheet. It took me far too long to work out how to do this!

spec/peeps_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+      peep = Peeps.create(peep: 'Test', time: Time.now)
+      persisted_data = persisted_data(id: peep.id)
+      expect(peep).to be_a Peeps
+      expect(peep.id).to eq persisted_data['id']
+      expect(peep.peep).to eq 'Test'
+    end
+
+    it "Shows time each peep was created" do
+      Peeps.create(peep: 'Today is Sunday', time: Time.now)
+      time = Time.now
```
willhowes: I guess what you could do here, to make sure that the time of the test peep is the same is the expected result, is to set the time manually, i.e. time: 2019/01/01?

views/peeps.erb
```diff
@@ -0,0 +1,26 @@
...
+            <form action="/peeps/<%= peep.id %>" method="post">
+             <input type='hidden' name='_method' value='DELETE'/>
+             <input type="submit" value="Delete" />
+           </form>
+          </li>
+        <% end %>
+      </ul>
+
+    <form action="/" method="get">
+    <button type="submit" value="return">Return to homepage</button>
```
willhowes: Really like that you have the option to go back to the home page.

PR #1294, opened 2019-07-22
Genny Chitter Challenge

gennyallcroft:


lib/database_connection.rb
```diff
@@ -0,0 +1,18 @@
+require 'pg'
+
+class DatabaseConnection
```
LeoRoma: Creating a DatabaseConnection class is a very nice idea to keep the code refactored.

lib/user.rb
```diff
@@ -0,0 +1,28 @@
...
+require_relative 'database_connection'
+require_relative 'database_connection_setup'
+
+class User
+
+  attr_reader :id, :name, :email, :username, :password
+
+  def initialize(id: id, name: name, email: email, username: username, password: password)
+    @id = id
+    @name = name
```
LeoRoma: Using name or username, one is enough

PR #1365, opened 2019-10-25
Update README.md

lukegittins01:


lib/app.rb
```diff
@@ -11,4 +11,15 @@ class Chitter < Sinatra::Base
...
   end
+
+  get '/chitter_feed/new' do
+    erb :new
+  end
+
+  post '/chitter_feed' do
+  content = params['content']
+  connection = PG.connect(dbname: 'chitter_test')
+  connection.exec("INSERT INTO feed (content) VALUES('#{content}')")
```
zero201926: these could be in another file to keep it thin

spec/chitter_feed_spec.rb
```diff
@@ -0,0 +1,15 @@
+describe '.content' do
+  it 'returns the chitter feed content' do
+    connection = PG.connect(dbname: 'chitter_test')
+
+    connection.exec("INSERT INTO feed (content) VALUES ('This is message by Luke');")
+    connection.exec("INSERT INTO feed (content) VALUES ('This is another message by Steve');")
+    connection.exec("INSERT INTO feed (content) VALUES ('This is the Chitter Challenge');")
+
```
zero201926: I like that you are adding more than one value, so the tests are accurate.

david-mears:
> steps to get code to to work on another computer:
> in sql(
> CREATE DATABASE chitter
> CREATE DATABASE chitter_test
> CREATE TABLE feed(id SERIAL PRIMARY KEY, content VARCHAR(280));
> )

Thanks a lot for this @zero201926 
It would have been great if the author had put this info in the README

spec/features/testing_redirect_feed_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'testing redirect' do
+  scenario 'it must redirect from / to /chitter_feed' do
+    visit '/'
+    expect(page).to have_content("We have redirected to the Feed")
```
david-mears: Or you could make an expectation about the current url

lib/app.rb
lukegittins01: Yeah that would be a good idea, will make sure I approach it like that in future projects 👍

spec/chitter_feed_spec.rb
lukegittins01: Thank you yeah It good to test multiple things

spec/features/testing_redirect_feed_spec.rb
lukegittins01: Ok yeah that would work much better in terms of what is displayed when that page loads. Thanks David 😃

PR #1366, opened 2019-10-25
Completed first user story - ready to implement a database

Will-Bath:


david-mears:
`bundle install` doesn't install capybara, so I'm guessing you have it installed globally. As a user of this repo, I had to add capybara to the Gemfile, than run `bundle install`.

david-mears:
Consider making commits more often!

PR #1368, opened 2019-10-25
first user story done

adamr27:


david-mears:
It's hard to run this on my computer because I don't know which databases and tables and columns to create. For future reference, it was database `Peeps`, table `user_peeps`, with a column `peep`. This is the kind of thing it's useful to put in your Readme.

PR #1369, opened 2019-10-25
Done a lot but no tests

Adam1824:
I got really carried away with myself, the code engulfed me, every functionality about the web app works though :) (I think lol)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -2,6 +2,11 @@ source 'https://rubygems.org'
 
 gem 'rake'
 gem 'rubocop', '0.71.0'
+gem 'sinatra'
+gem 'capybara'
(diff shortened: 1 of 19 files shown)


david-mears:
It would be very helpful to leave a description of how the databases were set up, with the tables and columns etc (for example leaving the exact SQL messages), so that it's easy to run this on another computer. Also in your README (or the index page?) you could direct us to the urls where we can find your functionality - i.e. `[localhost]/chitter`

david-mears:
Two more recommendations: try to get in the habit of committing more often, with messages that specifically tell people what you changed (e.g. 'add peep view erb'). And the second, you can probably guess - installing test-driving as a habit.

PR #1371, opened 2019-10-25
Sign up + broken peeps

2Blagden:
Not happy with where my functionality with writing / displaying peeps, spent too long on the signup :(


jamesbingham:
I ran his program and discussed with him that he could add the time feature when he has time to do it!

jamesbingham:
We also discussed that he could maybe get his peep messages working as that was a feature that wasnt working when I ran his program

setup.txt
```diff
@@ -0,0 +1,35 @@
+CREATE TABLE tweets(tweet_id SERIAL PRIMARY KEY, user_name INTEGER, FOREIGN KEY(user_name) REFERENCES users(user_name), tweet VARCHAR(240), time TIMESTAMP);
```
david-mears: I tried following this, which looks very helpful to have, but I think it's in the wrong order because I get `ERROR: relation "users" does not exist` (relation means table)

david-mears:
Perhaps something to look into is why does this allow me to sign up with a username `1` but not with a username `example man`?

```
PG::InvalidTextRepresentation - ERROR:  invalid input syntax for integer: "as"
LINE 1: ...user_name, email, password) VALUES('example man', 'as', 'exa...
                                                             ^
```

setup.txt
```diff
@@ -0,0 +1,35 @@
...
+
+
+CREATE TABLE users(username SERIAL PRIMARY KEY,);
+                                                       ^
+twitter_clone=# CREATE TABLE users(username SERIAL PRIMARY KEY);
```
david-mears: what does `SERIAL` do?

PR #1372, opened 2019-10-28
uploaded files

zt3h:
Not too happy on progress this challenge but I will continue to progress on it over next week, Messages are broken :(


2Blagden:
Nice one, as an improvement if you have any extra time, you could develop the timestamp alongside your peeps and make sure they are listed correctly. Besides that it's all really good

david-mears:
There are two fixes you need in order to make messages visible: one is a pesky `=` on the erb (which I'm giving you because it's so pesky. The difference between is `<%` and `<%=` is small but important). The other is about where you define `@message`. It needs to be 'available' or 'in scope' when you load your page.

david-mears:
Was this working at some point? If you commit at strategic times, i.e. when it was working, you (or a reviewer) can roll back to those points. Especially if you name the commits after the new functionality (e.g. `Added peeps view`)

david-mears:
> It was working at one stage but I didnt commit it as in it was able to display the messages but I think I changed the layout of it and it just broke so when I committed the project most of it was broken.

Cool. With the next project I'd recommend committing more often for this reason!

PR #1373, opened 2019-10-28
TObias Pull req

TobiasBenjaminYoungCGI:


lib/Peep.rb
```diff
@@ -1,4 +1,23 @@
 require 'pg'
 class Peep
+    def self.all
+      if ENV['ENVIRONMENT'] == 'test'
+        connection = PG.connect(dbname: 'Chitter_test')
+      else
+        connection = PG.connect(dbname: 'Chitter')
+      end
+      result = connection.exec("SELECT * FROM peep;")
+      result.map { |peep| peep['content'] }.reverse
```
Will-Bath: Good use of .reverse at this point - I personally used it in the HTML form, but this feels like a good place to do the reverse

PR #1377, opened 2019-10-28
not really working

luke-jones-1:
model nearly complete view and controller not so much


app.rb
```diff
@@ -0,0 +1,33 @@
...
+
+  post '/' do
+    session[:peeps] << params[:peep]
+    redirect '/peeps'
+  end
+
+  get '/peeps' do
+    @peeps = session[:peeps]
+    erb :index
+  end
```
EarthWormGym: Great use of sessions for refactoring, very concise and readable code.

PR #1379, opened 2019-11-08
Rhys' Partial Solution

rhyse1:
Everything just broke and the outputted array from the SQL query just stopped working with my data parser, although I have completed my goal of actually using feature tests and capybara for testing.

### DIFF
--- a/spec/creating_post_spec.rb
+++ b/spec/creating_post_spec.rb
@@ -1,0 +1,22 @@
+require 'capybara/rspec'
+
+feature 'user can create a new post' do
+	scenario "User logs in" do
+		visit '/'
+		fill_in('password', :with => 'rhys@rhys.com')
+		fill_in('password', :with => 'test')
+
+		click_button('submit')
+
+		visit '/create'
+
+		fill_in('post', :with => 'hello from tests')
+
+		click_button('submit')
+
+		visit '/view'
+
+		expect(page).to have_content 'hello from tests'
+
+	end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -2,8 +2,10 @@ source 'https://rubygems.org'
 
 gem 'rake'
 gem 'rubocop', '0.71.0'
+gem 'sinatra'
 
 group :test do
+  gem 'capybara'
   gem 'rspec'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
--- a/Gemfile
+++ b/Gemfile
@@ -3,6 +3,7 @@ source 'https://rubygems.org'
 gem 'rake'
 gem 'rubocop', '0.71.0'
 gem 'sinatra'
+gem 'pg'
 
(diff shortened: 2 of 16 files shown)


Stupot83:
Code is well structured and well laid out, with a nice use of CSS. Feature tests have been used well, with one failing test currently. Am unsure if this is me not creating the right things in the database and tables. Given the time restraint on the project of just one day, I would recommend unit tests for the future as this would help diagnose the bug that is stopping the app from working correctly at the moment. It is clear that the intention is there to develop further features and that the routing has been well planned for the next stages. The database helper extracts the database logic so that less code needs to be repeated - obeying the DRY principle.

spec/creating_post_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'capybara/rspec'
+
+feature 'user can create a new post' do
+	scenario "User logs in" do
+		visit '/'
+		fill_in('password', :with => 'rhys@rhys.com')
+		fill_in('password', :with => 'test')
```
david-mears: Each time you write the code to feature test the login, you’re filling in the password field twice, instead of the email and the password fields!

david-mears:
Thanks for the review @Stupot83 .

My main takeaway is to leave a paper trail (migrations) about how you set up the database (or even better, a Rakefile that automatically creates it), as I don’t think I’ve been able to get the same setup as you and so I can’t get the website working even on your earlier commits, though it seems Stuart had some more success (unclear how much!)

I noticed that the test for logging in was passing green, even though when I tried it manually, it didn’t work (got an error page). So the test must be not right, if it’s passing in that case.   The problem in the test was quite annoying, such that I’m more inclined to blame Capybara for passing, than you for writing a bad test. The problem was that it seemed that `page.to have_content ‘Rhys’` was true, even for the error page! I ended up saving the html of the error page to a file so I could search it for ‘rhys’. When I did that, I found the following:  ``` <li class="post-context-line"><code>    expect(page).to have_content &quot;Rhys&quot;
</code></li>
```
and: 
```
<li class="pre-context-line"><code>   fill_in(&#x27;password&#x27;, :with =&gt; &#x27;rhys@rhys.com&#x27;)
</code></li>
                
                <li class="pre-context-line"><code>   fill_in(&#x27;password&#x27;, :with =&gt; &#x27;test&#x27;)
</code></li>
```
It looks like the test is finding its own code on the error page (caused by the unknown error), and then counting that as finding the text it was looking for! That’s probably happening in your other tests too. That suggests that we need a test that’s a bit more sophisticated than looking for a string. It could look for an element and then check the contents of that, for example: expect(page).to have_selector '#test_id'

I went back to the first commit where the login test passed (called ‘passed login test’). There, I see the rspec test passing, but still get an error when I try to do it manually (and get an error when ). I presume it was working for you, so I don’t know where our setups are different.

PR #1381, opened 2019-11-08
Marc's Chitter Challenge

marcb1988:


app.rb
```diff
@@ -0,0 +1,24 @@
+require "sinatra/base"
+require "./lib/peep"
+
+class Chitter < Sinatra::Base
```
jtsalva: This file is very nice, you haven't crammed a bunch of logic into here which is good.

lib/peep.rb
```diff
@@ -0,0 +1,18 @@
+require "pg"
+
+class Peep
+  def self.all
+    ENV["ENVIRONMENT"] == "test" ?
```
jtsalva: You could write this instead as

`connection = PG.connect(dbname: ENV["ENVIRONEMENT"] == "test" ? "chitter_test" : "chitter")`

can't decide which is cleaner

spec/spec_helper.rb
```diff
@@ -1,10 +1,28 @@
...
+require "simplecov"
+require "simplecov-console"
+require "capybara"
+require "capybara/rspec"
+require "rspec"
+
+ENV["RACK_ENV"] = "test"
+ENV["ENVIRONMENT"] = "test"
+
+RSpec.configure do |config|
```
jtsalva: This is nice, automatically setting up database before each test.

lib/peep.rb
david-mears: Or even store the database connection in a separate file, and then require it in! (this is the Interface Segregation Principle)

app.rb
```diff
@@ -0,0 +1,24 @@
...
+    redirect :'/homepage'
+  end
+
+  get "/homepage" do
+    @peeps = Peep.all
+
+    erb :'chitter/homepage'
+  end
+
+  get "/new" do
```
david-mears: RESTful routing would be '/peeps/new' or '/peep/new'

david-mears:
This is a really good submission :) and some good comments from @jtsalva

PR #1382, opened 2019-11-08
Bradleys Chitter challenge

bradleyplater:


jade-genevieve:
Looks very classy and sleek. ✨The navigation bar and the forms look great. You could maybe simplify some of the form validation using `required` in the HTML markup instead, which would prevent a submit and the redirect meaning a user having to refill the parts they did. A Readme and a database migrations folder would also be beneficial. Very nice to see the User class working with the Peep class and a good staggering of commit breaks. 🌟

lib/peep.rb
```diff
@@ -0,0 +1,35 @@
...
+    @time = Time.now
+    @title = title
+    @body = body
+    @username = username
+  end
+
+  def self.get
+    @posts = DBhelper.select
+    posts_arr = []
+    @posts.each do |post|
```
lindseydew: You could simplify the creation of an empty array which you modify by changing the `each` method to a `map`

`each` is a method which will perform an action on each item of the list
`map` is a method which applies a _transformation_ of each item of the list and then returns the new list.

lib/user.rb
```diff
@@ -0,0 +1,39 @@
...
+      password = user['password']
+      
+      users = CurrentUser.new(username,email,password)
+      user_arr << users
+      #user = CurrentUser.new(@user[0]['username'], @user[0]['email'], @user[0]['password'])
+    end
+    return user_arr
+  end
+
+  class CurrentUser
```
lindseydew: Not sure you need a separate class for this? Could this initialize be in user?

app.rb
```diff
@@ -0,0 +1,89 @@
...
+    redirect '/home'
+  end
+
+  post '/redirect/check_user' do
+    username = params[:username]
+    password = Digest::MD5.new << params[:password]
+
+    user = User.get_user(username)
+
+    begin
```
lindseydew: I'd be tempted to put this logic into a helper method in the user class which returns true/false for a validUser

app.rb
```diff
@@ -0,0 +1,89 @@
...
+    body = params[:body]
+    username = session[:user].username
+
+    DBhelper.new_peep(Peep.new(title, body,username))
+    redirect '/home'
+  end
+
+  post '/redirect/check_user' do
+    username = params[:username]
+    password = Digest::MD5.new << params[:password]
```
lindseydew: Nice!

capybara-201911081053022232024549.html
```diff
@@ -0,0 +1,21 @@
+<html>
+  <head>
```
lindseydew: Ooh probably shouldn't commit capybara output?

PR #1383, opened 2019-11-08
David's chitter-challenge

DMartinBGL:
lost track of everything during this challenge, epic fail


lib/Peep.rb
```diff
@@ -0,0 +1,14 @@
+require 'pg'
+
+class Peep
+  def self.all
+    if ENV['Environment'] == 'test'
```
thorpelawrence: Might be better if `ENVIRONMENT` is in all caps, which helps to clearly indicate that it's an environment variable

db/migrations/01_create_peeps_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, usr VARCHAR(20), usrname VARCHAR(20), comment VARCHAR(60) );
```
thorpelawrence: Good idea to separate out the SQL statements so that they can be reused :+1:

app.rb
```diff
@@ -0,0 +1,17 @@
+require 'sinatra/base'
+
+class Chitter < Sinatra::Base
+
+  get '/' do
+    "Welcome to Chitter"
```
thorpelawrence: Could this use a view? Maybe could have consistent design/template across multiple views keeping the code in the controller uncluttered

lib/Peep.rb
```diff
@@ -0,0 +1,14 @@
...
+      connect = PG.connect(dbname: 'chitter_manager_test')
```
thorpelawrence: You stored the value into `connect` but later in the code use `connection`, this would potentially result in a `nil` error on line 8

One possible other way to do it could be (so that it always has a value)
```rb
db_name = ENV['Environment'] == 'test' ? 'chitter_manager_test' : 'chitter_manager'
connection = PG.connect(dbname: db_name)
```

app.rb
```diff
@@ -0,0 +1,17 @@
...
+  end
+
+  get '/peep' do
+    erb :peep
+  end
+
+  post '/peeproom' do
+    Peep.all
```
thorpelawrence: It doesn't look like you're using it right now, but in case the value was needed, this statement on its own would do nothing as the value is not being stored anywhere

```rb
@peeps = Peep.all
erb(:peeproom)
```
Would let you access e.g. `<%= @peeps %>` in the view

lib/Peep.rb
david-mears: This is on the right track towards the fix. The next error message I got was:

```
PG::UndefinedTable:
       ERROR:  relation "comment" does not exist
       LINE 1: SELECT * FROM comment;
```

spec/peep_spec.rb
```diff
@@ -0,0 +1,10 @@
+require './lib/peep'
+
+describe '.all' do
+  it 'returns a list of peeps' do
+    peeps = Peep.all
+
+    expect(peeps).to include "My first peep"
+    expect(peeps).to include "My second peep"
```
david-mears: Why do you expect it to include these?

david-mears:
I *think* the annoying reason it was passing when it shouldn't be was that the test caused a page to have an error - and the error page included the code for the test when describing the error. So the code `    expect(page).to have_content "A load of nonsense.."` *found itself*.

```html
<li class="post-context-line"><code>    expect(page).to have_content &quot;A load of nonsense..&quot;
</code></li>
```

This seems to be an issue with using string-checking with Sinatra - one way around would be to look for the contents of a specific element.

app.rb
david-mears: > NameError at /peeproom
> uninitialized constant Chitter::Peep

david-mears: By the way, a POST route should just act on things (e.g. models/databases), not serve up views - it should redirect to a GET route. That prevents the problem where if you reload the page you send another POST request (and accidentally buy the same sofa twice).

PR #1384, opened 2019-11-08
Initial pull request for chitter challenge

robbaile:


swade19:
Clear test driven commits with a nice looking webpage as the result :) 
Code clearly laid out, refactored neatly and well tailored to the user stories.
Thanks for showing me it up and running, looks really good and felt like using the real thing. 


Only suggestion is to include DB setup instructions. 
Nice job, Rob! 



To set up DB :  

CREATE TABLE "public"."messages" ("id" serial,"user_id" int4,"text" varchar,"time" timestamp, PRIMARY KEY ("id"));

CREATE TABLE "public"."users" ("id" serial,"email" varchar,"password" varchar, PRIMARY KEY ("id"));

spec/message_spec.rb
```diff
@@ -0,0 +1,16 @@
+require 'pg'
+require 'message'
+require 'user'
+
+describe Message do 
+    it 'should return a list of messages' do 
+        User.create('rob@rob.com', 'rob')
+        id = User.find('rob@rob.com', 'rob').first.id
+        time = Time.new
+        Message.create(id, 'Hello it is rob', time)
```
lindseydew: Could make use of mocking and test the time that gets set on a peep here?

PR #1385, opened 2019-11-08
initial PR

khan2597:


Slongboro:
Easy to read code and good test coverage. Easy to follow naming conventions. Nice work :)

david-mears:
Nice commit messages - clear and informative. The tests look good, but I can't run them because I can't replicate your database setup. That means I can only see your index page - which, I admit, does look slick.

My main criticism would be that there's no way to work out your database set up. I can see from your test that your messages table has three columns, and the first two of them are numbers, but it's unclear what those numbers represent. If you include a migrations file, you can then ensure your project is reproducible on another machine - and you can even set up a rake task to take care of it: `rake db:setup`

PR #1386, opened 2019-11-08
Chitter

6JoeB:


KrisN64209:
To set up databases:

CREATE DATABASE chitter_peeps_manager;

then

CREATE DATABASE chitter_peeps_test;

and in both databases

CREATE TABLE chitter_peeps(peep VARCHAR(160), username VARCHAR(30), dateposted VARCHAR(20), timeposted VARCHAR(10));

KrisN64209:
Hi Joe, good job. Code is clear and readable, and works. But, one of your test fails (it's only a typo though), and you haven't tested for your .all method. Also, I don't think your test database is being truncated after your tests, so you're filling it up every time you Use RSpec, which will stop you being able to test that multiple peeps are being made/stored when you add that.

david-mears:
Thanks for the review @KrisN64209 , and the database set up instructions!

In case you feel like some detective work / debugging practice, I did some debugging and found that the problem with the failing test was on line 40 of peep.rb: 
```
    Peep.new(peeptext: result[0]["peeptext"], username: result[0]["username"], dateposted: result[0]["dateposted"], timeposted: result[0]["timeposted"])
```
If you inspect the peep object created by that, you'll see that its peeptext attribute is nil. That suggests a problem with `result[0]["peeptext"]`.

spec/features/adding_peeps_spec.rb
```diff
@@ -0,0 +1,10 @@
+feature "adding a new peep" do
```
david-mears: Good feature test

spec/unit/chitter_spec.rb
```diff
@@ -0,0 +1,13 @@
+require "peep"
+
+describe Peep do
+  describe ".create" do
+    it "creates a new peep" do
+      connection = PG.connect(dbname: "chitter_peeps_test")
```
david-mears: This line doesn't do anything, as you don't refer to this `connection` variable in the `it` block.

david-mears:
It would be good to build the habit of testing things before you create them: the homepage route, and the `.all` method. One benefit of that would be to keep your process more aligned with the specifications: you lose time by e.g. building links to pages that weren't requested by the user (who won't pay you more for things that they didn't request).

Overall, this is a nice solution, showing a good understanding of building a simple web app with a database.

PR #1387, opened 2019-11-10
Chitter_challenge

Slongboro:


spec/feature/creating_peep_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'pg'
+
+feature 'Adding a new peep' do
+    scenario 'A user can add a peep' do
+      visit('/peeps')
```
khan2597: This test will pass if it is visit('/new'). As that is the route in app.rb for adding a new peep :).

Good test to check that functionality of the form is working!

spec/feature/chitter_spec.rb
```diff
@@ -4,16 +4,16 @@
...
         describe '.all' do
           it 'returns all messages' do
-            connection = PG.connect(dbname: 'chitter_test')
-            connection.exec("INSERT INTO chitter (message) VALUES ('The zombies are coming');")
-            connection.exec("INSERT INTO chitter (message) VALUES('Bring food');")
-            connection.exec("INSERT INTO chitter (message) VALUES('And Alcohol');")
+            connection = PG.connect(dbname:'chitter_test')
+            connection.exec("INSERT INTO peep (peep) VALUES ('The zombies are coming');")
+            connection.exec("INSERT INTO peep (peep) VALUES('Bring food');")
+            connection.exec("INSERT INTO peep (peep) VALUES('And Alcohol');")
```
khan2597: Good tests to check that database is being populated with new peeps

app.rb
```diff
@@ -0,0 +1,26 @@
...
+      
+      erb :index
+    end
+  
+    post "/peeps" do
+    session[:peep] = params[:peep]
+   
+    # peep = params['peeps']
+    # connection = PG.connect(dbname: 'chitter_test')
+    # connection.exec("INSERT INTO peeps (peep) VALUES('#{peep}')")
```
david-mears: To try and get these three commented out lines to work, I checked your migrations (very useful to have, thanks!) and saw `CREATE TABLE chitter` - so that made sense of the error page which said there was no relation (table) called peeps. I changed `INSERT INTO peeps` to `INSERT INTO chitter`, at which point I noticed in the database that I was able to create peeps from your website, but they were all empty. So I guessed that there was a problem with `#{peep}` here - it was probably empty. I'll leave it to you to wonder about what went wrong on line 15!

app.rb
```diff
@@ -0,0 +1,26 @@
...
+class Chitter < Sinatra::Base
+    enable :sessions
+    
+    get "/" do
+      
+      erb :index
+    end
+  
+    post "/peeps" do
+    session[:peep] = params[:peep]
```
david-mears: Currently this `post` method is returning the output of this line - in a later iteration of the program it would probably do its action on the database and then you would redirect to another page.

README.md
```diff
@@ -1,3 +1,13 @@
+To Create DB's 
+CREATE DATABASE chitter;
+CREATE DATABASE chitter_test;
+
+Create tables in DB's
+CREATE TABLE chitter(id SERIAL PRIMARY KEY, peep VARCHAR(60));
+CREATE TABLE chitter_test(id SERIAL PRIMARY KEY, peep VARCHAR(60));
```
david-mears: I think it's confusing to call your databases and tables the same thing. At the moment, you're asking for the user to create two databases, and then for each of those to create two tables within them with the same names (e.g. the database 'chitter_test' will contain the tables 'chitter' and 'chitter_test').

Databases contain tables, tables contain columns.

lib/chitter.rb
```diff
@@ -1,6 +1,24 @@
 require 'pg'
 
 class Chitter
+    def self.all
+        if ENV['ENVIRONMENT'] == 'test'
+          connection = PG.connect(dbname: 'chitter_test')
+        else
+          connection = PG.connect(dbname: 'chitter')
+        end
+        result = connection.exec('SELECT * FROM peeps')
```
david-mears: `FROM` looks for a table called `peeps`, but peeps is a column, according to your README.

spec/feature/chitter_spec.rb
```diff
@@ -5,15 +5,15 @@
...
+            connection.exec("INSERT INTO chitter_test (peep) VALUES('Bring food');")
+            connection.exec("INSERT INTO chitter_test (peep) VALUES('And Alcohol');")
       
-            chitter = Chitter.all
+            chitters = Chitter.all
       
-            expect(peep).to include("The zombies are coming")
-            expect(peep).to include("Bring food")
-            expect(peep).to include("And Alcohol")
+            expect(chitter_test).to include("The zombies are coming")
```
david-mears: What you mean to test is probably the output of `Chitter.all`.

'chitter_test' hasn't been defined - not in this Ruby program. Postgres (psql) knows about it, but postgres isn't responsible for parsing this line (you aren't using a `connection.exec("some SQL")`). So I expect this throws an 'undefined' error message.

PR #1411, opened 2019-11-11
Week 4 - chitter challenge

sarar0:


app.rb
```diff
@@ -0,0 +1,38 @@
...
+    @peeps = Peep.all
+    erb :index
+  end
+
+  get '/new' do
+    erb :new
+  end
+
+  post '/peeps' do
+    Peep.create(time: Time.now, post: params[:post])
```
topherb99: Maybe let the Peep class instantiate the time?

PR #1414, opened 2019-11-12
Chitter Challenge

swade19:


robbaile:
- Lots of nice bits of code in you solution to be happy with. 

`session[:username] = params[:username]`

- I like the use of session variables to be able to store the username. We can then use these to validate the pages we visit later in the journey.



`result.map { |row| Chitter.new(row["username"], row["peep"], row["time"]) }`
- A really good use of map to be able to create new chitters from the database. This is a really clean approach to create new chitters with data from the database.



```
  scenario "it can allow user to make a peep" do
    visit("/")
    fill_in :username, with: "Sarah"
    click_button "Submit"
    fill_in :peep, with: "Hello World"
    click_button "Submit"
    expect(page).to have_content "Hello World"
  end
```
- Nice and readable test. Suitable for new developers using the code and wanting to see how the application works. 



```
  def initialize(username = "", peep = "", time = nil)
    @username = username
    @peep = peep
    @time = time
  end

```
- We are initialising everything here but setting it automatically to nothing/nil. Do we want any of these to be blank? 



```
  def self.create(peep)
    connection.exec("INSERT INTO peeps_table")
  end
```
- We need to insert the peep into the peeps_table here, specifying the columns and values we want to use.  


```
<h1> Welcome to Chitter! </h1>
<h2>Enter a username to sign in and "peep"</h2>

<form action="/username" method="post">
    <label for="username">
        Username:
        <input type="text" name="username"> 
    </label> 
        <input type="submit" value="Submit"> 
</form>
```
- Your html is nicely laid out and would be acceptable for a screen reader. For example we have only used 1 h1 and 1 h2 on the page so a screen reader can clearly see what is the main piece of content on the page. The form also uses labels and the inputs have values. 
 
Overall, all the code that you have done is readable, concise and well thought out. I think all you needed is a little more time to implement some more features!! Good job

david-mears:
Thanks for the review @robbaile 

> * We are initialising everything here but setting it automatically to nothing/nil. Do we want any of these to be blank?

Using `=` in the parameter definition defines those as the default values, so if they for some reason are not provided, the operation still goes ahead. In practice we'd probably prefer that to produce an error message than to accept the operation with holes in.

spec/features/chitter_spec.rb
```diff
@@ -0,0 +1,8 @@
+require "capybara/rspec"
+
+feature "it can call the website" do
```
david-mears: I think 'call' --> 'visit'.
'Call' applies to methods (functions)

spec/features/chitter_spec.rb
```diff
@@ -1,15 +1,25 @@
...
+  #   end
+
+  scenario "it can allow user to make a peep" do
     visit("/")
     fill_in :username, with: "Sarah"
     click_button "Submit"
-    expect(page).to have_content "Sarah"
+    fill_in :peep, with: "Hello World"
+    click_button "Submit"
+    expect(page).to have_content "Hello World"
```
david-mears: Contrary to your commit message, this doesn't test for db access - it's a feature test, testing that it works from the user's point of view. I think you passed this test by using parameters, not db access.

app.rb
```diff
@@ -19,6 +19,12 @@ class ChitterApp < Sinatra::Base
 
   post "/display_peep" do
     session[:peep] = params[:peep]
+    redirect "/peep_page"
+  end
+  get "/peep_page" do
+    $chitter = Chitter.new
+    @fullPeep = "#{$chitter.username} peeped #{$chitter.peep} at #{$chitter.time}"
```
david-mears: One good reason to follow language-specific spelling conventions is eliminating bugs: `@fullPeep` in your app.rb, but `@full_peep` in your peep_display.erb

lib/chitter.rb
```diff
@@ -0,0 +1,21 @@
...
+  end
+
+  def self.all
+    connection = PG.connect(dbname: "chitterPeeps")
+    result = connection.exec("SELECT * FROM peeps_table")
+    result.map { |row| Chitter.new(row["username"], row["peep"], row["time"]) }
+  end
+
+  def self.create(peep)
+    connection.exec("INSERT INTO peeps_table")
```
david-mears: If you can correct this line, that will count as good evidence for progress in the learning objective of building a simple web app with a database.

PR #1415, opened 2019-11-12
Chitter Challenge

jade-genevieve:


bradleyplater:
Code looks good, could possibly extract some of the database functionality into its own class other than that looks good!

spec/features/view_peeps_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    fill_in("new_chit", with: "This is a Peep")
+    click_button("Peep!")
+  click_button("Peep a chit")
+  fill_in("new_chit", with: "This is another Peep")
+  click_button("Peep!")
+  click_button("Peep a chit")
+  fill_in("new_chit", with: "This is the lucky third Peep")
+  click_button("Peep!")
+
+    expect(page).to have_content "This is the lucky third Peep, This is another Peep, This is a Peep"
```
david-mears: I guess you want to test for these three existing in the page in this order, rather than tying your front-end to the use of commas. Perhaps you could use html id tags which increase in steps of 1, and then test the content of each of those. Or you could use `.to_s.split` on the page's html content.

PR #1416, opened 2019-11-12
Migle's solution

olkras03:
in the README file it would be nice if you could leave some instructions how you set up the project.


olkras03:
You have good layout of folders and files are named correctly.

david-mears:
> in the README file it would be nice if you could leave some instructions how you set up the project.

Yes, it would be really nice - at least there is a migration file!

spec/chitter_spec.rb
```diff
@@ -0,0 +1,22 @@
+require 'chits'
+describe Chits do
+    describe '.all' do
+      it 'returns all chitters' do
```
david-mears: You switch between 'chitter', 'chitters', 'chits' and 'chit' a lot! This would get confusing for someone else reading it.

lib/chits.rb
```diff
@@ -0,0 +1,22 @@
...
+            connection = PG.connect(dbname: 'Chitter')  
+           
+        end    
+     result = connection.exec('SELECT * FROM messages;')
+     
+
+    result.map{ |chitter|
+        puts chitter
+         chitter['messages']
+         }
```
david-mears: I can see you were part way through debugging this method. I found that this method doesn't return anything, by writing `puts chitter` in chitter_spec.rb.

david-mears: Further debugging here:

```
result.map{ |chitter|
   puts "chitter = " + chitter.to_s
   puts "chitter['messages'] = " + chitter['messages'].to_s
}
```
results in:
```
chitter = {"id"=>"61", "message"=>"Hey you! this is Chitter"}
chitter['messages'] =
chitter = {"id"=>"62", "message"=>"Have something to Chit?"}
chitter['messages'] =
chitter = {"id"=>"63", "message"=>"Chit it now!"}
chitter['messages'] =
```

david-mears: I think that's enough clues to start you off. If you'd like more, show me what you've tried.

spec/chitter_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+       
+        connection = PG.connect(dbname: 'Chitter_test')
+
+    # Add the test data
+        connection.exec("INSERT INTO messages (message) VALUES('Hey you! this is Chitter');")
+        connection.exec("INSERT INTO messages (message) VALUES('Have something to Chit?');")
+        connection.exec("INSERT INTO messages (message) VALUES('Chit it now!');")
```
david-mears: This works - I checked the database. 👍

views/messages/chits.erb
```diff
@@ -0,0 +1,4 @@
+<p> Hey you! this is Chitter</p>
+<p> Have something to Chit? </p>
+<p> Chit it now! </p>
```
david-mears: I imagine you were writing this here to make sure that you could write a good test. Do you know how to make this page use data from the database?

PR #1417, opened 2019-11-28
More housekeeping as a result of removing Rake from Bookmark Manager in 2018 

david-mears:
Continuing where [a pull on /course](https://github.com/makersacademy/course/pull/979) left off, and partly fixing https://github.com/makersacademy/course/issues/914.

The review rubric now links to the old Rake content from Bookmark Manager to prevent students being utterly confused as to why something called Rake is being mentioned, and to provide some introduction to what it's for and how to use it.

I moved the Rake stuff further down the review rubric.


jamesjoshuahill:
Thank you @david-mears 🎉

PR #1420, opened 2019-12-02
Jo - Weekly challenge Chitter

JNuckley:


lib/chitter.rb
```diff
@@ -0,0 +1,37 @@
...
+  def self.all
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_db_test')
+    else
+      connection = PG.connect(dbname: 'chitter_db')
+    end
+    result = connection.exec("SELECT * FROM peeps ORDER BY id DESC;")
+    result.map do |peep|
+      Chitter.new(id: peeps['id'], peep: peeps['peep'], posted_at: peeps['posted_at'] )
```
david-mears: This line gave me an error that said there is no such method as `peeps`. So I thought, that's odd, why is this line even trying to call a method? It shouldn't be calling a method at all: we should be cycling through all the things inside `result`, not calling some method somewhere else. It must not recognise the word `peeps`, and so it thinks it's supposed to be a method.

This was the clue that led me to fixing the bug!

PR #1421, opened 2019-12-02
Adding my README file

docela:


lib/peep.rb
```diff
@@ -11,14 +11,18 @@ def self.all
     end
     result = conn.exec("SELECT * FROM peeps;")
     result.map { |peep| peep["peep"] }
+    # result.map { |peep, name|
+    #   peep["peep"]
+    #   peep["name"]
+    # }
```
david-mears: `.map` can only create a output for each input (for example, `peep` goes in, `peep["peep"]` comes out), so you will have to combine these two lookups somehow. You could make a Peep object with `.new`. Or, less complicatedly, tie them together in an array.

views/homepage.erb
```diff
@@ -1,3 +1,3 @@
 <h1>Homepage</h1>
 
-<%= @peeps %>
\ No newline at end of file
+<%= @peeps %>
```
david-mears: This will need to be updated if `Peeps.all` is updated.

PR #1432, opened 2019-12-02
add a heading to the README stories completed

sibamunsanje:


spec/features/sign_up_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature "sign_up" do
+    scenario "a user can sign_up to chitter" do
+      visit ("/sign_up")
+      expect(page).to have_current_path("/sign_up")
```
david-mears: Were you planning to extend this test? I think it should have a different name at this stage in its life, because at the moment it just tests that the page exists. You can change the name to "can sign up" later, but it would be easy to forget to complete this test if you gave it this name, got to green, and then went away.

PR #1434, opened 2019-12-02
Set up project:

SoniaQ:
-download required gem files
 -modify README
-create app.rb
-set up smoke test in features and passing code in app.rb
-set up config.ru
-modify spec_helper to set up project

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,10 +1,16 @@
 source 'https://rubygems.org'
 
+ruby '2.6.5'
+
+gem 'pg'
 gem 'rake'
-gem 'rubocop', '0.71.0'
+gem 'sinatra'
+
 
 group :test do
+  gem 'capybara'
   gem 'rspec'
+  gem 'rubocop', '0.71.0'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
 end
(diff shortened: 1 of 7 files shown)


views/index.erb
```diff
@@ -0,0 +1,16 @@
+<h1>Welcome to Chitter!</h1>
+<br>
+<form action="/peeps" method="post">
+  <input type="text" value="enter peep here" name="enter peep">
```
david-mears: You can use a `placeholder` attribute for 'enter peep here' which greys it out and disappears it when the user starts typing.

views/index.erb
```diff
@@ -0,0 +1,16 @@
...
+  <input type="submit" value="Peep"
```
david-mears: My editor highlights </form> as causing a syntax error. That prevents the submit button from doing anything. Turns out you are missing a `>` on line 5 - now the button submits.

views/index.erb
SoniaQ: Just what I was looking for! Thanks. I'll swap in ```placeholder```.

views/index.erb
SoniaQ: Indeed! Spotted that yesterday morning and changed it. Just haven't pushed code yet. Thanks.

SoniaQ: Code has now been pushed :)

views/index.erb
SoniaQ: Updated.

PR #1439, opened 2019-12-03
Marjan's chitter-challenge

MarjanMagharehi:


spec/setup_test_database.rb
```diff
@@ -0,0 +1,5 @@
+require 'pg'
+p "Setting up test database..."
+
+connection = PG.connect(dbname: 'chitter_challenge_test')
+connection.exec("TRUNCATE chitter;")
```
anhnguyenis: Could also use DROP TABLE here instead of TRUNCATE, which I think may be better if you are changing lots of columns in your table, as TRUNCATE just empties the data in the table as far as I'm aware.

lib/chitter_peeps.rb
```diff
@@ -0,0 +1,30 @@
...
+    else
+      connection = PG.connect(dbname: 'chitter_challenge')
+    end
+    result = connection.exec("SELECT * FROM chitter")
+    result.map {|peep| Chitter_peeps.new(peep['peeps'])}
+  end
+
+
+  def self.create(peep:)
+    if ENV['RACK_ENV'] == 'test'
```
anhnguyenis: Nice and clear 👍 
My next steps would have been to separate the PG.connect lines into another method to DRY out the code.

README.md
```diff
@@ -131,3 +131,5 @@ SimpleCov.start
 ```
 
 You can see your test coverage when you run your tests. If you want this in a graphical form, uncomment the `HTMLFormatter` line and see what happens!
+
+`Stories completed`
```
anhnguyenis: would update your README of where you have got to.

PR #1457, opened 2019-12-09
Chitter challenge

jj49411:


spec/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+describe Peep do
+
+  describe '.all' do
+    it 'should return all peeps' do
+      peeps = Peep.all
+      expect(peeps).to include "Happy Sunday!"
+    end
+  end
+
+  describe '.add' do
```
IngramCapa: This one needs an 'end' :). You could check and see if rspec passes after that.

lib/peep.rb
```diff
@@ -0,0 +1,28 @@
...
+  def initialize(date, content, userID)
+    @date = date
+    @content = content
+    @userID = userID
+  end
+
+
+  def self.all
+    connection = PG.connect(dbname: 'chitter')
+    result = connection.exec("SELECT * FROM peeps;")
```
IngramCapa: "SELECT content, date FROM peeps ORDER BY date DESC;"

to order peep first, date after, date descending

views/index.erb
```diff
@@ -0,0 +1,6 @@
+<link rel="stylesheet" href="https://stackpath.bootstrapcdn.com/bootstrap/4.4.1/css/bootstrap.min.css" integrity="sha384-Vkoo8x4CGsO3+Hhxv8T/Q5PaXtkKtu6ug5TOeNV6gBiFeWPGFN9MuhOf23Q9Ifjh" crossorigin="anonymous">
```
IngramCapa: Super nice views and presentation!!

spec/features/test_infrastructure_spec.rb
```diff
@@ -0,0 +1,6 @@
+feature 'testing infrastructure' do
```
IngramCapa: You could probably do away with this test after you check the infrastructure.

app.rb
```diff
@@ -0,0 +1,44 @@
+require 'sinatra/base'
```
IngramCapa: I would suggest creating a test database when you have the time in order to test functionality

PR #1466, opened 2020-01-19
Chitter-Challenge

ratcliffetj:


lib/user.rb
```diff
@@ -0,0 +1,12 @@
+class User
+
+  def self.add(username)
+    ENV['RACK_ENV'] == 'test' ? connection = PG.connect(dbname: 'chitter_test') : connection = PG.connect(dbname: 'chitter')
+    connection.exec("INSERT INTO users (username) VALUES('#{username}')")
```
alicelieutier: This code should be extracted as it is used in a number of places.

views/send.erb
```diff
@@ -0,0 +1,61 @@
...
+</form>
+<br>
+<br>
+<center>
+  <%if !!@peeps%>
+    <% @peeps.each do |peep| %>
+    <br>
+    <br>
+    <% 
+    user_colour = (Digest::MD5.hexdigest peep.username)[0..5]
```
alicelieutier: This is a lot of logic to live in the views. You could have this code live in the User model, and then have the colour as a method of the user maybe.

PR #1467, opened 2020-01-19
Shaun-chitter-challenge

shaunlsy:


app.rb
```diff
@@ -0,0 +1,31 @@
...
+require_relative './lib/chitter'
+require_relative './lib/user'
+class ChitterManager < Sinatra::Base
+  enable :sessions
+
+  before do
+    @chitter = Chitter
+    @user = User
+  end
+
```
Uvalente: Using REST routes would make it easier to understand the app flow

app.rb
```diff
@@ -0,0 +1,31 @@
...
+  get '/' do
+    erb :form
+  end
+
+  post '/type_peeps' do
```
Uvalente: POST route should be used to change status and then redirect to a GET route to show data

db/migrations/01_create_peeps_table.sql
```diff
@@ -0,0 +1,5 @@
+CREATE TABLE chitters
+(
+  id SERIAL PRIMARY KEY,
+  peeps VARCHAR(60)
```
Uvalente: Does not reflect the column name used in the code

lib/chitter.rb
```diff
@@ -0,0 +1,34 @@
+require 'pg'
+
+class Chitter
+  attr_reader :id, :peep, :dateinserted
+
+  def initialize(id:, peep:, dateinserted:)
+    @id = id
+    @peep = peep
+    @dateinserted = dateinserted
```
Uvalente: Should follow snake_case

lib/chitter.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
Uvalente: The connection to the desired server should be extracted away from the methods

lib/user.rb
```diff
@@ -0,0 +1,33 @@
...
+
+class User
+  attr_reader :username, :password
+
+  def initialize(username:, password:)
+    @username = username
+    @password = password
+  end
+
+  def self.all
```
Uvalente: This method doesn't seems to be ever used in the code

spec/features/see_peeps_reverse_chronological_spec.rb
```diff
@@ -0,0 +1,14 @@
...
+
+feature 'See peeps in reverse chronological order' do
+  scenario 'See latest peep first' do
+    Timecop.freeze(Time.now) do
+      sign_up
+      fill_in "peeps", with: "I'm Shaun"
+      click_button "Tweet"
+      expect(current_path).to eq "/peeps"
+      expect(page).to have_content "I'm Shaun"
+      # expect(page).to have_content (Time.now.strftime('%Y-%m-%d %H:%M:%s')).to_s
```
Uvalente: Commented out code should be removed before a commit

spec/features/sign_up_chitter_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'Sign up for Chitter' do
+  scenario 'Sign up as new user' do
+    sign_up
+    expect(page).to have_content("Welcome to my Chitter!")
+    users = User.all
+    expect(users.first.username).to eq "Shaun"
```
Uvalente: Should check the feature as a user experiences it, without expecting a result from a behind the curtain method

lib/user.rb
```diff
@@ -0,0 +1,33 @@
...
+  end
+
+  def self.create(username:, password:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'users_manager_test')
+    else
+      connection = PG.connect(dbname: 'users_manager')
+    end
+    result = connection.exec("INSERT INTO users (username, password) VALUES('#{username}','#{password}') RETURNING username, password")
+    User.new(username: result[0]['username'], password: result[0]['password'])
```
Uvalente: Should not save the password in a user object for safety reasons

views/index.erb
```diff
@@ -0,0 +1,9 @@
+<h1>Welcome to my Chitter!<h1>
+
+<form action='/peeps' method='post'>
+  <label for="peeps">
```
Uvalente: The for='id' option is not needed because the input is inside the label tag

db/migrations/03_create_users_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE users
+(
+  id SERIAL PRIMARY KEY,
+  username VARCHAR(60),
+  password VARCHAR(60)
```
Uvalente: Could be useful setting a NOT NULL option where is expected to never receive a null value

PR #1468, opened 2020-01-19
Chitter Challenge

acmerlino1:


lib/peep.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+
+  def self.create(user_name:, user_handle:, peep:, timestamp:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'peep_manager_test')
+    else
+      connection = PG.connect(dbname: 'peep_manager')
+    end
```
alicelieutier: Could you find a way to DRY this code?

PR #1469, opened 2020-01-19
Pull request - project not complete

cpcwood:


app/models/email_client.rb
```diff
@@ -0,0 +1,30 @@
+require 'mail'
+require './email_server_settings.rb'
+
+class EmailClient
```
alicelieutier: Why do you have an email client 🤔 ?

app/controllers/app_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+  set :public_folder, Proc.new { File.join(root, "../public") }
+
+  DatabaseConnection.add_details(dbname: 'chitter', user: ENV['USER'], dbms: PG)
+  EmailClient.setup
+  Message.setup(dbconnection: DatabaseConnection, emailclient: EmailClient)
+
+
+  get '/' do
+    erb :homepage
+  end
```
alicelieutier: Where are the other routes?

cpcwood: Only other route required will be '/sign-up' since the rest of the functionaity can be done on the homepage. Not got round to finishing this yet

PR #1473, opened 2020-01-20
Chitter Challenge

nikimanoledaki:


lib/peeps.rb
```diff
@@ -0,0 +1,35 @@
+require 'pg'
+
+class Peeps
```
alicelieutier: Classes name should be singular.

PR #1474, opened 2020-01-20
Chitter pull req

notriapatel:


lib/peep.rb
```diff
@@ -0,0 +1,22 @@
+require 'pg'
+
+class Peep
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
alicelieutier: You may want to DRY this code.

PR #1475, opened 2020-01-20
Chitter Challenge

meghaniankov:


lib/peep.rb
```diff
@@ -0,0 +1,32 @@
...
+    @user_id = user_id
+    @username = username
+    @name = name
+  end
+
+  def self.all
+    DatabaseConnection.setup(environment)
+    query = "SELECT * FROM peeps INNER JOIN users ON peeps.user_id=users.id ORDER BY peeps.id DESC"
+    result = DatabaseConnection.query(query)
+    result.map { |peep| Peep.new(id: peep['id'], text: peep['text'], timestamp: peep['timestamp'], user_id: peep['user_id'], username: peep['username'], name: peep['name']) }
```
alicelieutier: This is too much on just one line. Also, would it work if you did `result.map { |peep| Peep.new(peep) }` ?

meghaniankov: Thank you, this was very helpful. I refactored the code to use Peep.new(peep) by updating my initialize method to take a database row and setting the instance variables based on that row, rather than setting them individually when calling Peep.new(id: row['id'], text: row['text'], ...etc).

PR #1476, opened 2020-01-20
monday morning upload

lucianboatright:


lib/messages.rb
```diff
@@ -0,0 +1,74 @@
...
+    @time = time
+  end
+
+
+  def self.add(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_manager_test')
+    else
+      connection = PG.connect(dbname: 'chitter_manager')
+    end
```
alicelieutier: You may want to DRY this code.

PR #1477, opened 2020-01-20
Mediocre effort

Benauld:
Didn't have a huge amount of spare time to do this this weekend.


Benauld:
pointed Travis at non-test database

lib/peep.rb
```diff
@@ -0,0 +1,27 @@
...
+  def initialize(id:, message:, created_on:)
+    @id = id
+    @message = message
+    @created_on = created_on
+  end
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM peeps ORDER BY created_on DESC")
+    result.map do |peep|
+      Peep.new(id: peep['id'], message: peep['message'], created_on: peep['created_on'])
```
alicelieutier: Could you use `Peep.new(peep)` here? (and in several other places below)

PR #1479, opened 2020-01-20
Chitter-challenge

asiaellis5:


Gemfile
```diff
@@ -5,4 +5,10 @@ group :test do
   gem 'rubocop', '0.71.0'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
+  gem "capybara"
```
nikimanoledaki: Gem files could be split into environments such as test groups.

nikimanoledaki: Also - gems could be listed in alphabetical order

lib/peep.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def self.all
+    result = DatabaseConnection.query('SELECT * FROM peeps;')
+
+    result.map { |peep| self.new(peep['peep'], peep['time']) }
+  end
+
+  def self.add(peep)
+    time = Time.now.strftime("%k:%M")
```
nikimanoledaki: Nice one @asiaellis5 !!! 💯

PR #1480, opened 2020-01-20
All but last user story covered

squareben1:
The most up to date version of this program features ability to register - this is NOT included in this Pull as it is not functioning properly and requires more work. 
This version is fully functioning and covers:

As a Maker

As a maker

As a Maker
So that I can better appreciate the context of a peep
I want to see the time at which it was made

### DIFF
--- a/lib/chitter.rb
+++ b/lib/chitter.rb
@@ -2,13 +2,13 @@ require 'pg'
 
 class Chitter_Peep
 
-  attr_reader :body #, :time, :name, :handle
+  attr_reader :body, :time, :name, :handle
 
-  def initialize(body:) #, time:, name:, handle:)
+  def initialize(body:, time:, name:, handle:)
     @body = body
-    # @time = time
-    # @name = name
-    # @handle = handle
+    @time = time
+    @name = name
+    @handle = handle
   end 
 
   def self.view_all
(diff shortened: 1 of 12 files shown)


lib/chitter.rb
```diff
@@ -0,0 +1,37 @@
...
+    end 
+  end 
+
+  def self.post_peep(body:, name:, handle:)
+    if ENV['RACK_ENV'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else 
+      connection = PG.connect(dbname: 'chitter')
+    end 
+    result = connection.exec("INSERT INTO peeps (body, time, name, handle) VALUES ('#{body}', '#{Time.now.round}', '#{name}', '#{handle}') RETURNING body, time, name, handle;")
```
Marcellofabbri: cool way of changing the time format

spec/units/chitter_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'chitter'
+require 'database_helpers'
+
+describe Chitter_Peep do 
+  describe '#post' do
```
Marcellofabbri: should be '.post' (class method) rather than '#post' (instance method)

spec/units/chitter_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+
+    it 'posts a peep with name & handle' do 
+      peep = Chitter_Peep.post_peep(body: "My first peep", name: "Ben", handle: "squareben")  
+      expect(peep.time).to eq Time.now.round.to_s
+      expect(peep.name).to eq "Ben"
+      expect(peep.handle).to eq "squareben"
+    end 
+  end   
+
+  describe '#view_all' do
```
Marcellofabbri: should be '.view_all' (class method) rather than '#view_all' (instance method)

lib/chitter.rb
alicelieutier: This line is too long for me. Here's a link to creating multiline strings in ruby: https://commandercoriander.net/blog/2014/11/09/a-multiline-string-cheatsheet-for-ruby/

Something you could do for example is:
```rb
    query = <<-QUERY
INSERT INTO peeps (body, time, name, handle)
VALUES ('#{body}', '#{Time.now.round}', '#{name}', '#{handle}')
RETURNING body, time, name, handle;
   	QUERY
   	result = connection.exec(query)
```

lib/chitter.rb
```diff
@@ -0,0 +1,37 @@
...
+    # p "-------result: #{result}"
+    # Chitter_Peep.new(body: result[0]['body'])
+    Chitter_Peep.new(body: result[0]['body'], time: result[0]['time'], name: result[0]['name'], handle: result[0]['handle'])
```
alicelieutier: This line is also too long.
Would it work if you just did: `Chitter_Peep.new(result[0])`

spec/features/post_message_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(page).to have_content 'squareben'
+  end 
+
+  scenario 'peeps have TIME posted' do 
+    visit('/')
+    fill_in :peep_body, with: 'My first peep'
+    fill_in :peep_name, with: 'Ben'
+    fill_in :peep_handle, with: 'squareben'
+    click_button 'Post'
+    expect(page).to have_content "#{Time.now.round}"
```
alicelieutier: Does this test always pass? Tests that could randomly fail are often worse than test that just fail. You may want to look at freezing time in your test to make them stronger.

PR #1484, opened 2020-01-21
Masha's chitter challenge

marleyhill:


README.md
```diff
@@ -1,5 +1,33 @@
...
+  c) Incorporate timestamp (User story 3) and sort by timestamp.
+The easiest approach will be to sort the array. I will create a feature test for checking posts are displayed in reverse chronological order.
+11. User story 3. In order to keep track of the time when a post was created we need to add a time column to the database and introduce the relevant parameters to the Model, and ensure this is reflected in the View.
+Create and run feature and unit tests to ensure we can see the time.
+12. Creating a date and time columns with automatic date and time setting.
+13. Adding a new parameter :time to the Peeps class and accessing it in Controller and passing to the View. Starting with tests.
+14. User story 4. For this we need to create another table to store user information including id, name, email address, password.
+15. Creating feature test for user sign up. It expects the user to see a welcome message upon registration.
+16. Adding routes and Model class Users functionality.
+
```
iryna-jurkiv: Hi Masha. Thank you for leaving detailed notes of the approach that you took :) I think they will be a great reference point when you do similar projects to this.

app.rb
```diff
@@ -0,0 +1,40 @@
+require 'sinatra/base'
+require_relative './lib/peeps'
+require './lib/users'
+require_relative './database_connection_setup'
+
+class Chitter < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    @user = Users.find(session[:user_id])
```
iryna-jurkiv: Great use of sessions!

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    erb :'index'
+  end
+
+  get '/peeps' do
+    @peeps = Peeps.all.sort_by { |a| -a.id.to_i }
```
iryna-jurkiv: I like that you left your .all method the same and just sorted by using .sort. Well done!

spec/features/displaying_time_spec.rb
```diff
@@ -0,0 +1,7 @@
+feature 'posts are displayed with the time of creation' do
+  scenario 'user can see time next to peep' do
+    sign_in_and_play
+    expect(page).to have_content ("#{Time.new.strftime('%Y-%m-%d')}")
```
iryna-jurkiv: I found it difficult to write a feature test for posts being displayed with the time of creation. I like that you use expect(page).to have_content Time formatted to Years, Months and Date. Good idea!

app.rb
```diff
@@ -0,0 +1,40 @@
...
+    Peeps.create(message: params[:peep])
+    redirect '/peeps'
+  end
+
+  get '/peeps/new' do
+    erb :'/peeps/new_peep'
+  end
+
+  get '/users/new' do
+    Users.create(username: params[:username], email: params[:email], password: params[:password])
```
iryna-jurkiv: I just wanted to check with you why you called Users.create here. My understanding is it can be called in post '/user-add' for example as once the user loads this page (get), they will enter username, email and password which during a post can be used to create a User account for them like you did below. However at get '/users/new' they are loading the page so haven't entered any params for username, email and password.

PR #1493, opened 2020-01-22
Ike chitter challenge 

Ike2017:


spec/spec_helper.rb
```diff
@@ -1,5 +1,24 @@
...
+require 'rspec'
+
+# tell Capybarac about our app class
+Capybara.app = Chitta
+
+RSpec.configure do |config|
+  config.before(:each) do
+    setup_test_database
+  end
+
```
sarita-sumbria: spec_helper.rb is not setup properly to use sinatra framework.

spec/spec_helper.rb
```diff
@@ -1,5 +1,24 @@
...
+# require our Sinatra app file
+require File.join(File.dirname(__FILE__), '..', 'app.rb')
+
+require 'capybara'
+require 'capybara/rspec'
+require 'rspec'
+
+# tell Capybarac about our app class
+Capybara.app = Chitta
```
sarita-sumbria: app.rb class name is  different from the name used in spec_helper.rb to connect with Capybara.

PR #1494, opened 2020-01-22
Sam Chitter Challenge

samidm13:
User stories 1-3 finished


CalMcD:
- nice clean code

- good use of css

- can't see anywhere you could make the code cleaner than it already is. the app file looks as clean and short as it can be (in my opinion)

PR #1495, opened 2020-01-24
commiting challenge

dextor16:


lib/peep.rb
```diff
@@ -0,0 +1,14 @@
+require 'pg'
+
+class Peep
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
lindseydew: Excellent!

app.rb
```diff
@@ -0,0 +1,21 @@
...
+
+class Chitter < Sinatra::Base
+
+  get '/' do
+    @peeps = Peep.all
+    erb :index
+  end
+
+  get '/peeps' do
+    @peeps = Peep.all
```
lindseydew: Great!

app.rb
```diff
@@ -0,0 +1,21 @@
...
+    erb :peeps
+  end
+
+  post '/peeps' do
+   @peeps = params[:peep]
```
lindseydew: What would you do after this?

dextor16:
Would love to have a conversation with you when you have time.

👍

PR #1496, opened 2020-01-24
pull

Reddo12:


app.rb
```diff
@@ -0,0 +1,11 @@
+require 'sinatra/base'
+require './lib/peep'
+
+class Peeps < Sinatra::Base
```
lindseydew: Nice!

db_setup.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/connect_to_db'
+
+if ENV['RACK_ENV'] == 'test'
```
lindseydew: Excellent

lib/peep.rb
```diff
@@ -0,0 +1,15 @@
...
+class Peep
+    def self.all
+      [
+        "peep1",
+        "peep2",
+       ]
+    end
+
+    def self.create
+      DBCconection.query
```
lindseydew: Do you know the sql would write here?

PR #1509, opened 2020-02-17
chitter-challenge

elebecca:


db/01_create_peep_table.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TABLE users(id SERIAL PRIMARY KEY, user_name VARCHAR(30), password VARCHAR(30)); 
+ CREATE TABLE peeps(id SERIAL PRIMARY KEY, user_id INTEGER REFERENCES users(id), peep VARCHAR(140));
```
lookupdaily: can you split this into two files? I think each sql file needs to have one instruction

Gemfile
```diff
@@ -1,6 +1,9 @@
 source 'https://rubygems.org'
 
 group :test do
```
lookupdaily: great you're grouping gems in your gemfile!

lib/peeps.rb
```diff
@@ -0,0 +1,37 @@
...
+     if ENV['ENVIRONMENT'] == 'test'
+       connection = PG.connect(dbname: 'chitter_manager_test')
+     else
+       connection = PG.connect(dbname: 'peeps_manager')
+     end
+    end
+    def self.all
+      connection = Peeps.connect
+
+      result_peeps = connection.exec( "SELECT * FROM peeps INNER JOIN users ON (peeps.user_id = users.id);" )
```
lookupdaily: Great that you've extracted the database connection into a new method! You could go one step further and extract it into a database class.

views/make_a_peep.erb
```diff
@@ -0,0 +1,9 @@
+<h1> Make a new peep! </h1>
+ <br>
+ <form>
```
lookupdaily: if you add your action and post method inside form you can set up route to the next page:
`< form action="/make_a_peep" method ="post" >

views/index.erb
```diff
@@ -0,0 +1 @@
+Welcome to Chitter
```
lookupdaily: can you add links on this page to take you to the other routes? Can simply be:

`<a href="/make_a_peep">Make a peep</a>`
`<a href="/view_peeps">View peeps</a>`

elebecca: Thank you I didn't think about that!

PR #1513, opened 2020-03-15
Chitter_challenge

Kittaru87:


KG700:
I love how your's looks! And I like the diagram you've added to your README. Plus your code looks tidy. Small improvements: using the PG escape method to fix the apostrophe bug when loading to SQL (although I believe you may have already sorted this), and try using an escape clause to make it not possible for a user to sign up with an email address that already exists in the database

PR #1523, opened 2020-03-16
First user story complete

Benjamaker:


dariathompson:
I like how detailed and easy to read your README is! And also how you creating your tables and asking not to be null. Methods and short and easy to understand
```
CREATE TABLE users(user_id SERIAL PRIMARY KEY, username VARCHAR(50) UNIQUE NOT NULL, password VARCHAR(50) UNIQUE NOT NULL);
```

dariathompson:
Suggestions: make travis CI build work (same as I need to do), use your database more and work on further user stories
is Peep self.create would be better just to call on your database connection
```
def self.create(peep:)
    if ENV['ENVIRONMENT'] == 'test'
      connection = PG.connect(dbname: 'chitter_test')
    else
    end
    result = connection.exec("INSERT INTO peeps (peep) VALUES('#{peep}') RETURNING peep_id, peep;")
    Peep.new(peep_id: result[0]['peep_id'], peep: result[0]['peep'])
  end
```
database connection file
```
require 'pg'

if ENV['ENVIRONMENT'] == 'test'
  connection = PG.connect(dbname: 'chitter_test')
else
end
```

PR #1527, opened 2020-03-16
My chitter challenge - Peter

Thatguy560:


db/migrations/01_create_chitter_table.sql
```diff
@@ -0,0 +1,3 @@
+CREATE TABLE chitter ( id SERIAL PRIMARY KEY,  Peeps VARCHAR(140), 
+Time_Stamp TIMESTAMP WITHOUT TIME ZONE NOT NULL 
+DEFAULT (current_timestamp AT TIME ZONE 'UTC'));
```
plkujaw: nice one, was thinking about doing similar thing (getting time directly from database not app but didnt spent much time trying to figure out how to do it :P)

spec/spec_helper.rb
```diff
@@ -1,5 +1,15 @@
 require 'simplecov'
 require 'simplecov-console'
+require 'rspec'
+require 'sinatra'
+require 'rubocop'
+require 'capybara'
+require 'selenium-webdriver'
+require 'capybara/rspec'
```
plkujaw: just realised now that I dont have these in my spec_helper lol so not sure if I was actually testing the whole thing correctly :DDD

Gemfile
```diff
@@ -1,6 +1,11 @@
...
 
 group :test do
+  gem 'rspec', group: :test 
+  gem 'pg'
+  gem 'sinatra'
+  gem 'capybara', group: :test
+  gem 'selenium-webdriver'
   gem 'rspec'
   gem 'rubocop', '0.71.0'
   gem 'simplecov', require: false
```
plkujaw: I think it would be worth trying to tidy up this list, ie. delete duplicate rspec, delete` group: :test` reference for rspec and capybara as they are already incuded in "global" `group :test`

PR #1528, opened 2020-03-16
nowhere near finished

gdharris1000:
register, post and view messages working


artemis-p:
You've done a great job with the challenge! Your code is clean and easy to read and the same applies to your domain model.

PR #1531, opened 2020-04-13
Weekend challenge

FayeCarter:
To do:
Authentication for user log in 
User profile page
Tagging users
Comments


.travis.yml
```diff
@@ -5,6 +5,11 @@ services:
 
 before_script:
   - psql -U postgres -c 'CREATE DATABASE chitter_test;'
-
+  - psql -U postgres -d chitter_test -f db/migrations/01-create_peeps_table.sql
+  - psql -U postgres -d chitter_test -f db/migrations/02-update_with_date.sql
+  - psql -U postgres -d chitter_test -f db/migrations/03-create_users_table.sql
+  - psql -U postgres -d chitter_test -f db/migrations/04-update_peeps_wit_peep_name.sql
+  - psql -U postgres -c 'CREATE DATABASE chitter;'
```
PhilipVigus, line 12: I'm not sure why this line is needed? I thought Travis was only testing, so would only need the test database?

app.rb
```diff
@@ -0,0 +1,66 @@
+require 'sinatra/base'
```
PhilipVigus, line 1: Although this is generally well laid out, I'd suggest you take a look at RESTful design for your routes. The review document for the challenge gives a specific example of how to structure your routes in a RESTful way, and there are tonnes of good resources online. By applying them, you'll end up with a more consistent route naming policy that also consistently utilises the GET/POST/DELETE stuff too

lib/db_connection.rb
```diff
@@ -0,0 +1,16 @@
+require 'pg'
+
+class DbConnection
+
+  def self.setup(dbname)
+    @connection = PG.connect(dbname: dbname)
+  end
+
+  def self.connection
```
PhilipVigus, line 9: Does this need to be public?

lib/peep.rb
```diff
@@ -0,0 +1,42 @@
+require 'pg'
```
PhilipVigus, line 1: I'm afraid I found the naming here a bit confusing. If this is the Peep class, what does the @peep variable represent?

views/chitter/home.erb
```diff
@@ -0,0 +1,32 @@
+<!DOCTYPE html>
```
PhilipVigus, line 1: Generally really liking your layout for your views - well formed, easy to follow and understand.

spec/db_connection_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'db_connection'
```
PhilipVigus, line 1: A very elegant way of testing this class. I like your use of expect....to receive.

spec/features/home_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    expect(page).to have_link('Log in', href: '/login')
+  end
+
+  scenario 'User can get to Login' do
+    visit '/home'
+    click_link 'Log in'
+    expect(page).to have_button "log_in"
+  end
+
+  # scenario 'Home displays peeps' do
```
PhilipVigus, line 16: I know you probably left this in for future reference, but having large amounts of commented-out code is generally seen as a bad thing.

spec/features/log_out_spec.rb
```diff
@@ -0,0 +1,11 @@
+require 'peeper'
```
PhilipVigus, line 1: Is this test really testing the log out feature? I would expect there to be some set up where a user is logged in first, then logged out?

spec/features/login_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'Login Page Loads' do
```
PhilipVigus, line 1: A good feature test

spec/features/signup_spec.rb
```diff
@@ -0,0 +1,13 @@
+feature 'Signup Page Loads' do
```
PhilipVigus, line 1: There's a mismatch between the feature description and what's being tested here I think?

PhilipVigus:
What you've got here is a good start, and seems really solid. I like how you've structured your unit tests - they're really well thought out and clear.

I'd definitely recommend having a look at RESTful design for your routes, as I think that would make their structure a lot more consistent and easy to follow.

PR #1533, opened 2020-04-13
Chitter-challenge Alex

AlexLewis10:


chitter.rb
```diff
@@ -0,0 +1,35 @@
...
+    erb :signup
+  end
+
+  post "/sign_in_details" do
+    session[:handle] = params[:handle]
+    User.create(name: params[:name], handle: params[:handle], email: params[:email], password: params[:password])
+    redirect '/'
+  end
+
+  post "/send_compose_peep" do
```
grahamfalconer, line 23: -> post should be 'get'

PR #1537, opened 2020-04-14
untagged & imperfect

Edward-Phillips:
Missing a bunch of edge cases like invalid email address and the invalid credentials screen is the same for sign up or username/password mismatch when logging in, not very informative.

Issues: 
Manipulating PG query result objects - this tripeed me up at the last step really, I had tags working (not emailing, but displaying who'd been tagged) and then realised I'd neglected to add a name column to the users table and for sign up and somehow adding this in completely broke my tagging, so please do have a look and see if you can let me know where I've gone wrong.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -12,8 +12,14 @@ class Chitter < Sinatra::Base
   end
   post '/users/' do
     p params
-    redirect '/'
+    redirect '/peeps'
   end
+
+  get '/peeps' do
+    @posts = Peep.all
+    erb(:view_peeps)
+  end
+  
   get '/peeps/new' do
     erb(:new_peep)
   end
--- a/app.rb
+++ b/app.rb
@@ -5,4 +5,11 @@ class Chitter < Sinatra::Base
   get '/' do
     erb(:index)
   end
+
+  get '/users/new' do
+    erb(:signup)
+  end
+  post '/users/' do
+    redirect '/'
+  end
 end
--- a/app.rb
+++ b/app.rb
@@ -16,8 +16,9 @@ class Chitter < Sinatra::Base
     who = params['username']
     password = params['password']
     email = params['email address']
+    name = params['name']
     unless User.valid_signup(handle: who, mail: email)
-      User.create(usr: who, pass: password, email: email)
+      User.create(usr: who, pass: password, email: email, name: name)
       User.sign_in(user: who, pass: password)
       redirect '/peeps'
     else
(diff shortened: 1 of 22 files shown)


spec/features/signing_up_spec.rb
```diff
@@ -0,0 +1,15 @@
+feature 'signing up' do
```
Angelica137, line 1: I like that the feature test goes through all the fields expected in the form.

app.rb
```diff
@@ -0,0 +1,64 @@
...
+  post '/sign_in/verify' do
+    User.sign_in(user: params['username'], pass: params['password'])
+    if User.logged_in == false
+      redirect ('/sign_in/invalid')
+    else
+      redirect ('/peeps')
+    end
+  end
+
+  get '/sign_in/invalid' do
```
Angelica137, line 38: I really liked you implemented cases for failure scenarios

lib/user.rb
```diff
@@ -0,0 +1,68 @@
+require 'pg'
+
+class User
+  attr_reader :username, :password, :email_address, :id
```
Steven-Klavins, line 4: Essure you restrict the access of the password as it's very visible in an attribute reader :)

lib/user.rb
```diff
@@ -0,0 +1,68 @@
...
+
+  def self.sign_in(user:, pass:)
+    if self.authenticate(username: user, pw: pass)
+      self.log_in(user)
+    else 
+      self.log_out
+    end
+  end
+
+  def self.valid_signup(handle:, mail:)
```
Angelica137, line 40: nice :)

spec/user_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  context "databasics: " do
+    it " should be able to add a user to the database:" do
+      User.create(usr: 'ed', pass: 'ed_pass', email: 'ed_email')
+      expect(User.all[-1].username).to eq('ed')
+     end
+  end
+
+  describe '.all' do
+    it 'returns a list of users' do
+      User.create(usr: 'Steven', pass: 'Steve_secure', email: 'Steve_email')
```
Steven-Klavins, line 24: That's me! Nice! haha

lib/user.rb
```diff
@@ -0,0 +1,68 @@
...
+    self.connect
+    check = @con.exec("SELECT * FROM users WHERE username = '#{handle}' OR email = '#{mail}'")
+    check.any?
+  end
+
+  def self.authenticate(username:, pw:)
```
Angelica137, line 46: Is there a way to account for user names already taken?

views/invalid_login.erb
```diff
@@ -0,0 +1,10 @@
+<h1>
+Invalid Login details.
```
Angelica137, line 2: Nice thinking about error recovery for the user

PR #1544, opened 2020-05-10
Gareth Woodman - Chitter Challenge

GarethWoodman:


app.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  # Returns list of peeps
+  get '/' do
+    @user = session[:current_user]
+    @peeps = Peep.all
+    erb :'peeps/index'
+  end
+
+  # Form to submit peep
+  get '/peeps/new' do
```
will-head, line 17: Nice use of routing, makes the path clear!

views/peeps/index.erb
```diff
@@ -0,0 +1,41 @@
...
+    <link rel="stylesheet" type="text/css" href="/style.css" />
+  </head>
+  <body>
+
+    <div class="forepage">
+      <h1> Chitter Board </h1>
+
+      <% if @user %>
+        <p>Signed in as <%= @user.name %></p>
+        <a href="http://localhost:9292/peeps/new">Make a peep</a>
```
will-head, line 14: Maybe make these links relative? The app won't always be running on http://localhost:9292/, so "/peeps/new" would be better.

PR #1557, opened 2020-05-11
Catriona's challenge

cmb84scd:
I tried...


README.md
```diff
@@ -14,6 +14,26 @@ As usual please start by forking this repo.
 
 We are going to write a small Twitter clone that will allow the users to post messages to a public stream.
 
+How to set up the database
+-------
+
+1. Connect to psql
```
ZsofiaS, line 20: Clear instructions :)

database_connection_setup.rb
```diff
@@ -0,0 +1,7 @@
+require './lib/database_connection'
```
ZsofiaS, line 1: You require the DatabaseConnection class, but didn't use it inside the script.

spec/database_helpers.rb
```diff
@@ -0,0 +1,7 @@
+require 'pg'
+
+def persisted_data(id:, table:)
+  connection = PG.connect(dbname: 'Chitter_Manager_test')
+  result = connection.query("SELECT * FROM peeps WHERE id = #{id};")
```
ZsofiaS, line 5: You could refer to the 2nd parameter(`table:`) in your SQL query here, instead of hardcoding `peeps` so this helper script can be used for other tables too.

app.rb
```diff
@@ -0,0 +1,31 @@
...
+class Chitter < Sinatra::Base
+  enable :sessions
+
+  get '/' do
+    @user = User.find(session[:user_id])
+    @peeps = Peep.all
+    erb(:index)
+  end
+
+  post '/new-peep' do
```
ZsofiaS, line 15: I think it would be better to use '/peeps/new' route or similar that is RESTful (see code review doc).

lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
+require 'pg'
+
+class DatabaseConnection
```
ZsofiaS, line 3: Nice use of abstraction & refactor

PR #1560, opened 2020-05-11
Chitter Challenge

ellistrickett:


app.rb
```diff
@@ -0,0 +1,46 @@
+require 'sinatra'
+require './lib/peep'
+require './lib/user'
+
```
mbrad26, line 4: I really like that you kept the controllers skinny .

lib/peep.rb
```diff
@@ -0,0 +1,33 @@
...
+    Peep.new(id: peep[0]['id'], body: peep[0]['body'], created_at: peep[0]['created_at'])
+  end
+
+  def self.all
+    peeps = connect_to_database.exec("SELECT * FROM peeps")
+    peeps.map { |peep| Peep.new(id: peep['id'], body: peep['body'], created_at: peep['created_at'])}.reverse
+  end
+
+  private
+
```
mbrad26, line 24: The next step will be to extract the database set up into an object.

lib/user.rb
```diff
@@ -0,0 +1,37 @@
...
+  def initialize(id:, name:, username:, email:, password:)
+    @id = id
+    @name = name
+    @username = username
+    @email = email
+    @password = password
+  end
+
+  def self.sign_up(name:, username:, email:, password:)
+    user = connect_to_database.exec "INSERT INTO users (name, username, email, password) VALUES('#{name}', '#{username}', '#{email}', '#{password}') RETURNING id, name, username, email, password;"
```
mbrad26, line 13: I think that will be best to keep the SQL query inside () so you are consistent with the rest of your code.

PR #1569, opened 2020-06-07
Chitter over n out

nadinedelia:


db/migrations/02_create_user_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE users (id SERIAL PRIMARY KEY, name VARCHAR(10), email VARCHAR(40), username VARCHAR(12), password VARCHAR(15));
```
alicelieutier, line 1: These fields are really short. Names, emails, passwords etc should be able to be long!

PR #1572, opened 2020-06-08
Chitter Challenge

Sumner1185:
Realised part way through Sunday that I had gone about this incorrectly and should have started with a user log in to get info on the user before moving on to creating and showing peeps. I also spent too much time trying to get CSS to work properly. Frustratingly I couldn't get the time stamp to work on Sunday and was not able to get very far with the challenge.


bin/bundle
```diff
@@ -0,0 +1,114 @@
+#!/usr/bin/env ruby
```
alicelieutier, line 1: Should this file be committed?

bin/htmldiff
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
alicelieutier, line 1: Same here.

bin/tilt
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
alicelieutier, line 1: It feels like you have plenty of files in bin, that should not be part of this project.

db/01_create_peep_manager_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE "peep_manager"(id SERIAL PRIMARY KEY, title TEXT, peep VARCHAR(200))
```
alicelieutier, line 1: This query is missing a semicolon :(

lib/peep.rb
```diff
@@ -0,0 +1,37 @@
...
+
+  def self.all
+    if ENV['ENVIRONMENT'] = 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    result = connection.exec("SELECT * FROM peep_manager")
+    result.map do |peep|
+      Peep.new(id: peep['id'], title: peep['title'], peep: peep['peep'], time:['time'])
```
alicelieutier, line 22: You may want to use the 'splat' operator here. Here is an example use:
```rb
def go(x:, y:)
end

point = { x: 100, y: 200 }
go(**point)
```

spec/feature/peeps_options_spec.rb
```diff
@@ -0,0 +1,21 @@
+feature 'peeps' do
+  
+  scenario "it should show a list of peeps" do
+    Peep.create(title: "Test", peep: "this is a test peep")
+    Peep.create(title: "Hello", peep: "feeling ok")
+    Peep.create(title: "Bye", peep: "feeling sad")
```
alicelieutier, line 6: I would not create things directly in code within a feature test. The whole test should be from the point of view of the user!

lib/peep.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+  end
+
+  def self.create(title:, peep:)
+
+    if ENV['ENVIRONMENT'] = 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
alicelieutier, line 32: You could DRY this code out.

app.rb
```diff
@@ -0,0 +1,28 @@
...
+    @peeps = Peep.all
+    p @peeps
+    erb :"peeps/index"
+  end
+
+  get '/peeps/new' do
+    erb :"peeps/new"
+  end
+
+  post '/peeps' do
```
alicelieutier, line 21: This route should maybe be `/peeps/new` or `/peeps/create`

lib/peep.rb
alicelieutier: Also... this may actually solve your "time" bug :) You have a typo at the end of this line.

PR #1573, opened 2020-06-08
Katie Chitter challenge

Katie-McDonagh:
I really struggled with the challenge this weekend, in particular with how to connect to the data base, but also with deciding my routes, any feedback would be appreciated,
Thanks!


spec/features/display_peeps_spec.rb
```diff
@@ -0,0 +1,12 @@
+require 'pg'
+
+feature 'homepage displays all the peeps' do
+  scenario 'a user is displayed a page with all the peeps' do
+    connection = PG.connect(dbname: 'chitter_test')
+
```
ChrisCooney05, line 6: Since you have an add-peep route, it may be worth using the feature test to fully test the process of adding the peep then going back to the homepage to view rather than using SQL to fill the database

lib/peep.rb
```diff
@@ -0,0 +1,34 @@
...
+  end
+
+  def self.all
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
```
ChrisCooney05, line 19: This is a good start for the next steps.  you could look at mapping the results to new instances of the Peep class rather than just returning an array of peeps

lib/peep.rb
```diff
@@ -0,0 +1,34 @@
+require 'pg'
+
```
ChrisCooney05, line 2: overall the code and test quality looks great, really easy to read and understand exactly what is going on, I can follow the code and work out in my head what a user would interact with and how the info is stored and recalled in the database, really good!

app.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  get '/' do
+    @peeps = Peep.all
+    erb(:view_peeps)
+  end
+
+  get '/new' do
+    erb(:add_peep)
+  end
+
```
ChrisCooney05, line 16: Like that you directly use the params to add info to your .add method rather than save the info into variables and then pass that in!

spec/features/display_peeps_spec.rb
ChrisCooney05: also, the only reason I think of for this test failing is that title will have a nil value, so may be worth adding a title to the inserted data and see how that goes.

PR #1575, opened 2020-06-08
Rich Chitter Challenge

RichEwin:
I have completed user stories 1, 2 and 4.

**Incomplete**
Due to time constraints I have been unable to implement the time of peep feature, however I have added a timestamp column to my database that records the data correctly. I just need to write the code.

**Improvements** 

I would still like to implement my time feature, therefore have left my controller untidy with a view to slim down when this feature is implemented.

I would like to progress from a basic HTML web page to a more modern layout. I will spend some time this week learning how I can make the necessary changes with css.

Thank you for reviewing my code, feel free to clone my repo and try my site out. I am looking forward to hearing your feedback and how I can improve.


app.rb
```diff
@@ -0,0 +1,23 @@
...
+
+class Chitter < Sinatra::Base
+
+  get '/' do
+    erb :index
+  end
+
+  post '/sign_up' do
+    $username = params[:username]
+    $password = params[:password]
```
RaeRachael, line 12: do these need to be global variables?

spec/setup_test_database.rb
```diff
@@ -0,0 +1,8 @@
+require 'pg'
+
+p "Running Test Database"
```
RaeRachael, line 3: Do we need to see this on every Rspec run?

spec/spec_helper.rb
```diff
@@ -1,10 +1,29 @@
+require_relative './setup_test_database'
+
+ENV['ENVIRONMENT'] = 'test'
+# ENV['RACK_ENV'] = 'test'
```
RaeRachael, line 4: should be saved in a git commit, so no need to save it here in a comment

app.rb
```diff
@@ -0,0 +1,23 @@
+require 'sinatra/base'
+require './lib/peep'
+
+class Chitter < Sinatra::Base
+
+  get '/' do
+    erb :index
```
RaeRachael, line 7: maybe index could have been renamed?

views/index.erb
```diff
@@ -0,0 +1,30 @@
...
+    <h4> Create your account </h4>
+    <div>
+      <form action="/sign_up" method="post">
+
+        <input type='text' placeholder='Username' name='username'>
+
+        <input type='password' placeholder='Password' name='password'>
+
+        <input type="submit" value="Sign up">
+      </div>
```
RaeRachael, line 27: should this appear on every homepage visit?

app.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  post '/sign_up' do
+    $username = params[:username]
+    $password = params[:password]
+    erb :sign_up
+  end
+
+  post '/peeps' do
+    Peep.create(peep: params[:peep])
+    @peeps = Peep.all.reverse
```
RaeRachael, line 18: is .reverse good enough to ensure reverse chronological order, what if peep database table got unsorted?

views/sign_up.erb
```diff
@@ -0,0 +1,23 @@
...
+   }
+  </style>
+</head>
+<body>
+
+<h1> Welcome to Chitter <%= $username %>! </h1>
+<h2>Thanks for signing up</h2>
+<form action="/peeps" method='post'>
+  <input type="submit" value="Start peepin!">
+</form>
```
RaeRachael, line 21: i see why this is a post request, but wouldn't making it a get request fit better, maybe the process to add a new peep could be a post '/new_peep' redirect to get '/peeps'?

PR #1578, opened 2020-06-08
Chitter test_SB_200608

sofyloafy:


lib/peep.rb
```diff
@@ -0,0 +1,35 @@
...
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+    result = connection.exec("INSERT INTO peeps(peep, time_stamp) VALUES ('#{peep}','#{time_stamp}') RETURNING peep, time_stamp;")
+    Peep.new(peep: result[0]['peep'], time_stamp: result[0]['time_stamp'])
+  end
+
+  def self.all
+    if ENV["ENVIRONMENT"] == "test"
```
tristanlangford, line 23: Look into moving the 'choose database' to a separate file which only has the if/else statement and then require that in the app.rb. https://github.com/makersacademy/course/blob/master/bookmark_manager/walkthroughs/14.md (about half way down)

lib/user.rb
```diff
@@ -0,0 +1,40 @@
+require 'bcrypt'
+# require_relative 'database_connection'
+
+class User
+
+  def self.create(email:, password:)
+    encrypted_password = BCrypt::Password.create(password)
```
tristanlangford, line 7: Nice use of BCrypt!

app.rb
```diff
@@ -0,0 +1,39 @@
...
+
+class Chitter < Sinatra::Base
+  enable :sessions
+
+  get '/' do 
+    erb :index
+  end
+
+  get '/create_peep' do
+    @user = User.find(session[:user_id])
```
tristanlangford, line 13: Good use of sessions to store the user_id

PR #1582, opened 2020-07-05
Pull request for chitter challenge

jamesAforster:


samlandman:
What I love:
Clean interface, and great functionality on the actual website
Clean code, very logical and easy to follow
The fact that the time stamp is displayed on each Peep
A clear readme

What could improve:
- DRY principle on database_manager.rb file - lines 6-9 & 27-30 are identical and could be a private method (eg. database_selector)

Something to be aware of
- The name of one of the classes is DatabaseManager. The name could get confusing when there are multiple databases. Just something to keep in mind :)

PR #1585, opened 2020-07-06
WIP pull request for chitter

indisaurusrex:
I have really only just begun, the current stage is attaching the database to the code so I can add peeps to them. 

Test is in place for creating a new peep, some code has been written to receive peep from the user, nothing has been done with that peep info yet.


lib/peep.rb
```diff
@@ -0,0 +1,9 @@
+class Peep
+  def self.all
+  end
+  
+  def self.create(body:, user:)
+    connection = PG.connect(dbname: 'chitter')
+    connection.exec("INSERT INTO peeps (body, user) VALUES('#{body}', '#{user}')")
+  end
```
stupot1, line 8: Could you extract the database functionality into a separate class

lib/peep.rb
```diff
@@ -0,0 +1,9 @@
+class Peep
+  def self.all
+  end
+  
+  def self.create(body:, user:)
```
stupot1, line 5: Nice use of symbols in the method definition

db/create_tables.sql
```diff
@@ -0,0 +1,15 @@
+CREATE TABLE peeps
```
stupot1, line 1: Useful to include the table setup SQL in the project.  These can be executed automatically by the travis.yml file for testing or a Rakefile (not sure how though...)

views/new_peep.erb
```diff
@@ -0,0 +1,5 @@
+<form action='/posted' method='post'>
+<label for='body'>peep: </label>
+<input type='text' id='body' name='body'/>
+<input type='submit' value='chirp'/>
```
stupot1, line 4: I like that the button is called chirp :)

spec/chitter_db_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'peep'
+
+describe Peep do
+  describe '.all' do
+    it 'returns all peeps' do
+      connection = PG.connect(dbname: 'chitter_test')
+
+      peep = Peep.create(body: 'My first peep!', user: '@indisaurusrex')
```
stupot1, line 8: You dont need the "peep =" as you do not do anything with the variable peep after it is assigned and do not return anything from the create method.

PR #1586, opened 2020-07-06
Chitter Weekend Challenge - Josh

JoshuaNg2332:
**Completed User Story 1 & 2 and started 3**

Pretty happy with how the weekend challenge went. Struggled when trying to implement the testing database but got it working in the end.
Found that I wrote a lot of unneeded tests so omitted them towards the end.
Something I need to do better is commit changes more frequently but found that I had written a test that made me have to build a majority of the app's posting functionality in one main test. Thus this is the reason for not as frequent pushing.


README.md
```diff
@@ -7,6 +7,71 @@ Chitter Challenge
...
+- Database to store new peep and time created
+- Text form to enter text
+- Submit button
+
+Input:      Output:
+Peep        New page shows 
+Submit
+
+Progress So Far:
+-------
```
Dhara-95, line 43: Really like the detailed outline of what has been completed so far.

README.md
```diff
@@ -7,6 +7,71 @@ Chitter Challenge
...
+4. Here you want to create 2 new tables called, "chitter" and "chitter-test" using "CREATE DATABASE (name)" and "CREATE TABLE peep".
+5. Then run 'rackup' in your terminal.
+6. On your web browser, go to localhost:9292
+
+### How to run tests
+
+1. Fork/Clone this repository.
+2. Run bundle install in your terminal
+3. Then run rspec to see tests.
+
```
Dhara-95, line 73: Good README.md overall especially on the installation and testing instructions. As discussed I know you would have also liked to add a domain model if you had more time, which would be a great visual representation of your planning process.

app.rb
```diff
@@ -0,0 +1,26 @@
...
+  post '/add_post' do
+    Chitter.create(message: params[:peep_txt])
+    redirect to('/live')
+  end
+
+  get '/live' do
+    @chitter = Chitter.all
+    erb :live
+  end
+
```
Dhara-95, line 24: Good 'clean' code in the controller file, which seems to adhere to SRP well with each get/post route having one responsibility.

lib/chitter.rb
```diff
@@ -0,0 +1,37 @@
...
+class Chitter
+
+  attr_reader :id, :message
+
+  def initialize(id:, message:, posting_date:)
+    @id  = id
+    @message = message
+    @posting_date = posting_date
+  end
+
```
Dhara-95, line 12: Good refactoring to use instance variables here.

lib/chitter.rb
```diff
@@ -0,0 +1,37 @@
+require 'pg'
+
+class Chitter
+
```
Dhara-95, line 4: Although it is clear what this class is doing, could consider naming this the 'Peep' class instead as this model file specifically adds and creates new 'peeps'. Then for other user stories could have a 'User' class, which is responsible for signing in users.

migrations/01_create_peep_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peep(id SERIAL PRIMARY KEY, message VARCHAR(140), posting_date DATE NOT NULL DEFAULT CURRENT_TIMESTAMP);
```
Dhara-95, line 1: Not entirely necessary but could have another file here for the command to create a table in the test database.

PR #1588, opened 2020-07-06
First Pull Attempt

samlandman:
Some functionality still to do.


jamesAforster:
- Super impressive how you have been able to implement the user sign up function as well as the entire CRUD gamut over the weekend - big kudos!
- No readme which would have been good to be included as well but considering everything else you've done it's understandable.
- Some of the functions in your tests could be put in a web_helper.rb file so that you can call a 'User signs in and adds peep' method in multiple tests and keep things nice and DRY.
- Really like how you've got the error message to display for the log in/password.

PR #1598, opened 2020-08-03
Chitter Challenge

phump81:


spec/message_spec.rb
```diff
@@ -0,0 +1,28 @@
+require 'message'
+
+describe Message do
+
+  describe '.all' do
+    it 'shows a list of messages' do
+      connection = PG.connect(dbname: 'chitter_test')
+      message = Message.create(message: "The President is an idiot")
+      Message.create(message: "The PM is a bellend")
```
harrylb14, line 9: Rate this, I would use this site

spec/setup_test_database.rb
```diff
@@ -0,0 +1,6 @@
+require 'pg'
+
+def setup_test_database
+  connection = PG.connect(dbname: 'chitter_test')
```
harrylb14, line 4: this could utilise your DatabaseConnection class rather than manually connecting to PG again

lib/user.rb
```diff
@@ -0,0 +1,19 @@
...
+class User
+
+  attr_reader :user_id, :name, :username, :email, :password
+
+  def initialize(user_id:, name:, username:, email:, password:)
+    @user_id = user_id
+    @name = name
+    @username = username
+    @email = email
+    @password = password
```
harrylb14, line 12: perhaps don't want password so easily accessible!

views/sign_up.erb
```diff
@@ -0,0 +1,11 @@
+<form action="/sign_up" method="post">
+  <input type="text" name="name" placeholder="Name" />
+  <br>
```
harrylb14, line 3: use of <br> definitely makes it look much nicer!

spec/features/posts_new_message_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'posting new messages' do
```
harrylb14, line 1: perhaps could do with a few more feature tests throughout? perhaps may help for edge cases etc

spec/setup_test_database.rb
```diff
@@ -0,0 +1,6 @@
...
+  connection.exec("TRUNCATE messages;")
```
harrylb14, line 5: useful tip i found last week, if you do "TRUNCATE messages, RESTART IDENTITY;" it resets the id counter as well. as currently everytime you reset it, while it wipes the table it still keeps counting from where you left off, so after a few rspecs youll be on silly numbers in the ID column. restart identity resets it to 1 each time. probably not essential, but quite nice I thought!

app.rb
```diff
@@ -0,0 +1,35 @@
+require 'sinatra/base'
```
harrylb14, line 1: v skinny controller, we love to see it

PR #1599, opened 2020-08-03
Ralph Mallett - chitter challenge

ralphm10:
### Partial solution

- The four user stores in STRAIGHT UP above have been completed, however a username is not displayed with a peep
- At present when signing up, only the username is stored and this is in a parameter
- The next stage I plan to implement is to create a USER model and table which would store the users name, email and password, and then to implement functionality for users to login and logout


ROCHAAL:
The partial solution of the project is clear and the next stage that will be implemented "create a user model" will complete this stage of the project.

PR #1600, opened 2020-08-03
Chitter-Challenge Week 4

Pi-hils:


app.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  get '/peep' do
+    erb :index
+  end
+
+  get '/peep/new' do
+    @peep = Peep.all
+    erb :peep
+  end
```
SilverLongjohns, line 16: Clean controller. Could maybe have it even cleaner by having the root show your index view!

views/index.erb
```diff
@@ -0,0 +1,13 @@
+<h1>Welcome to Chitter</h1>
+
```
SilverLongjohns, line 2: Could have a segment here that shows your list of peeps?

spec/features/view_peeps_spec.rb
```diff
@@ -0,0 +1,8 @@
+require 'spec_helper'
```
SilverLongjohns, line 1: I don't think 'spec_helper' needs to be required.

app.rb
```diff
@@ -0,0 +1,20 @@
...
+  
+  get '/' do
+    "Welcome to Chitter"
+  end
+
+  get '/peep' do
+    erb :index
+  end
+
+  get '/peep/new' do
```
SilverLongjohns, line 13: There probably needs to be a post method here somewhere

PR #1602, opened 2020-08-03
makers week 4 weekend challenge

melanieanderson1995:


spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,15 @@
+feature 'Viewing peeps' do
+  scenario 'A user can see peeps in reverse chronological order' do
```
rjkviegas, line 2: I do not think the order of the peeps is being tested here, just that they are on the page.

lib/peep.rb
```diff
@@ -0,0 +1,24 @@
+require 'pg'
+
+class Peep
```
rjkviegas, line 3: Refactoring the Peep class would be desirable as it has too many responsibilities. Perhaps a DatabaseConnection class.

db/migrations/01_create_chitters_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE chitters(id SERIAL PRIMARY KEY, url VARCHAR(60));
```
rjkviegas, line 1: This table is not referenced in the rest of the code, a table called 'peeps' with a column 'message' is though.

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+    connection = PG.connect(dbname: 'chitter_test')
+
+    connection.exec("INSERT INTO peeps VALUES(1, 'Good morning');")
+    connection.exec("INSERT INTO peeps VALUES(2, 'Good afternoon');")
+    connection.exec("INSERT INTO peeps VALUES(3, 'Good night');")
+
+    visit('/peeps')
+
+    expect(page).to have_content "Good night"
```
rjkviegas, line 11: I think if you add a 'class' to the list items:
```<li class='peep'><%= peep.text %></li>```
you can check specific list items using:
```expect(first('.list_item_class')).to have_content 'content text'```

File: views/index.erb [hunk 5]
```diff
@@ -0,0 +1,5 @@
+<h1> Welcome to Chitter </h1>
+
+<form action = "/peeps">
+  <input type="submit" value="View peeps">
```
rjkviegas, line 4: I like this feature but I don't see a test, was this TDD'd? Also does this work with the form method being stated?

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,15 @@
+feature 'Viewing peeps' do
+  scenario 'A user can see peeps in reverse chronological order' do
+    connection = PG.connect(dbname: 'chitter_test')
+
+    connection.exec("INSERT INTO peeps VALUES(1, 'Good morning');")
```
rjkviegas, line 5: If the 1 is the peep/message_id i do not think you need to insert it if the column is a SERIAL PRIMARY KEY, it will do that automatically for you.

spec/setup_test_database.rb
```diff
@@ -0,0 +1,5 @@
+require 'pg'
+
+connection = PG.connect(dbname: 'chitter_test')
+
+connection.exec("TRUNCATE peeps;")
```
rjkviegas, line 5: Just incase you weren't aware or ran into any issues later on if you use:
```TRUNCATE peeps RESTART IDENTITY```
It will restart the peep_id values

File: spec/peep_spec.rb [hunk 8]
```diff
@@ -0,0 +1,26 @@
+require './lib/peep.rb'
+
+describe Peep do
+  describe '.all' do
+    it 'returns all peeps' do
```
rjkviegas, line 5: Good test of the method.

PR #1608, opened 2020-08-31
Chitter challenge

Charlotte-91:
- 100% test coverage
- Did not add user log in functionality
- Added some CSS


imogen-k:
Website runs great and the code & tests and very clear, only thing you could have done to refactor further is extract the database connection into a separate class

PR #1610, opened 2020-09-01
Chitter-Challenge

Josh-Steadman:
was unable to make user username and password unique


lib/peep.rb
```diff
@@ -0,0 +1,30 @@
...
+        Peep.new(id: peeps[0]['id'], peep: peeps[0]['peep'], time: peeps[0]['time'], username: peeps[0]['username'], name: peeps[0]['name'])
+     end
+
+     def self.get_peeps
+       peeps = DatabaseConnection.query("SELECT * FROM peeps;")
+       all_peeps = peeps.map do |row|
+          Peep.new(id: row['id'], peep: row['peep'], time: row['time'], username: row['username'], name: row['name']) 
+          end
+       
+        all_peeps.sort_by(&:time).reverse
```
ZishJawaid, line 26: Nice use of the sort_by method

spec/features/posting_message_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature "posts a message" do
```
ZishJawaid, line 1: For the feature testing, one suggestion I have is to create the data you want to test (i.e. Peep.new(...)) and then check to see if the data is present.

PR #1612, opened 2020-09-01
Chitter Challenge

yu2491:


views/sign_up.erb
```diff
@@ -0,0 +1,50 @@
...
+
+<form action='/new_user' method="post">
+  <label for="name">Name:</label>
+  <input type="text" name="name"><br><br>
+  <label for="user_handle">User Handle:</label>
+  <input type="text" name="handle" placeholder="@"><br><br>
+  <label for="email">Email Address:</label>
+  <input type="text" name="email"><br><br>
+  <label for="password">Password:</label>
+  <input type="text" name="password"><br><br>
```
fraserbrookhouse, line 48: It would be best if you had the password as type="password" so that the password is not visible as you inputting it.

views/sign_up.erb
```diff
@@ -0,0 +1,50 @@
+<head>
+  <link href="https://fonts.googleapis.com/css2?family=Quicksand&display=swap" rel="stylesheet">
+</head>
+
+<style>
```
fraserbrookhouse, line 5: I like your page styling, it is quite approachable.

spec/user_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'user'
+require 'database_helpers'
+
+describe User do
+
+  describe '.create' do
+    it 'creates a new user' do
+      user = User.create(name: 'Test Person', handle: '@mrtest', email: 'test@example.com', password: 'password123')
```
fraserbrookhouse, line 8: I like that your users have names and handles so that you aren't displaying their email address publicly on the peep feed.

lib/peep.rb
```diff
@@ -0,0 +1,27 @@
+require 'pg'
+require 'database_connection'
```
fraserbrookhouse, line 2: This needs to be require_relative or it won't be able to load the file.

lib/user.rb
```diff
@@ -0,0 +1,31 @@
+require 'bcrypt'
+require 'database_connection'
```
fraserbrookhouse, line 2: This needs to be require_relative or it won't be able to load the file.

spec/database_connection_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+
+      DatabaseConnection.setup('chitter_manager_test')
+    end
+  end
+
+  describe '.query' do
+    it 'executes a query via PG' do
+      connection = DatabaseConnection.setup('chitter_manager_test')
+
+      expect(connection).to receive(:exec).with("SELECT * FROM bookmarks;")
```
fraserbrookhouse, line 16: Need to update to refer to chitter, rather than bookmarks.

PR #1615, opened 2020-09-25
Shore - a better twitter experience

TheDancingClown:


LinusMjorn:
Doubles would have been welcomed in the testing. 
Could have had a few more tests for edge cases.
Rspec configure not done properly.
Tests could have been a bit more DRY.

PR #1623, opened 2020-09-27
Changes in dir path

cfujiname:
I had to change the directory path. All commits are gone, remaining just two: one for the 'dir replacement' and the other one after checking everything before pull request. Please read notes on how I solved this challenge.


spec/setup_test_database.rb
```diff
@@ -0,0 +1,8 @@
+require 'pg'
+
+def setup_test_database
+connection = PG.connect(dbname: 'chit_manager_test')
+
+# this clears the chits table
+connection.exec("TRUNCATE chits;")
+end
```
CRTW26, line 8: Nice that you ensure empty tables before testing

PR #1625, opened 2020-09-28
Chitter challenge

CRTW26:


app.rb
```diff
@@ -0,0 +1,55 @@
+require 'sinatra/base'
+require 'sinatra/flash'
+require './lib/peeps.rb'
+require './lib/user.rb'
+
+class Chitter < Sinatra::Base
+
+  enable :sessions 
+  register Sinatra::Flash
```
cfujiname, line 9: Interesting use of flash! never occurred to me heheh!

app.rb
```diff
@@ -0,0 +1,55 @@
...
+    session[:id] = user.id 
+    redirect '/'
+  end 
+
+  get '/sessions' do 
+    erb(:sessions)
+  end 
+
+  post '/sessions' do
+    user = User.authenticate(email: params[:email], password: params[:password])
```
cfujiname, line 38: Nice authentication method - I have no experience doing this, even though I tried to insert something similar on mine, I didn't manage it properly :P

lib/user.rb
```diff
@@ -0,0 +1,45 @@
+require 'pg'
+require 'bcrypt'
+
+class User
```
cfujiname, line 4: Nice User Class to separate the sign up methods and encrypting

spec/features/sign_in_spec.rb
```diff
@@ -0,0 +1,40 @@
+feature 'sign in' do
```
cfujiname, line 1: Complete feature tests for the sign in feature - thumbs up!

views/sign_up.erb
```diff
@@ -0,0 +1,9 @@
+<form action='/sign_up' method='post'>
```
cfujiname, line 1: I haven't used much <div> yet, but I was wondering if there is no need to create a Value for submit, to be able to click on the button?

spec/units/peeps_spec.rb
```diff
@@ -0,0 +1,22 @@
...
+require 'pg'
+
+RSpec.describe Peeps do 
+
+  describe ".all" do 
+    it 'returns all peeps' do
+      Peeps.post('first peep')
+      Peeps.post('second peep')
+      peeps = Peeps.all
+      expect(peeps).to be_instance_of Array
```
cfujiname, line 11: Nice! Sort of a good refactoring here with be_instance_of Array

lib/peeps.rb
```diff
@@ -0,0 +1,36 @@
...
+  def self.all
+    set_environment
+    result = @connection.exec("SELECT * FROM peeps;")
+    peeps = result.map { |peep| Peeps.new(peep: peep['peep'], date: peep['date']) }
+    peeps.reverse
+  end
+
+  def self.post(peep)
+    set_environment
+    now = Time.new
```
cfujiname, line 16: Using inbuilt Class Time - good! I embedded my 'time' in the database, different approach!

PR #1634, opened 2020-10-25
chitter finished STRAIGHT UP + HARDER and peep edit/delete functionality

kiriarf:


app/app.rb
```diff
@@ -0,0 +1,81 @@
+require 'sinatra/base'
```
tatsiana-makers, line 1: WIth regards to your question about whether your routes are RESTful, I've commented on the specific routes I think could be changed to make them more restful.

For non-session routes, I wonder if you've seen [this](https://github.com/sjmog/rest)? It's a document and game you can use to check your REST understanding. [This](https://gist.github.com/alexpchin/09939db6f81d654af06b) is also a code snippet where someone summarises their restful routes.
For the _session_ routes, I found [this guide](https://stackoverflow.com/a/7252311)

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+class Chitter < Sinatra::Base
+  enable :sessions
+  enable :method_override
+
+  get '/' do 
+    @sorted_peeps = Peep.all.sort_by { |peep| peep.posted_on }.reverse
+    erb(:index)
+  end
+
+  get '/registrations/new' do
```
tatsiana-makers, line 15: I think this should be `get /users/new`

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    erb(:'registrations/new')
+  end
+
+  post '/registrations' do
```
tatsiana-makers, line 19: Then this could be `post /users`

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+      session[:signup_fail] = true
+      redirect('/registrations/new')
+    end
+  end
+
+  get '/login' do
+    erb(:'sessions/new')
+  end
+
+  post '/sessions' do
```
tatsiana-makers, line 35: [This stack-overflow answer](https://stackoverflow.com/a/7252311) suggests the session routes should be:
- `post /session` 
- `get /session/new`
- `delete /session`



(since each end-user only ever have one session)

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+      redirect('/login')
+    end
+  end
+
+  delete '/sessions' do
+    session.delete(:user_id)
+    redirect('/')
+  end
+ 
+  get '/peeps' do
```
tatsiana-makers, line 52: Based on the gist I posted above, should this be `get /peeps/new`?

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    erb(:'peeps/add')
+  end
+
+  post '/peeps' do
+    Peep.create(username: params[:username], body: params[:body])
+    redirect('/')
+  end
+
+  delete '/peeps' do
```
tatsiana-makers, line 61: We're deleting a specific peep, so should this be `delete /peeps/:id`?

app/app.rb
```diff
@@ -0,0 +1,81 @@
...
+    Peep.delete(id: params[:peep_id])
+    redirect('/')
+  end
+
+  get '/peeps/:id/edit' do
+    @peep = Peep.find_by(id: params[:peep_id])
+    erb(:'peeps/edit')
+  end
+
+  put '/peeps' do
```
tatsiana-makers, line 71: Similarly here, we're updating a specific peep, so should it be `put /peeps/:id`?

app/models/user.rb
```diff
@@ -0,0 +1,101 @@
...
+
+  def initialize(id:, username:, email:, name:, password:)
+    @id = id
+    @username = username
+    @email = email
+    @name = name 
+    @password = password
+  end
+
+  # def self.find_by(id:)
```
tatsiana-makers, line 15: Very minor, but for clean-up, you could remove this comment (:

tatsiana-makers (commented):
This is a nice go at the Chitter challenge! I've posted a few resources I find useful about RESTful routes and suggested some changes to the routes. It might be worth updating the view names to match the routes as well.

PR #1636, opened 2020-10-25
Chitter submission

dan-holmes:


app/models/user.rb
```diff
@@ -0,0 +1,19 @@
+class User < ActiveRecord::Base
+  include BCrypt
+
+  def self.register(username:, full_name:, email:, password:)
+    encrypted_password = Password.create(password)
+    user = self.create(username: username, full_name: full_name, email: email, password: encrypted_password)
+  end
+
+  def self.authenticate(username:, password:)
```
tatsiana-makers, line 9: Can you use `p` to get visibility on what is happening when a user tries to log in with the correct password and then checking the server logs.

For example, you might consider printing out:
- username
- password
- user
- user.password


With these printed out, can you see why the login feature isn't working?

README.md
```diff
@@ -1,132 +1,38 @@
...
+This code currently has an issue with logins. All unit and feature tests pass, but when trying it yourself, it will not let you log in.
 
-We are going to write a small Twitter clone that will allow the users to post messages to a public stream.
+To use the site currently, register and do not log out!
 
-Features:
--------
+## Installation
 
+```shell
```
tatsiana-makers, line 13: Nice job including the setup commands! One minor amendment to add: I had to change `dan` in `database.yml` to my username (:

tatsiana-makers (commented):
This is a nice go at the challenge! I've left a comment hinting at where the bug with login might be coming from. Let me know if you have any questions 😄

app/models/user.rb
dan-holmes: Yes!! Thank you I've got it.

I should have spotted this but was so confused by my positive tests. Once I solved the issue I figured out that I was accidentally creating a user with an unhashed password in one of my support files which was giving a false positive

README.md
dan-holmes: Yes good call! For the next challenge I'm going to look into using Rake files instead.

PR #1637, opened 2020-10-25
Completed first four 'straight-up' and the two 'harder' user-stories

emilyalice2708:


spec/spec_helper.rb
```diff
@@ -8,6 +8,26 @@
 ])
 SimpleCov.start
 
+require_relative './reset_test_database.rb'
```
CamLogie, line 11: in the review doc they say to ensure asset routes are set correctly. I'm sure you've done this elsewhere, but I haven't been able to find it. 

https://github.com/emilyalice2708/chitter-challenge/blob/master/docs/review.md#ensure-asset-routes-are-set-correctly

spec/features/web_helpers.rb
```diff
@@ -0,0 +1,13 @@
...
+  visit('/users/new')
+  fill_in :name, with: "John Smith"
+  fill_in :username, with: "JSmith"
+  fill_in :email, with: "testing@testmail.com"
+  fill_in :password, with: "Rosebud"
+  click_button "Sign Up"
+end
+
+def create_karen
+  User.create(name: "Karen", username: "shopaholic", email: "karen@entitled.com", password: "theworst")
```
CamLogie, line 11: this is class hahaha

spec/features/authentication_spec.rb
```diff
@@ -0,0 +1,48 @@
+feature 'authentication' do
```
CamLogie, line 1: consider possibly splitting this into sign in/ sign out feature files instead of one large one. 

https://github.com/emilyalice2708/chitter-challenge/blob/master/docs/review.md#avoid-rspec-feature-scenarios-organized-like-unit-tests

app.rb
```diff
@@ -0,0 +1,67 @@
+require 'sinatra/base'
```
CamLogie, line 1: consider splitting this into different controller files to keep each one lighter 

https://github.com/emilyalice2708/chitter-challenge/blob/master/docs/review.md#split-routes-into-separate-controller-files

CamLogie (commented):
I mean, this was really good, I loved the app interface and styling. When I was going through the app itself everything ran very smoothly. Some slight stylistic changes based off of the review, but as mentioned probably best to get a coach to review as well

lib/peep.rb
```diff
@@ -0,0 +1,59 @@
...
+    user = User.find(@user_id)
+    user.username
+  end
+
+  def user_name
+    return "Anonymous" if @user_id.nil?
+
+    user = User.find(@user_id)
+    user.name
+  end
```
alicelieutier, line 58: This is ok, but inefficient, as it means you are doing several queries to the database to fetch the handle and name. There are several solutions here. If you believe that you will need both these pieces of information everytime you fetch the peeps, you could add them to the Peep.all query:

```
SELECT peeps.*, users.name, users.handle FROM peeps
INNER JOIN users ON peeps.user_id=users.id;
```

Or you could have them both in one query, something like:
```rb
def user
    return anonymous_user if @user_id.nil?
    User.find(@user_id)
end
```

alicelieutier (commented):
Good attempt! Your code is complete, well separated and easy to read.
Your routes are not following strict REST, so I've added a few comments.
Overall this is good.

PR #1648, opened 2020-10-26
Chitter Challenge!

CamLogie:


README.md
```diff
@@ -1,132 +1,37 @@
...
-As a Maker
-So that I can let people know what I am doing  
-I want to post a message (peep) to chitter
+- Step 2: ```CREATE DATABASE chitter_test;```
+- Step 3: ``` \c chitter_test;```
 
-As a maker
-So that I can see what others are saying  
-I want to see all peeps in reverse chronological order
+**Adding time to the databases**
```
kacesera, line 22: Can you refactor the above to include time in one step, rather than running two queries? I haven't looked further along in your code yet, but could you implement a 'timestamp' in SQL, so that the table will automatically give it a time?

README.md
```diff
@@ -1,132 +1,37 @@
 Chitter Challenge
```
kacesera, line 1: I would love to know more about your process in creating this. What did you learn, where did you get stuck, and how did you plan?

chitter.rb
```diff
@@ -0,0 +1,39 @@
...
+      con = PG.connect :dbname => 'chitter'
+    end
+    retrieve_all = con.exec "SELECT * FROM peeps"
+    retrieve_all.map do |column|
+      Chitter.new(message: column['message'], id: column['id'], time: column['time'])
+    end
+
+  end
+
+  def self.add(message)
```
kacesera, line 20: Remember that SQL doesn't particularly like apostrophes? I ran into this same bug in my own code. I'd LOVE to be able to use an apostrophe in my peep!

views/chitters.erb
```diff
@@ -0,0 +1,4 @@
+<% Chitter.all.each do |chitter| %>
+<h3><%=chitter.message%> </h3><h5><%= + "   Sent at: " + chitter.time%> </h5>
+<br>
+<%end%>
```
kacesera, line 4: rubocop offense!

app.rb
```diff
@@ -0,0 +1,19 @@
...
+  get '/' do
+    erb :index
+  end
+
+  post '/chitters' do 
+    Chitter.add(params[:peep])
+    redirect('/chitters/all')
+  end
+  
+  get '/chitters/all' do
```
kacesera, line 14: Is there a way to add this to the main site, or add a link so I can see the peeps, or go back to add a peep without physically typing '/' in the URL bar?

PR #1656, opened 2020-11-22
Chitter Challenge - Mel

TamMelPer:
Unfortunately, I cannot claim to have met any of the user stories.


README.md
```diff
@@ -35,41 +21,21 @@ As a Maker
...
-
-**ActiveRecord**
-- [ActiveRecord ORM](https://guides.rubyonrails.org/active_record_basics.html)
-- [Sinatra, PostgreSQL & ActiveRecord recipe](http://recipes.sinatrarb.com/p/databases/postgresql-activerecord?#article)
+Connect to the database:
+-----
+1. Connect to `psql`
+2. Create the database using the `psql` command `CREATE DATABASE chitter;`
+3. Connect to the database using the `pqsl` command `\c chitter;`
+4. Run the query I have saved in the file `01_create_peeps_table.sql`
```
joshuaretallick, line 38: Ah yeah I should have done this, good idea!

db/migrations/01_create_peeps_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60), time TIMESTAMP WITH TIME ZONE NOT NULL);
```
joshuaretallick, line 1: For 'message' I used VARCHAR(255) with my theory that it was closer to the character limit in a Twitter tweet, no idea if this is true or not

lib/database_connection.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def self.connection
+    @connection
+  end
+
+  def self.query(sql)
+    @connection.exec(sql)
+  end
+
+end
```
joshuaretallick, line 20: How are you finding the idea of defining 'self'? Is it become a bit clearer?

spec/spec_helper.rb
```diff
@@ -1,10 +1,10 @@
 require 'simplecov'
 require 'simplecov-console'
 require 'capybara/rspec'
-
+require_relative './setup_test_database'
 
 # Set the environment to "test"
-ENV['RACK_ENV'] = 'test'
+ENV['ENVIRONMENT'] = 'test'
```
joshuaretallick, line 7: I also used ENVIRONMENT, would be curious to know the difference between this and RACK_ENV

spec/features/post_peep_spec.rb
```diff
@@ -1 +1,10 @@
+require 'pg'
 
+feature 'Post a Peep' do
+  scenario 'A user can post a message to chitter' do
+    visit('/post')
+    fill_in('message', with: 'weekend challenges are so much fun')
```
joshuaretallick, line 6: Love this positive message!

lib/peep.rb
```diff
@@ -6,7 +6,7 @@ class Peep
   def self.all
     connection = PG.connect(dbname: 'chitter')
     result = connection.exec("SELECT * FROM peeps;")
-    result.map { |peep| peep['message']}
+    result.map { |peep| peep['message'] }.reverse
```
joshuaretallick: I used a similar method, apparently another way to look at it would be like in Line 8
SELECT * FROM peeps DESC; (syntax might be a little off but this reverses the list order too)

joshuaretallick (commented):
Good work Mel, your intentions were super clear and your commit messages were alot more helpful than mine I bet!

PR #1662, opened 2020-11-23
Weekend challenge(4) complete!

MykeNuLeng:


lib/peep.rb
```diff
@@ -0,0 +1,49 @@
+require 'pg'
+require_relative './database_connection'
+class Peep
+  attr_reader :id, :peep, :time
+
+  def initialize(id:, peep:, time:, username:)
+    @id = id
+    @peep = peep
+    @time = time
+    @username = username
```
benfurber, line 10: You're doing this in addition to the attr_reader?

lib/peep.rb
```diff
@@ -0,0 +1,49 @@
...
+    time = date_and_time[1]
+    hours_minutes_seconds = time.split(":")
+    hours = hours_minutes_seconds[0]
+    minutes = hours_minutes_seconds[1]
+    years_months_days = date.split("-")
+    years = years_months_days[0]
+    months = years_months_days[1]
+    days = years_months_days[2]
+    "#{hours}:#{minutes} #{days}/#{months}/#{years}"
+  end
```
benfurber, line 43: The length of this model is a [bad smell](https://en.wikipedia.org/wiki/Design_smell).

spec/chitter_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+
+    peeps = Peep.all.map { |peep| peep.peep }
+
+    expect(peeps).to include 'testing...'
+    expect(peeps).to include 'testing!'
+    expect(peeps).to include '1, 2, 3!'
+  end
+
+  it '.peep posts a peep to the database' do
+    new_peep = Peep.peep(peep: 'Testing... testing! 1, 2, 3!')
```
benfurber, line 19: Any better naming ideas than Peep.peep???

PR #1667, opened 2020-11-23
chitter

adam2021:


app.rb
```diff
@@ -0,0 +1,23 @@
+require 'sinatra/base'
+require 'pg'
+require_relative './lib/peep'
+
+class App < Sinatra::Base
+
+  attr_reader :content, :author
```
AdamusBG, line 7: What is this line doing? Think this might be accidental repetition from the peep class.

lib/peep.rb
```diff
@@ -0,0 +1,18 @@
...
+    @content = content
+  end
+
+
+  def self.all
+    result = PG.connect(dbname: 'chitter').exec("SELECT * FROM peeps;")
+    result.map do |peep|
+    Peep.new(author: peep['author'], content: peep['content'])
+  end
+  end
```
AdamusBG, line 16: Extra end statement

views/head.erb
```diff
@@ -0,0 +1,3 @@
+<head>
+    <link rel="stylesheet" href="style.css">
+</head>
```
AdamusBG, line 3: Love the idea of using a shared head.erb file to keep code DRY

views/index.erb
```diff
@@ -0,0 +1,30 @@
...
+
+  <div id="peeps">
+
+  <% @peeps.each do |peep| %>
+  <ul style="list-style: none;">
+  <div id="a_peep" style="width: 370px;">
+    <li><b><%= peep.content[:author] %></b> <br>
+      <%= peep.content[:content] %></li>
+    </div><!-- // a_peep -->
+    <div id="peep_length" style="float: right;">
```
AdamusBG, line 23: If you wanted to style these in your style.css file, you should use classes on lines 19 & 23 instead of ids.

PR #1671, opened 2021-01-02
Chitter Submission

Will-Helliwell:
Part way through 'harder' user stories (log-in/log-out). See Readme for full details.


README.md
```diff
@@ -1,5 +1,52 @@
-Chitter Challenge
-=================
+# Will's README
```
IrinaG29, line 1: Lovely work!
- Nicely written README
- The code is very easy to read, good use of the SRP in your methods and classes 
- Couldn't find anything that I would've done differently

lib/peep.rb
```diff
@@ -0,0 +1,58 @@
...
+  def initialize(content, time, username)
+    @content = content
+    @time = time
+    @user_id = username
+  end
+
+  def self.all
+    begin
+      if ENV["Environment"] == 'test'
+        con = PG.connect :dbname => 'chitter_test', :user => 'whelliwell1'
```
alicelieutier, line 16: Naming - don't use abbreviations.

lib/peep.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+  end
+
+  def self.add(content, time, user_id)
+    begin
+      if ENV["Environment"] == 'test'
+        con = PG.connect :dbname => 'chitter_test', :user => 'whelliwell1'
+      else
+        con = PG.connect :dbname => 'chitter', :user => 'whelliwell1'
+      end
```
alicelieutier, line 43: This code looks like it's repeated from above and could be extracted. - DRYed

lib/peep.rb
```diff
@@ -0,0 +1,58 @@
...
+      rs = con.exec "INSERT INTO peeps (content, time, user_id) VALUES ('#{content}', '#{time}', '#{user_id}')"
+
+      rescue PG::Error => e
+
+      puts e.message
+
+      ensure
+
+      con.close if con
+    end
```
alicelieutier, line 54: This one too. In fact, you could probably have code that takes a query string, and then does the connection + error handling.

lib/user.rb
```diff
@@ -0,0 +1,75 @@
+class User
+
+  @@current_username = "admin"
+  @@current_user_id = 0
```
alicelieutier, line 4: These should be stored in session not in model.

alicelieutier: Otherwise two different people connect to the server at the same time and can see each other's content.

views/peeps.erb
```diff
@@ -0,0 +1,22 @@
...
+
+  <head>
+  </head>
+
+  <body>
+    <h1>Latest Peeps:</h1>
+    <ul>
+      <% @peep.each do |peep| %>
+      <li>
+        <%= User.get_username(peep.user_id) %>:
```
alicelieutier, line 11: This is a lot of work for the view. Maybe the controller of even better, the peep model could have a method that fetches username for each peep.

app.rb
```diff
@@ -0,0 +1,42 @@
...
+    erb(:peeps_new)
+  end
+
+  post '/peeps' do
+    Peep.add(params[:peep_content], Time.now.strftime("%I:%M %p"), User.current_user_id)
+    redirect('/peeps')
+  end
+
+  get '/peeps' do
+    @peep = Peep.all.reverse
```
alicelieutier, line 21: Relying on reverse may cause problems + it should not be the controller job to sort them. It would be better to have a `all_in_chronological_order` method in Peep, that returns a sorted list (maybe with the username for each, too).

app.rb
```diff
@@ -0,0 +1,42 @@
...
+    erb(:peeps)
+  end
+
```
alicelieutier, line 24: Your peep routes are RESTful, but your sign-in/sign-up routes not as much. Remember, REST is a standard, but it's not everything either, and there are many ways to do routes. If you want to look more into this, this is a good place; https://stackoverflow.com/questions/7140074/restfully-design-login-or-register-resources

alicelieutier (commented):
Good job overall. You can make your code more organised and DRYer.
About planning ahead vs TDD -> They are not a contradiction! It makes sense to read all user stories and think about the overall system, but then use TDD to build one feature at a time instead of trying to build it all together.

PR #1672, opened 2021-01-02
Chitter Challenge - Lisa Bardelli

lisabardelli:


lib/tweet.rb
```diff
@@ -0,0 +1,31 @@
...
+  attr_reader :message, :username, :id, :time
+
+  def initialize(id:, username:, message:, time:)
+    @id = id
+    @username = username
+    @message = message
+    @time = Time.parse(time).strftime("%Y/%m/%d %k:%M:%S")
+    end
+
+  def self.display_tweets
```
xavierloos, line 13: You could call it only `self.display`/`self.add` so it's not repetitive tweet, as you are in the class Tweet it will make sense what are you calling to.

lib/database_connection.rb
```diff
@@ -0,0 +1,22 @@
...
+    @connection = PG.connect(dbname: db_name)
+    @connection.exec("DEALLOCATE ALL")
+    @connection.prepare('list_tweets', 'SELECT * FROM tweets ORDER BY time DESC;')
+    @connection.prepare('add_new_tweet', "INSERT INTO tweets(username, message) VALUES ($1, $2);")
+    @connection.prepare("insert_new_user", "INSERT INTO users (username, email, first_name, last_name, password) VALUES ($1, $2, $3, $4, $5);")
+    @connection.prepare('sign_in', "SELECT COUNT(*) FROM users WHERE username=$1 AND password=$2;")
+  end
+
+  class << self
+    attr_reader :connection
```
xavierloos, line 15: You can call it attr_reader without the class

dm/migrations/01_create_tables.sql
```diff
@@ -0,0 +1,24 @@
...
+  PRIMARY KEY(username)
+  );
+
+CREATE TABLE tweets(
+    tweet_id INT GENERATED ALWAYS AS IDENTITY, 
+    username VARCHAR(255) NOT NULL,
+    message VARCHAR(140) NOT NULL,
+    time timestamp without time zone default (now() at time zone 'utc'),
+    PRIMARY KEY(tweet_id),
+    CONSTRAINT fk_user
```
xavierloos, line 20: where is `fk_user` coming from?

xavierloos (commented):
To summarise: I like the way you code, I think it is straight forward.
Bonus for implementing a style on the home page.

alicelieutier (commented):
Well organised code, easy to read.

PR #1675, opened 2021-01-02
Natasha's Chitter Challenge

natashamcintyre:
I have questions about my MVC, CSS and project setup (rake!) - please see README.


app.rb
```diff
@@ -0,0 +1,80 @@
...
+    peep = @user.peeps.create(params[:peep])
+    peep.tags.create(user_id: params[:tag_user]) unless params[:tag_user] == "nil"
+    redirect '/peeps'
+  end
+
+  get '/users/new' do
+    erb :'users/new'
+  end
+
+  post '/save_user' do
```
PiperS52, line 42: I really like this duplicate check - good idea - I totally overlooked this

app.rb
```diff
@@ -0,0 +1,80 @@
...
+      user = User.create(params[:user])
+      session[:user_id] = user.id
+      redirect "/users/#{user.id}/welcome"
+    else
+      flash[:notice] = duplicate
+      redirect '/users/new'
+    end
+  end
+
+  get '/users/:id/welcome' do
```
PiperS52, line 54: I wonder whether this (54-56) could instead be captured within your 'peeps/index' within a conditional?

PiperS52:
Really great project, with some really cool features -  nice work getting to grips with ActiveRecord as well

app.rb
natashamcintyre: I like this idea, thank you 👍

app.rb
```diff
@@ -0,0 +1,80 @@
...
+    @peeps = Peep.all.order(created_at: :desc)
+    erb :'peeps/index'
+  end
+
+  get '/peeps/new' do
+    erb :'peeps/new'
+  end
+
+  post '/save_peep' do
+    peep = @user.peeps.create(params[:peep])
```
tatsiana-makers, line 33: Rather than creating the peep through `@user.peeps` could we do something like `Peep.create` here?

README.md
```diff
@@ -1,5 +1,81 @@
...
+
+I wanted to call the user_id column in the tag table "tagged_user_id", and call "peep.tagged_users" to retrieve them, but couldn't quite figure out how to change the names via activerecord.
+
+## My learning
+
+#### Advice needed:
+Project Setup - I tried but haven't yet worked out how to run the database migrations using rake. Help needed here!
+
+MVC - I think some of my controller is too 'fat' but I'm not sure how to make them thinner. Not sure I've got the MVC balance correct as the views also seem to be doing quite a lot. Is flash[:notice] done correctly? I haven't used next or now. Also, should I have more views? For example, one view for viewing the peeps before logging in and another view for viewing the peeps once logged in? This would minimise some of the <% if %> statements in the views, but just not sure if it's considered to be a better approach or not. Any advice would be really appreciated.
+
```
tatsiana-makers, line 51: Are the flash messages working as you would like them to?

What is the difference between using the `.now` with flash messages and not? The code comments in https://github.com/SFEley/sinatra-flash#setting-up might be useful

app.rb
```diff
@@ -0,0 +1,80 @@
+# HARDER User Stories:
+# If signed in, don't give "Sign Up" or "Sign In" options
+# Make sure paths flow and edit CSS
+
+require 'sinatra/base'
+require 'sinatra/flash'
+require_relative './lib/user.rb'
+require_relative './database_connection.rb'
+
+class Chitter < Sinatra::Base
```
tatsiana-makers, line 10: Which part of the controller felt too bloated to you? It feels ok to me, but perhaps you have an instinct or idea that something should be pulled out into a model?

spec/units/user_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+        name: "User2",
+        email: "Email",
+        username: "Username2",
+        password: "password"
+      )
+
+      expect(User.check_duplicate(user2)).to eq 'That email address is already registered.'
+    end
+  end
+end
```
tatsiana-makers, line 35: Nice testing!

views/index.erb
```diff
@@ -0,0 +1,17 @@
+<link rel="stylesheet" type="text/css" href="/stylesheets/chitter.css" />
+<body>
+  <div class="header">
+    <% if @user %>
```
tatsiana-makers, line 4: Having these if statements feels ok to me because it's the logic of what's displayed to the user

views/peeps/index.erb
```diff
@@ -0,0 +1,40 @@
+<link rel="stylesheet" type="text/css" href="/stylesheets/chitter.css" />
```
tatsiana-makers, line 1: If you split this into two views, one for before logging in and one after, there would be a repetition of how you would  show the peeps across the two views, which violates the DRY principle and is a potential source of bugs since you'd have to consistently change both of these views if you wanted to change how peeps were displayed in the future.

tatsiana-makers (commented):
Thanks for asking for a review! This is a really nice go at the Chitter challenge! You showed a really nice division of responsibility between the M, V, and C components (: I've left a few comments, let me know if you have any further questions 🙂

README.md
```diff
@@ -1,5 +1,81 @@
...
+Tag Table
+|id|peep_id|user_id|
+|-|-|-|
+
+I wanted to call the user_id column in the tag table "tagged_user_id", and call "peep.tagged_users" to retrieve them, but couldn't quite figure out how to change the names via activerecord.
+
+## My learning
+
+#### Advice needed:
+Project Setup - I tried but haven't yet worked out how to run the database migrations using rake. Help needed here!
```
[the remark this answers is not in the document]
natashamcintyre: This is so helpful, thank you. I spent a huge amount of time trawling the internet for something like this - I think my research skills need to improve!

app.rb
natashamcintyre: I did it this way because I read that this is a bit of ActiveRecord 'sugar' that means a peep will be created with the user_id column filled in with @user.id.

README.md
natashamcintyre: Yes I will explore this more, thank you

views/peeps/index.erb
natashamcintyre: Ah, so this justifies why it is better practice to use the if statements within the views to display different things depending on session data. Thanks!

app.rb
natashamcintyre: It was the post '/save_user' (row 42) and post '/sessions' (row 62) that just felt a bit long, but I'm not sure what could be pulled out into a model so perhaps they're okay

PR #1678, opened 2021-01-03
Ian does Chitter

iantfleming:
I completed up to challenge 4. Had I gone any further, I understand that there is certain code I would have to change   in order to make it work with multiple users. I didn't spend much time on the layout of my site but it should still be clear as to what is going on.


README.md
```diff
@@ -1,135 +1,50 @@
...
-require 'simplecov-console'
 
-SimpleCov.formatter = SimpleCov::Formatter::MultiFormatter.new([
-  SimpleCov::Formatter::Console,
-  # Want a nice code coverage website? Uncomment this next line!
-  # SimpleCov::Formatter::HTMLFormatter
-])
-SimpleCov.start
-```
+# How to create your database
```
WillDixon93, line 45: nice and clear instructions on setting up the database

spec/features/view_post_spec.rb
```diff
@@ -0,0 +1,15 @@
+require 'pg'
+
+feature 'view posts' do
```
WillDixon93, line 3: good cleaning of the database. maybe could be in the spec_helper so you don't have to do it in every feature test? I may be misremembering that you can do that so disregard if so.

spec/features/sign_up_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'pg'
+
+feature 'sign up to Chitter' do
+  before(:each) do
+    empty
+    add_base_posts
+  end
+  scenario 'A user can sign up and post as themselves to Chitter' do
```
WillDixon93, line 8: well structured, clear feature test

lib/chitter.rb
```diff
@@ -0,0 +1,38 @@
+require 'pg'
+
+class Chitter
+
+
+
+  def self.all
```
WillDixon93, line 7: could you create a setup_test_database spec in order to avoid writing out the if statements in each method. This would clear it up nicely. E.g. with the bookmarks 

def reset_table
  connection = PG.connect(dbname: "bookmark_manager_test")

  # Clear the bookmarks table
  connection.exec("TRUNCATE bookmarks;")
end

def add_bookmarks
  connection = PG.connect(dbname: "bookmark_manager_test")
  # Add the test data
  connection.exec("INSERT INTO bookmarks VALUES(1, 'http://www.makersacademy.com','Makers Academy');")
  connection.exec("INSERT INTO bookmarks VALUES(2, 'http://www.destroyallsoftware.com','Destroy all software');")
  connection.exec("INSERT INTO bookmarks VALUES(3, 'http://www.google.com','Google');")
end

spec/spec_helper.rb
```diff
@@ -8,6 +11,11 @@
 ])
 SimpleCov.start
 
+ENV['RACK_ENV'] = 'test'
+
+Capybara.app = Chitter_Main
+
+
```
WillDixon93, line 18: is it also worth putting this in to reset the table to empty? 
RSpec.configure do |config|
  config.before(:each) do
    reset_table

WillDixon93 (commented):
Really good (Y)

PR #1680, opened 2021-01-03
Chitter challenge attempted

lukewickens1989:
Week 4 weekend challenge - Chitter. Had issues trying to use DatabaseManager (could not install). So had to do it the old fashioned way. One test falling and had some issues applying the user.

### DIFF
--- a/db/migrations/02_adding_timestamp_column.sql
+++ b/db/migrations/02_adding_timestamp_column.sql
@@ -1,0 +1,2 @@
+ALTER TABLE cheeps ADD COLUMN posted_at TIMESTAMP;
+ALTER TABLE cheeps ALTER COLUMN posted_at SET DEFAULT now();
--- a/lib/cheep.rb
+++ b/lib/cheep.rb
@@ -1,0 +1,9 @@
+class Cheep
+  def self.all
+    [
+      "First cheep",
+      "Second cheep",
+      "Third cheep"
+     ]
+  end
+end
--- a/lib/cheep.rb
+++ b/lib/cheep.rb
@@ -3,9 +3,24 @@ require 'pg'
 class Cheep
 
   def self.all
-    connection = PG.connect(dbname: 'cheeps')
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'cheeps_test_database')
+    else
+      connection = PG.connect(dbname: 'cheeps')
+    end
+
     result = connection.exec("SELECT * FROM cheeps;")
     result.map { |cheep| cheep['message'] }
+  end
+
+  def self.create(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'cheeps_test_database')
+    else
+      connection = PG.connect(dbname: 'cheeps')
+    end
+  
+    connection.exec("INSERT INTO cheeps (message) VALUES('#{message}')")
   end
 
 end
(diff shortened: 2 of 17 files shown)


db/migrations/02_adding_timestamp_column.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE cheeps ADD COLUMN posted_at TIMESTAMP;
+ALTER TABLE cheeps ALTER COLUMN posted_at SET DEFAULT now();
```
Aracho1, line 2: Include the instruction in README to add the timestamp functions in the db

lib/cheep.rb
```diff
@@ -0,0 +1,39 @@
...
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'cheeps_test_database')
+    else
+      connection = PG.connect(dbname: 'cheeps')
+    end
+
+    result = connection.exec("SELECT * FROM cheeps;")
+    result.map do |cheep|
+      Cheep.new(id: cheep['id'], author: cheep['author'], message: cheep['message'])
```
Aracho1, line 23: Since timestamp is already part of the db, you could map the timestamp to the model as well

spec/cheep_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+      Cheep.create(author: "Lisa", message: "Second cheep")
+      Cheep.create(author: "Zoe", message: "Third cheep")
+      
+      cheeps = Cheep.all
+
+      expect(cheeps.length).to eq 3
+      expect(cheeps.first).to be_a Cheep
+      expect(cheeps.first.id).to eq cheeps[0].id
+      expect(cheeps.first.author).to eq 'Jason'
+      expect(cheeps.first.message).to eq 'First cheep'
```
Aracho1, line 19: Nice and clean tests, easy to follow what is being tested

views/cheep_feed.erb
```diff
@@ -0,0 +1,19 @@
...
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+    <link rel="stylesheet" href="">
+  </head>
+  <body>
+    <ul>
+    <% @cheeps.reverse.each do |cheep| %>
+      <li><%= cheep.author %>, <%= cheep.message %></li>
+    <% end %>
+    </ul>
+
```
Aracho1, line 17: You could consider adding a button which redirects to the create_cheep route

lib/cheep.rb
```diff
@@ -0,0 +1,39 @@
...
+
+  def self.create(message:, author:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'cheeps_test_database')
+    else
+      connection = PG.connect(dbname: 'cheeps')
+    end
+  
+   
+    result = connection.exec("INSERT INTO cheeps (message, author) VALUES('#{message}', '#{author}') RETURNING id, message, author")
```
Aracho1, line 35: Including special character in the message (e.g. '') raises an error. To avoid this, you could wrap the message inside like this $$#{message}$$

Aracho1 (commented):
Clean source codes and tests - they were highly readable and easy to follow. The database sets up and functions correctly as per the README. The posting function works well, creating the messages and authors correctly in reverse order.

My other suggestions are:
- You could add in README that users should run timestamp function to be added to the db
- To improve user experience, you could add buttons in the different routes to redirect to different pages e.g. adding cheeps from the feed page
- You could add the timestamp to the Cheep object since it's already part of the database and this way it can be printed out in the view
- To avoid raising errors when messages include special characters, consider wrapping the variable inside $$

PR #1686, opened 2021-01-03
A maker can post a peep, see all peeps in reverse chronological order…

IrinaG29:
…, the time they were posted, sign up

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,4 +5,5 @@ group :test do
   gem 'rubocop', '0.79.0'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
+  gem 'pg'
 end
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,45 @@
+require 'sinatra/base'
+require_relative './lib/chitter'
+require_relative './lib/user'
+require 'pry'
+
+class ChitterManager < Sinatra::Base
+  set :session_secret, 'super secret'
+  enable :sessions
+
+  get '/' do
+    'Chitter'
+  end
+
+  get '/chitter' do
+    @peeps = Chitter.all.reverse
+    erb :index
+  end
+
+  post '/chitter' do
+    Chitter.create(peep: params['peep'], time: Time.now)
+    redirect '/chitter'
+  end
+
+  get '/chitter/new' do
+    erb :'chitter/new'
+  end
+
+  get '/chitter/sign-up' do
+    erb :'sign_up'
+  end
+
+  post '/chitter/sign-up' do
+    @user = User.sign_up(name: params['name'], username: params['username'], email: params['email'], password: params['password'])
+    session[:user_id] = @user.id
+    redirect '/chitter'
+  end
+
+  def current_user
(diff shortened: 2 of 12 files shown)


Gemfile
```diff
@@ -5,4 +5,5 @@ group :test do
   gem 'rubocop', '0.79.0'
```
Will-Helliwell, line 5: I think you also need to require sinatra and capybara

app.rb
```diff
@@ -0,0 +1,45 @@
+require 'sinatra/base'
+require_relative './lib/chitter'
+require_relative './lib/user'
+require 'pry'
+
+class ChitterManager < Sinatra::Base
+  set :session_secret, 'super secret'
+  enable :sessions
+
+  get '/' do
```
Will-Helliwell, line 10: Perhaps your URLs could be a bit more RESTful? For example, your homepage does not have a function (and is not visited in your feature tests) and the peeps page will currently have a url of something like www.chitter.com/chitter. Perhaps call it peeps and allow the user to navigate there from the homepage?

app.rb
```diff
@@ -0,0 +1,45 @@
...
+    'Chitter'
+  end
+
+  get '/chitter' do
+    @peeps = Chitter.all.reverse
+    erb :index
+  end
+
+  post '/chitter' do
```
Will-Helliwell, line 19: post redirect get loop, lovely

app.rb
```diff
@@ -0,0 +1,45 @@
...
+    erb :'chitter/new'
+  end
+
+  get '/chitter/sign-up' do
+    erb :'sign_up'
+  end
+
+  post '/chitter/sign-up' do
+    @user = User.sign_up(name: params['name'], username: params['username'], email: params['email'], password: params['password'])
+    session[:user_id] = @user.id
```
Will-Helliwell, line 34: Love the use of the session variable to store the current user - much better than my class variable solution! I think this means you can have multiple users...

lib/chitter.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def self.create(peep:, time:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_manager_test')
+    else
+      connection = PG.connect(dbname: 'chitter_manager')
+    end
+
+    result = connection.exec("INSERT INTO chitter (peep, time) VALUES('#{peep}', '#{time}') RETURNING id, peep, time;")
+    Chitter.new(id: result[0]['id'], peep: result[0]['peep'], time: result[0]['time'])
```
Will-Helliwell, line 34: Great wrapping of the database result into a Ruby object (same thing for the User class)

Will-Helliwell (commented):
Great stuff Irina! I really liked the logic, and the way you wrap the database extractions into Ruby objects.

Only other general points I would make are that you are missing tests for your classes (and so I presume they haven't been TDD'd) and I couldn't run your program as there weren't any instructions on how to do so, or how to set up the databases etc. But I hope it looked bea-utiful!

PR #1699, opened 2021-02-01
Pete's chitter challenge

PetarSimonovic:


lib/peep.rb
```diff
@@ -0,0 +1,53 @@
...
+  def self.create(message:, sender:)
+
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+      else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    date = "#{Time.new.year}-#{Time.new.month}-#{Time.new.day}"
+    time = "#{Time.new.hour}:#{Time.new.min}"
```
HamishArro, line 43: Might have been better to use a timestamp provided by SQL? saves a few lines of code.

app.rb
```diff
@@ -0,0 +1,70 @@
...
+    else
+      redirect('/check_password')
+    end
+  end
+
+  get '/no_user' do
+    erb(:no_user_sign_up)
+  end
+
+  get '/check_password' do
```
HamishArro, line 32: might have been better to have this functionality in the model to keep the controllers skinny, however, it does provide extra error messages.

HamishArro (commented):
really great! :)

PR #1702, opened 2021-02-01
Chitter

bateso88:


spec/database_helpers.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'pg'
+
+def persisted_data(id:,table:)
+  connection = PG.connect(dbname: 'bookmark_manager_test')
```
GlykeriaStr, line 6: I know you changed it already! It should be chitter_test and not bookmark_manager_test.

lib/peep.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def self.all
+    result = DatabaseConnection.query('SELECT * FROM peeps;')
+    result.map do |peep|
+      Peep.new(id: peep['id'], message: peep['message'], time_last_altered: peep['time_last_altered'], user_id: peep['user_id'])
+    end
+  end
+
+  def self.create(message:, time_last_altered: Time.now, user_id:)
```
GlykeriaStr, line 21: Maybe the time could be in a specific format?

lib/peep.rb
```diff
@@ -0,0 +1,47 @@
+require_relative 'user'
+
+class Peep
```
GlykeriaStr, line 3: I love the idea of find_users_username and update!

spec/peep_spec.rb
```diff
@@ -0,0 +1,84 @@
...
+
+      expect(peeps.length).to eq 3
+      expect(peeps.first).to be_a Peep
+      expect(peeps.first.id).to eq peep.id
+      expect(peeps.first.message).to eq 'This is peep 1'
+    end
+  end
+
+  describe '.create' do
+    it 'adds a new bookmark to the database' do
```
GlykeriaStr, line 21: I don't think that's what you meant to say :)

views/sessions/new.erb
```diff
@@ -0,0 +1,20 @@
...
+  <%= flash[:notice] %>
+</h2>
+
+<form action="/sessions" method="post">
+  <input type="email" name="email" />
+  <input type="password" name="password" />
+  <input type="submit" value="Sign in" />
+</form>
+
+<h6>Not yet registered?</h6>
```
GlykeriaStr, line 11: Nice touch ! Also that there is on in users/new ! Shows consistency!

PR #1706, opened 2021-02-01
Incomplete chitter challenge

twigz826:


spec/features/posting_peeps_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'Posting a new peep' do
+  scenario 'A user can post a peep to Chitter' do
+    visit('/peeps/new')
+    fill_in('username', with: 'timmy_toes')
+    fill_in('peep', with: 'God I love coding so much I could cry')
```
KarstenFinlay, line 5: Interesting placeholder for the tests hahaha.

db/migrations/02_add_username.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE peeps ADD COLUMN username VARCHAR(20);
```
KarstenFinlay, line 1: I think this is a good step, though I reckon it would be easier from a user standpoint, just to have the username be part of the `01_create_peeps_table.sql`

KarstenFinlay (commented):
Honestly the code here works well, there is not much for me to add. It reads smoothly.

I mean the only real note (and it is an obvious one) is that you just need to write up the sign in method. Other than that, this is looking good.

PR #1709, opened 2021-02-01
Katy Chitter Challenge

day-katy:


views/peeps/index.erb
```diff
@@ -0,0 +1,141 @@
...
+<input type='submit' value='Peep!' />
+</form>
+
+
+<h2>
+  <%= flash[:notice] %>
+</h2>
+
+</div>
+</body>
```
RTurney, line 141: Create setup and use of html here.

RTurney (commented):
Really good overall! Obviously you have some issue with the user id which is throwing up errors but this is very good code. Good SRP overall and love the files layout. Also a really pretty website. Great effort.

PR #1710, opened 2021-02-01
Morne's chitter challenge

Mornevanzyl:
Currently debugging user id issue.


lib/peep.rb
```diff
@@ -0,0 +1,35 @@
...
+          else
+            connection = PG.connect(dbname: 'chitter')
+          end
+        result = connection.exec('SELECT * FROM peeps;')
+        result.map do |peep|
+            Peep.new(id: peep['id'], message: peep['message'], date: peep['date'], time: peep['time'])
+        end
+    end
+
+    def self.post(message:)
```
HolsDuckett, line 26: Maybe .create would be a better variable name, as .post seems very similar to the function of .all

lib/user.rb
```diff
@@ -0,0 +1,32 @@
...
+        result = connection.exec("SELECT * FROM users WHERE id = #{id};")
+        # result = connection.exec("SELECT * FROM users WHERE id = 1;")
+        User.new(id: result[0]['id'], email: result[0]['email'])
+    end
+
+    attr_reader :email, :id
+
+    def initialize(id:, email:)
+        @id = id
+        @email = email
```
HolsDuckett, line 30: Not sure on the etiquette of ruby initialize methods but I think it would read better if the initialize method was first. Same with the attr reader

spec/unit/peep_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+        it 'returns a list of peeps' do
+            Peep.post(message: "Test peep 1")
+            Peep.post(message: "Test peep 2")
+            Peep.post(message: "Test peep 3")
+            peeps = Peep.all
+
+            expect(peeps.length).to eq 3
+            expect(peeps.first).to be_a Peep
+        end
+    end
```
HolsDuckett, line 16: Not sure you have a test for reverse chronological order. Could just change bottom of .all test to be expect(peeps.first).to eq('Test peep 3')

spec/unit/user_spec.rb
```diff
@@ -0,0 +1,24 @@
+require 'user'
+require 'database_helpers'
+
+describe User do
+    describe '.create' do
+        it 'create a new user' do
+            user = User.create(email: 'test@chitter.com', password: 'password')
+            persisted_data = persisted_data(table: :users, id: user.id)
```
HolsDuckett, line 8: What does persisted_data(table: :users, id: user.id) return??

views/peeps/list.erb
```diff
@@ -0,0 +1,9 @@
+<% if @user %>
+    <h2>Welcome, <%= @user.email %>!</h2>
+<% end %>
+
+<% @peeps.each do |peep| %>
+    <%= peep.date %>
+    <%= peep.time %>
+    <%= peep.message %><br>
```
HolsDuckett, line 8: This is really coherent and easy to read

views/peeps/list.erb
```diff
@@ -0,0 +1,9 @@
...
+<% end %>
```
HolsDuckett, line 9: You could think about adding the peep form to the list page so the user can see the list of peeps whilst creating a new one.

app.rb
```diff
@@ -0,0 +1,39 @@
+require 'sinatra/base'
+require_relative './lib/peep'
+require_relative './lib/user'
+
+class Chitter < Sinatra::Base
+    enable :sessions
+
+    get '/' do
+        erb :'peeps/index'
```
HolsDuckett, line 9: Did you create a peeps folder with an index, list and new .erb file within it? Good organisation of files

HolsDuckett (commented):
Overall your code is really easy to read and organsied well. Only improvements i could would be cleaning up the lapyout of files, adding the new peep form to the homepage of Chitter with the list of peeps and having a go at some CSS.

PR #1712, opened 2021-02-01
Hamish Arro

HamishArro:
Sorry it's not completed

### DIFF
--- a/lib/user.rb
+++ b/lib/user.rb
@@ -1,0 +1,29 @@
+require_relative 'encryptor.rb'
+require_relative 'database_connection.rb'
+
+class User
+  attr_reader :user_id, :username
+
+  def initialize(user_id: user_id, username: username)
+    @user_id = user_id
+    @username = username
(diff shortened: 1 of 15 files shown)


lib/user.rb
```diff
@@ -0,0 +1,34 @@
+require_relative 'encryptor.rb'
+require_relative 'database_connection.rb'
+
+class User
```
PetarSimonovic, line 4: Very clear code - easy to read and all methods are well-named

spec/unit/encryptor_spec.rb
```diff
@@ -0,0 +1,20 @@
+require 'encryptor'
+
+describe String do
```
PetarSimonovic, line 3: This is good. I didn't encrypt my passwords.

PetarSimonovic (commented):
What you've got so far looks good: short classes, well-named, easy to read. I like the encryption element too

PR #1715, opened 2021-02-28
Ian_McNicholas Chitter_Challenge

ianmcnicholas:


ianmcnicholas:
I have really struggled with this weekend challenge.  The first three user stories I was able to complete by simply copying and pasting my code from the afternoon Bookmarks Manager challenge this week.
After that, I had no idea.
Something that was really frustrating/disappointing was the fact that the work in the afternoon challenge that dealt with having a new user sign up to the app was in the last chapter, something that I expect most people wouldn't have gotten anywhere near during the week.

spec/features/posting_peeps_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'posting peeps' do
```
olgachmaj, line 1: I like your tests, they're super clear and easy to read. Also very detailed making it easy to pick up your code really fast.

lib/peep.rb
```diff
@@ -0,0 +1,44 @@
...
+  attr_reader :id, :peep, :created_at
+
+  def initialize(id:, peep:, created_at:)
+    @id = id
+    @peep = peep
+    @created_at = created_at
+  end
+
+  def self.all
+    if ENV['ENVIRONMENT'] == 'test'
```
olgachmaj, line 13: Something I have to do as well - it's good to dry up the data base connection based on enviroment variable into seperate file and wrap it into method there. It helps to make code look much cleaner especially later on where there is a lot of methods that connect to the data base.

app.rb
```diff
@@ -0,0 +1,19 @@
...
+require './lib/peep.rb'
+
+class Chitter < Sinatra::Base
+
+  get '/' do
+    @peeps = Peep.all
+    erb(:index)
+  end
+
+  post '/postpeep' do
```
olgachmaj, line 12: Something I have to improve as well, again 😄 We should make the routes as RESTful as possible, in this case the route would look a little like that '/peep/new'

app.rb
```diff
@@ -0,0 +1,19 @@
...
+    Peep.create(peep: params["peep"])
```
olgachmaj, line 13: Bit nit picky but convention is to user params[:peep]. Using a symbol instead of string makes sure that it's going to always be unmutable.

olgachmaj (commented):
I like your code, it's definitely following SRP and is easy to pick up and read!

PR #1716, opened 2021-02-28
Chitter-challenge

a-nawrocka:


AdamP2000:
Code is clear and easy to see what each method and variable does - Well named. And it works.

Only thing it doesn't have is a README to explain how to use the databases.

PR #1717, opened 2021-02-28
Chitter-challenge - WK4

sandyMax974:
All `STRAIGHT UP` user stories have been implemented. I didn't get a chance to attempt the harder one as I decided to play with CSS instead.
I didn't get to implements solution for some edge cases but I've put my thoughts at the bottom of the README.

Feedbacks on any of the below would be welcome:

- Use of isolated test
- Use of SRP
- Use of RESTful route
- How to handle very long database query string
- Anything else


lib/chitter.rb
```diff
@@ -0,0 +1,18 @@
+require_relative 'peep'
+
+class Chitter
+
+  def self.create(username, content)
+    connection = PG.connect :dbname => "chitter_#{ENV['RACK_ENV']}"
+    result = connection.exec("SELECT * FROM users WHERE username = '#{username}'").first
+    insertion = connection.exec("INSERT INTO peeps (user_id, posted_time, content) VALUES(#{result['id']}, current_timestamp, '#{content}') RETURNING id, user_id, posted_time, content").first
```
kerisic, line 8: Could think about extracting database connection to run at the beginning of running the app - and extract connection.exec into a database connection class such as [this ](https://github.com/makersacademy/course/blob/master/bookmark_manager/walkthroughs/14.md)

views/feed.erb
```diff
@@ -0,0 +1,81 @@
+<!DOCTYPE html>
+  <head>
+    <title>Home/Chitter</title>
+    <style>
+      body {
+        font-family: "Helvetica Neue", Roboto, "Segoe UI", Calibri, sans-serif;
+        background-color: white;
+        background-image: url("https://wallpaperaccess.com/full/1459046.png")
+      }
```
kerisic, line 9: Similar css across views can be saved in an external stylesheet where views could have a link to.

db/migration/02_seeting_up_peeps_database.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps (id SERIAL PRIMARY KEY, user_id INT, posted_time TIMESTAMP, content VARCHAR(280));
```
kerisic, line 1: user_id in peeps could be referenced as a foreign key in table creation as it can be linked to the primary key in the user table

lib/peep.rb
```diff
@@ -0,0 +1,11 @@
...
+  attr_reader :id, :username, :posted_time, :content
+
+  def initialize(id, username, posted_time, content)
+    @id = id
+    @posted_time = posted_time
+    @content = content
+    @username = username
+  end
+
+end
```
kerisic, line 11: Good class encapsulation so that each class has a clear responsibility!

lib/user.rb
```diff
@@ -0,0 +1,15 @@
+class User
+  attr_reader :name, :lastname, :username, :email, :id
+
+  def initialize(name, lastname, username, email, password)
+    connection = PG.connect :dbname => "chitter_#{ENV['RACK_ENV']}"
+    insertion = connection.exec("INSERT INTO users (name, lastname, username, email, password) VALUES('#{name}', '#{lastname}', '#{username}', '#{email}', '#{password}') RETURNING id, name, lastname, username, email, password")
```
alicelieutier, line 6: It would be nicer to read this on several lines. You can look up ruby multiline strings as there are a number of solutions to this, and the one I give here may not be your favourite, but you could do:
```rb
query = %{
  INSERT INTO users (name, lastname, username, email, password)
  VALUES('#{name}', '#{lastname}', '#{username}', '#{email}', '#{password}')
  RETURNING id, name, lastname, username, email, password;
}
insertion = connection.exec(query)
```

Also this way you can easily log the query by adding a `puts query` between the query and insertion.

spec/features/feed_spec.rb
```diff
@@ -0,0 +1,15 @@
+feature 'Feed' do
+  scenario 'A user can see a feed of all posted peeps' do
+    connection = PG.connect :dbname => "chitter_#{ENV['RACK_ENV']}"
+    user = User.new('Sandy', 'Pounoussamy', 'sandy974', 'pounoussamy.sandy@gmail.com', '1234')
+    peep = "Who wants to know what I had for lunch?!"
+    connection.exec("INSERT INTO peeps (user_id, posted_time, content) VALUES('#{user.id}', current_timestamp, '#{peep}')")
```
alicelieutier, line 6: In a feature test, you may want to only interact with your app through the user interface. So you could have your user create tow peeps, and then see them, for example.

spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'Posting' do
+  scenario 'A user can post/create a new peep' do
+    visit('/signup')
+    signup
+
+    fill_in('content', with: "This is an automated peep")
+    click_button('Post')
+
+    expect(page).to have_content("This is an automated peep")
+  end
```
alicelieutier, line 10: This is good!

spec/units/chitter_spec.rb
```diff
@@ -0,0 +1,37 @@
+require 'chitter'
+
+describe Chitter do
+
+  describe '.list' do
+    it 'display all the peeps' do
+      connection = PG.connect :dbname => "chitter_#{ENV['RACK_ENV']}"
+      demo_user_one = connection.exec("INSERT INTO users (name, lastname, username, email, password) VALUES('Hannah', 'Lowe', 'hanalo13', 'test@gmail.com', '1234') RETURNING id, name, lastname, username, email, password").first
+      demo_user_two = connection.exec("INSERT INTO users (name, lastname, username, email, password) VALUES('Sandy', 'Pounoussamy', 'sandy974', 'mytest@gmail.com', '1234') RETURNING id, name, lastname, username, email, password").first
```
alicelieutier, line 9: These could be helper methods.

views/feed.erb
alicelieutier: agreed.

alicelieutier (commented):
Overall, your code looks like it works, so you've definitely cracked how to create a webapp that talks to a database.
What you can improve most on here IMO is the readability and organisation of your code. 
@kerisic added some good suggestions, and I would push even further, especially about naming and line length.

See comments below.

Here's an article I wrote about naming in code: https://medium.com/swlh/an-opinionated-guide-to-naming-your-code-aimed-at-new-developers-6f769d60e098

PR #1722, opened 2021-03-01
Chitter-Challenge

peter-james-allen:
Complete
Covers most user stories
Looks quite nice
Added additional features such as search by hashtag and user

Unfinished
Does not email user when tagged in a post
Testing is not complete
Refactor

To Improve
Add edit and delete buttons

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,3 +6,8 @@ group :test do
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
 end
+
+gem 'sinatra'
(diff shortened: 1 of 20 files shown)


iainaitken:
This is really good. I like the design of the web pages and overall the code is easy to follow and well-written. I'm not sure how much useful feedback I can give, to be honest, as I didn't get to adding users and I'm not sure I understand half of the code without studying it for a while. I'll see if I can add some specific comments shortly.

PR #1733, opened 2021-03-01
omg im chitter famous

spaek14:


charlierdm:
Hi Sol,

- nice work on getting to the sign in part, your app.rb is very clear in what's happening. 
- nice method on getting the timestamp in there. 
- There is no config.ru file so I'm not sure how you ran it to see what it looked like? 
- It would be nice to have a readme or some info about your project but I understand that this one was very time consuming. 
- spechelper/database_setup_helper all look good and setup well.
- 'liberal' testing like you say but I'm sure given more time these would be bulked up a bit! 
- I think you've done well the user and chitter separate classes :)
- I've just spotted that you do have a readme... in the future definitely set your readme to be called README.md so it appears first thing on the page. 
- I think that the app file should be a class so that it can inherit from sinatra base, perhaps this isn't essential though?
- Overall I think you show a good understanding of the concepts covered in week4 and have done well with this challenge!

PR #1742, opened 2021-03-28
Chitter Challenge :)

louballantyne:


spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+  scenario 'User navigates to Homepage' do
+    visit '/'
+    expect(page).to have_text "Chitter"
+  end
+end
+
+feature "Posting peeps" do
+  scenario "User adds a peep saying they love Makers" do
+    login
+    fill_in('peep_text', :with => 'I love Makers')
```
Emmapr123, line 19: cute first peep! Makers is the best

spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,21 @@
+require_relative "../../app"
+require_relative "../helper_methods"
+
+feature "Homepage" do
+  before do
+    clear_table
```
Emmapr123, line 6: why are you clearing the table here and not in the spec_helper file?

user_stories.md
```diff
@@ -0,0 +1,42 @@
+As a Maker
```
Emmapr123: Very nice idea to add this to a separate file, instead of having such a full readme!

lib/peeps.rb
```diff
@@ -27,7 +27,7 @@ def self.post(text, user_id, date, time)
 
   def self.all
     @peeps = []
-    DBConnection.connection.exec("SELECT * FROM peeps").map do |peep|
+    DBConnection.connection.exec("SELECT * FROM peeps ORDER BY date DESC, time DESC").map do |peep|
```
Emmapr123: nice way of dealing with the order

app.rb
```diff
@@ -2,18 +2,21 @@
...
   get '/' do
+    @user = User.find_name_by_id(session[:user_id])
     erb(:index)
   end
 
   post '/login' do
-    session[:username] = params[:username]
-    session[:password] = params[:password]
+    user = User.authenticate(params[:email],params[:password])
+    session[:user_id] = user.first['id'] unless user.first == nil
```
Emmapr123: Why do you have to use .first here? Is there an other way you could have written the user file with which you are able to find the id in a more efficient way?

spec/features/authentication_spec.rb
```diff
@@ -0,0 +1,11 @@
+feature 'authentication' do
+  it 'a user can sign in' do
+    create_account
+    visit '/'
+    fill_in(:email, with: 'caesar@conures.com')
+    fill_in(:password, with: 'parrots1')
```
Emmapr123, line 7: got to love the parrots

spec/units/user_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+      User.create('Bart', 'bart@conures.com', 'parrots1')
+    end
+  end
+
+  describe '.find_name_by_id' do
+    before do
+      clear_table
+    end
+    it 'finds a user by ID' do
+      DBConnection.connection.exec("INSERT INTO users (name, email, password, id) VALUES ('Bart', 'bart@conures.com', 'parrots1', '100');")
```
Emmapr123, line 30: Is it not easier to use .create here instead of going straight to the connection?

app.rb
```diff
@@ -4,24 +4,34 @@
 require_relative 'db_connection_helper'
```
Emmapr123, line 4: LOVE the commit name, I haven't heard the word fugly in ages!

views/index.erb
```diff
@@ -1,27 +1,43 @@
...
-<% end %>
-
-<ul>
-<% Peeps.all.each do |peep| %>
-<li class = "peep" >
-  <%=peep.date %> <%=peep.time %><br>
-  <%=peep.name%> peeped: <%=peep.text %><p>
-<% end %>
-</ul>
+<link rel="stylesheet" href="/style.css">
```
Emmapr123, line 1: If you want to link your stylesheet to all files without having to go through every single file, you can add it to a layout.erb, which uses yield in the body. That means that every single page has got the same stylesheet - that way you only need to specify a stylesheet in each erb file if it uses a different one!

db/migrations/02_create_user_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE users (id SERIAL PRIMARY KEY, email VARCHAR(60), password VARCHAR(100), name VARCHAR(30));
```
Emmapr123, line 1: Great work DRYing up these files

spec/units/peeps_spec.rb
```diff
@@ -10,11 +10,22 @@
...
       expect(Date.parse(Peeps.all[0].date)).to be > Date.parse(Peeps.all[1].date)
     end
+    it 'retrieves all peeps from the database' do
+      expect(Peeps.all.length).to eq 3
+    end
+    it 'skips a Peep if there is no text' do
+      Peeps.post_peep('', '1')
+      expect(Peeps.all.length).to eq 3
+    end
+    it 'removes apostrophes from a tweet' do
```
Emmapr123, line 26: great idea to get rid of these! did not get around to doing this myself, but really helps to prevent bugs

public/style.css
```diff
@@ -0,0 +1,61 @@
+#heading img {
```
Emmapr123, line 1: Id love to see what it actually looks like!

views/index.erb
```diff
@@ -0,0 +1,43 @@
...
+<html>
+<head>
+  <title>CHITTER :: Home</title>
+</head>
+<body>
+  <h2>
+    <%= flash[:notice] %>
+  </h2>
+  <div id = "heading">
+    <h1>Chitter <img src = "logo.png"></h1>
```
Emmapr123, line 12: Also, love that you added a logo and background!

views/login.erb
```diff
@@ -0,0 +1,22 @@
+<link rel="stylesheet" href="/style.css">
+<!DOCTYPE html>
+<html>
+<head>
+  <title>CHITTER :: Login </title>
+</head>
+<body>
+  <div id = "new_user_heading"
```
Emmapr123, line 8: did this still work? (since you didnt close the div)

Emmapr123 (commented):
Great work! would really love to see what it looks like, the stylesheet sounds lovely!

views/index.erb
louballantyne: Thanks for this tip :D

.travis.yml
```diff
@@ -5,6 +5,10 @@ services:
```
Emmapr123, line 5: just ran it through my localhost and I LOVE it ! It looks really great. I did have to run the tables in different order as well, what you commented on mine too, but aside from that everything seems to work perfectly!

louballantyne: Thanks :) Creating table order is definitely something I'll make sure to check for next time!

PR #1750, opened 2021-03-29
Practically non-existent submission

inbarvernia:
Hardly spent any time on this, so very little code to actually review! Only a couple of feature tests there, and no unit tests or model. README also mostly untouched, except for set-up instructions temporarily placed at the bottom.


cspoppuppy (commented):
Nicely structured html, all good :)

PR #1761, opened 2021-04-25
Abdur chitter

Abd27:


lib/peep.rb
```diff
@@ -0,0 +1,21 @@
+require_relative 'database_connection'
+
+class Peep
+  attr_reader :text, :time
+
+  def initialize(args = {})
+    @user = args[:user] || 'Anonymous'
```
YohannTisserand, line 7: nicely done

spec/setup_test_database.rb
```diff
@@ -0,0 +1,8 @@
+require 'pg'
+
+p 'Setting up test database...'
+
+def setup_test_database
+  connection = PG.connect(dbname: 'chitter_test')
+  connection.exec('TRUNCATE peeps;')
```
YohannTisserand, line 7: you could also truncate the user table

PR #1764, opened 2021-04-26
Tiffany's Chitter Challenge

tiffanyvallo:
User stories 1-3 completed. User Story 4 partial completion only - all tests and methods created and passed, but web app not working after adding users


fizzy-fifs (commented):
This is a really good piece of work. There is a clear separation of responsibilities between the classes, model, view and controller. I like the idea of having a flip button, giving the user the option to decide how they want to view their peeps.

PR #1774, opened 2021-05-23
weekend challenge submission

rdupplaw:


lib/database_connection.rb
```diff
@@ -0,0 +1,16 @@
+require 'pg'
+
+# Handles persistent connection to database
+class DatabaseConnection
+  class << self
```
GunelMC, line 5: Your code is just mind-blowing! Can you explain what this code is doing?

PR #1775, opened 2021-05-23
chitter challenge

cms718:


app.rb
```diff
@@ -0,0 +1,54 @@
...
+
+class Chitter < Sinatra::Base
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    @user = session[:user]
```
rdupplaw, line 16: storing the whole user object in session can create issues because session can't be very big. it might be better to just store user id in session then find the corresponding user when you need them

app.rb
```diff
@@ -0,0 +1,54 @@
...
+require_relative 'database_connection_setup'
+
+class Chitter < Sinatra::Base
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
```
rdupplaw, line 15: the RESTful route here would be /peeps or /messages

app.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  get '/register' do
+    erb(:register)
+  end
+
+  get '/login' do
+    erb(:login)
+  end
+
+  post '/' do
```
rdupplaw, line 29: restful route would be /peeps or /messages

app.rb
```diff
@@ -0,0 +1,54 @@
...
+    redirect('/')
+  end
+
+  post '/login' do
+    session[:user] = User.login(params['email'], params['password'])
+    redirect('/')
+  end
+
+  post '/logout' do
+    session[:user] = User.logout
```
rdupplaw, line 40: you don't need a User method here. You could just say session.clear or session[:user] = nil to log them out

lib/database_connection.rb
```diff
@@ -0,0 +1,15 @@
...
+  def self.setup(database)
+    @connection = PG.connect(dbname: database)
+  end
+
+  def self.connection
+    @connection
+  end
+
+  def self.query(query_string)
+    @connection.exec(query_string)
```
rdupplaw, line 13: what would happen if a user sent a peep with a ' or a ; in it? Message.self just interpolates the message into the SQL string, then DatabaseConnection.query would fail because the sql string is invalid.

rdupplaw: You can use connection.exec_params instead, which can interpolate data into placeholders for you and avoid this error

lib/user.rb
```diff
@@ -0,0 +1,42 @@
...
+    User.new(result['first_name'], result['last_name'], result['username'], result['email'], result['id'])
+  end
+
+  def self.logout
+    nil
+  end
+
+  def self.login(email, password)
+    result = DatabaseConnection.query(
+      "SELECT * FROM users WHERE email = '#{email}' AND password = '#{password}'"
```
rdupplaw, line 23: you shouldn't store passwords in plain text. you could use bcrypt to create a hash of the password to store, and to check the hash of a given password against the hash of a stored password

lib/user.rb
```diff
@@ -0,0 +1,42 @@
...
+    )
+    User.get_current(result.first['id'])
```
rdupplaw, line 25: why call another method to make another query to the database for user info when you already have the user info here? you could just call User.new

spec/user_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+  
+  describe '.logout' do
+    it 'sets the current user to nil' do
+      User.create('bob', 'michel', 'itsbob', 'bob@bob.com', 'password')
+      User.logout
+      expect(User.current).to eq(nil)
+    end
+  end
+
+  describe '.login' do
```
rdupplaw, line 20: this test is failing for me, but feature is working. i can't work out why.

PR #1809, opened 2021-08-22
Weekend challenge: week four

edpackard:
A tough challenge - I'd hoped to include a log in/log out feature (hence why there are encrypted passwords and unique email requirements in my code) but ran out of time.

I'm happy with the functionality of my code, and think my tests are OK, would appreciate feedback on:

1) whether my tests - both unit and feature - cover the core functionality of the program: have I missed anything? Could I make the existing tests more rigorous?

1a) I  included a User instance in my  peep_spec.rb tests: I would not include another class usually, but is this OK when testing with databases? I know the bookmark challenge had something similar. Is it possible to mock when using databases, and were we expected to?

2) how clear is my code - I found adding the database interface made it harder to write 'clean' code: is it clear what's going on, especially in my peep and user classes?

3) how could I improve my controller/routes? I think it's fairly lean, but I haven't done much reading around REST yet.

4) have I used too many attr readers in my classes?


spec/unit/peep_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+      Peep.add('This is a peep', '17:00:00')
+      
+      peep = Peep.feed.first
+      
+      expect(peep.handle).to eq '@Ed209'
+      expect(peep.content).to eq 'This is a peep'
+      expect(peep.time).to eq '17:00'
+    end
+
+    it 'sorts array of peeps into reverse chronological order based on time and date' do
```
ClaimingCookie5, line 32: I think there might be too much going on in this test, what you're testing is different to what th `it` is saying.

PR #1815, opened 2021-08-23
Lizzy Ballantine - Chitter Challenge

eballantine:


lib/peep.rb
```diff
@@ -0,0 +1,34 @@
+require 'PG'
+
+class Peep
+  attr_reader :peep_id, :peep, :user, :timestamp
+
+  def initialize(id:, peep:, user:, timestamp:)
+    @peep_id = id
```
Nicola-Carroll, line 7: I'm slightly unclear on what the \peep_id\ is and when/how it is generated? If it's an id which is used purely for indexing within the \peeps\ table then wouldn't it be best for it to be automatically generated by the table itself?

lib/user.rb
```diff
@@ -0,0 +1,58 @@
+class User 
+  attr_reader :user_id, :name, :username, :email, :password
```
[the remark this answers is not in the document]
Nicola-Carroll: also - i think a lot of web frameworks either have something built in or have a popular compatible library which deals with [password hashing ](https://learn.co/lessons/sinatra-password-security) so that you don't actually have to store the password at all, and instead just a hash of it

lib/peep.rb
```diff
@@ -0,0 +1,34 @@
...
+    @peep = peep 
+    @user = user
+    @timestamp = timestamp
+  end
+    
+  def self.add(peep, user)
+    results = DatabaseConnection.connection.exec_params("
+      INSERT INTO peeps (peep, user_fk, datetime) 
+      VALUES ($1, $2, CURRENT_TIMESTAMP)
```
Nicola-Carroll, line 16: what are the variables `$1` and `$2` ? aren't these global variables? since I didn't try the exercise i feel unqualified to comment but what was the reason that you used global variables?

PR #1822, opened 2021-08-23
PR for review

Battery0:
PR for review


db/migrations/01_setup_chitter_database.sql
```diff
@@ -0,0 +1,2 @@
+CREATE DATABASE chitter;
```
gjstirling, line 1: Clever ! didn't think to run these in a stored file. Nice one.

lib/peep.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  attr_reader :id, :message
+
+  def initialize(id = nil, message = "no message entered")
+    @id = id
+    @message = message
+  end
+
+  def self.all_peeps
+    database_connection
```
gjstirling, line 13: Good use of web helper

lib/peep.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+  end
+
+  def self.create(message = "no message entered")
+    database_connection
+    result = @con.exec_params("INSERT INTO peeps (message) VALUES ($1) RETURNING id, message;", [message])
+    Peep.new(result[0]["id"], result[0]["message"])
+  end
+
+  def self.newest_peeps
```
gjstirling, line 26: SRP !!!

PR #1845, opened 2021-10-17
Bernard's Chitter Challenge

bernardleanse:
I was blocked for a long time to work out why all my tests were inserting data into the main database not the test, found out it was in the database connection class. This limited how far I was able to get.


README.md
```diff
@@ -1,131 +1,25 @@
...
 
-Challenge:
--------
+Made domain model on draw.io, plan to start with Users and Peeps databases.
 
-As usual please start by forking this repo.
+# Database Setup
 
-We are going to write a small Twitter clone that will allow the users to post messages to a public stream.
+Create database in postgres with "CREATE DATABASE dbname;" and run the sql commands found in db/migrations to setup the dev database.
```
ConorButler, line 9: Could make the setup easier by using a Rakefile to set up the database.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+  enable :sessions
+
+  get '/' do
+    
+    @peeps = Peep.all
+    @user = User.find(id: session[:user_id]) 
+    puts "Theres a user #{@user} " if @user
```
ConorButler, line 18: Best to proof read your code so you don't forget to remove print statements like this, which could be a security risk.

app.rb
```diff
@@ -0,0 +1,40 @@
...
+  get '/users/new' do
+    erb :sign_up
+  end
+
+  post '/users' do
+    user = User.create(email: params[:email], username: params[:username], 
+password: params[:password])
+    session[:user_id] = user.id
+    redirect '/'
+  end
```
ConorButler, line 37: Good RESTful routes.

db/migrations/01_users_table_setup.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE users(id SERIAL PRIMARY KEY, email varchar(60), username varchar(60), password varchar(60));
```
ConorButler, line 1: Could include First name and Last name fields on sign up to meet the specification.

db/migrations/02_peeps_table_setup.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, user_id int REFERENCES users(id), content varchar(140), date_posted date, time_posted time);
```
ConorButler, line 1: Date_posted and Time_posted can be merged into one field. Ruby has DateTime or just using a Time object would still work.

lib/peep.rb
```diff
@@ -0,0 +1,24 @@
+class Peep
+  attr_reader :content, :date
+
+  def initialize(id: nil, content: nil, date: nil)
```
ConorButler, line 4: There probably shouldn't be a scenario where you create a Peep with any of these values as nil, as they are all required (and should be NOT NULL in the database).

lib/peep.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    result.map do |peep| 
+      Peep.new(id: peep['id'], content: peep['content'], date: peep['date_posted'])
+    end
+
+  end
+
+  def self.create(content:)
+    DatabaseConnection.query("INSERT INTO peeps(content, date_posted) VALUES ($1, $2) ", 
+[content, Date.today.to_s])
```
ConorButler, line 21: I think Postgresql can handle a Time/Date object as input, so converting to a string might not be necessary.

spec/features/posting_peep_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+    Peep.create(content: "I should be here too")
+    visit '/'
+    expect(page).to have_content "hello"
+    expect(page).to have_content "I should be here too"
+  end
+  
+  xscenario "reverse chronological order" do
+    Peep.create(content: "hello")
+    Peep.create(content: "I should be here too")
+    expect("I should be here too").to appear_before("hello")
```
ConorButler, line 20: Nice test, I didn't know you could do this.

spec/unit/peep_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it "should create a peep" do
+      
+      expect(peep.content).to eq "hello there"
+      
+    
+    end
+
+    it "should have a date" do
+
+      expect(peep.date).to eq Date.today.to_s
```
ConorButler, line 38: You can test this inside the Peep.create test, testing that the created Peep has all of the fields.

views/layout.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<head>
+  <title>Chitter</title>
+  <link href="/styles.css" rel="stylesheet" type="text/css">
+  <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.6.0/dist/css/bootstrap.min.css" integrity="sha384-B0vP5xmATw1+K9KRQjQERJvTumQW0nPEzvF6L/Z6nronJ3oUOFUFpCjEUQouq2+l" crossorigin="anonymous">
+</head>
+
+<body>
+<div class="content">
+<%= yield %>
```
ConorButler, line 12: Does this yield to anything?

PR #1849, opened 2021-10-17
Serena's Chitter Challenge

serenabertozzi:
I've managed to complete four user stories plus half of the fifth (I'm working on the unhappy paths). 
I've refactored using ORM.
Unfortunately, I wasn't able to complete all the user stories and add styling, due to the lack of time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,17 +1,14 @@
 Chitter Challenge
 =================
+### Setting up the database
 
-* Feel free to use Google, your notes, books, etc. but work on your own
-* If you refer to the solution of another coach or student, please put a link to that in your README
-* If you have a partial solution, **still check in a partial solution**
-* You must submit a pull request to this repo with your code by 9am Monday morning
+* Connect to psql and create the chitter and chitter_test databases.
+`CREATE DATABASE chitter;`
+`CREATE DATABASE chitter_test;`
 
-Challenge:
--------
-
-As usual please start by forking this repo.
-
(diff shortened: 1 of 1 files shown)


spec/features/view_peeps_spec.rb
```diff
@@ -0,0 +1,21 @@
+feature "Viewing a list of peeps" do
+  new_time = Time.new(2021, 10, 16, 0, 0)
+
+  around { |block| Timecop.freeze(new_time, &block) }
```
s-palmer, line 4: Hadn't heard about Timecop, looks useful!

views/chitter/index.erb
```diff
@@ -0,0 +1,23 @@
+<h1>Chitter</h1>
+
+<% if @user %>
+  <h1>Welcome, <%= @user.email %></h1>
+  <% else %>
+  <a href="chitter/users/new">Sign up</a>
+<% end %>
+ <a href="chitter/sessions/new">Log in</a>
```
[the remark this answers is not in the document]
serenabertozzi: Thanks! I intended to do if, after implementing the log out, maybe I didn't write my test correctly, but at this point, I create a user and then I would be automatically logged in, so I was afraid my test wouldn't be able to click on the link! :P Good point!

spec/features/view_peeps_spec.rb
serenabertozzi: It was super useful to stub time in the tests!

PR #1851, opened 2021-10-17
Chitter-Challenge, Luke B submission

blue-luke:
See README2 for my reflections. Basic user stories met in full, but a few things missing in background. Much time spent debugging.


app.rb
```diff
@@ -0,0 +1,57 @@
...
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  get '/flow' do
+    # @user = User.find(session[:user_id])
```
TTurvey, line 20: Is the commented out code needed in the future? If not, it would be best to remove.

lib/message.rb
```diff
@@ -0,0 +1,29 @@
...
+    connection = PG.connect(dbname: 'chitter')
+    result = connection.exec('SELECT * FROM messages;')
+    result.reverse_each.map do |message|
+      Message.new(content: message['message'], id: message['id'], time: message['time'], 
+user_id: message['user_id'])
+    end
+  end
+
+  def self.create(content:, user_id: "Anonymous")
+    current_time = Time.new.to_s
```
TTurvey, line 24: Does 'Time.new' return the current time? I was under the impression that Time.now returned the current time and Time.new would create a new instance of the Time class which may not be the current time that the peep was posted?

views/flow/index.erb
```diff
@@ -0,0 +1,32 @@
...
+  Message content is:
+  <br>
+  <b> <%= message.content %> </b>
+  <br>
+
+  Message id is:
+  <%= message.id %>
+  <br>
+
+  Message time and date are:
```
TTurvey, line 23: I think you've done well here by splitting out the time/date and message parts of the peep and then bringing them together here in the view file. It makes for easy-to-read code which is always good!

spec/features/reverse_chrono_flow_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature ChitterController do
+  scenario 'post a message to Chitter and see it posted' do
+    visit '/flow/new'
```
TTurvey, line 3: Perhaps a 'web_helper.rb' file could assists with the following lines of test code...
```
visit '/flow/new'
fill_in('message', with: 'Skynet has become self-aware')
click_button('Submit')
```
Usually the 'helper' files would be used for repetitive code and you've only used these lines twice as far as I can see so perhaps at this stage it's not necessary. If you plan to work on this challenge any further then maybe it would become more relevant.

app.rb
blue-luke: Thanks for pointing this out - hashed out code needs to be deleted.

spec/features/reverse_chrono_flow_spec.rb
blue-luke: Thanks - yes, some helper files and additional tests would be beneficial

PR #1852, opened 2021-10-18
Chitter Challenge - Conor Butler

ConorButler:
I was going to try and use an ORM but ran out of time and got a bit carried away implementing features. As a result, I have a lot of very busy methods and tests that I'm not sure are perfect. I enjoyed having a go and figuring everything out, but I'd appreciate any feedback on how I could have done things better, particularly authentication.


Rakefile
```diff
@@ -0,0 +1,76 @@
...
+    users << user[0]['id']
+  end
+
+  DUMMY_POSTS.each do |text|
+    connection.exec(
+      "INSERT INTO peeps(text, time, author) VALUES('#{text}',
+      '#{rand((Time.now.utc - 10_000_000)..Time.now.utc)}', #{users.sample});"
+    )
+  end
+
```
bernardleanse, line 58: don't quite understand this bit, maybe you can explain

app.rb
```diff
@@ -0,0 +1,124 @@
...
+        redirect('/')
+      end
+      Peep.create(text: params[:peep_text], author: user.id)
+      session[:peep] = nil
+    else
+      flash[:notice] = "You are not logged in"
+    end
+    redirect('/')
+  end
+
```
bernardleanse, line 98: This logic could be refactored into the model

PR #1856, opened 2021-10-18
Chitter - Jacob

jtc27:


lib/user.rb
```diff
@@ -0,0 +1,43 @@
+require 'bcrypt'
+require_relative './database_connection'
+
+class User
+  def self.create(email:, password:)
+    encrypted_password = BCrypt::Password.create(password)
```
Matt-Warnock, line 6: Nice work! 
I may have put this encrypting in a private method.

lib/user.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    result = DatabaseConnection.query(
+      "INSERT INTO users (email, password) VALUES($1, $2) RETURNING id, email;", [email, password]
+    )
+    User.new(
+      id: result[0]['id'],
+      email: result[0]['email'],
+    )
```
Matt-Warnock, line 14: and I would have also to be put into a private methed too.

lib/user.rb
```diff
@@ -0,0 +1,43 @@
...
+    result = DatabaseConnection.query(
+      "SELECT * FROM users WHERE id = $1", [id]
+    )
+    User.new(
+      id: result[0]['id'],
+      email: result[0]['email'],
+    )
+  end
+
+  def self.authenticate(email:, password:)
```
Matt-Warnock, line 28: The behaver of this method is very intresting.
I think your problem solving skills shine here! 

I would have gone about this in a different way maybe, however I don't think I fully understand the design yet.
Maybe, having it return information saying weather or not the user has an account. And then giving an option if they would like to sign up.

spec/feature/user_spec.rb
```diff
@@ -0,0 +1,57 @@
+require 'user'
+require_relative 'database_helpers'
+
+describe User do
+  describe '.create' do
+    it 'creates a new user' do
```
Matt-Warnock, line 6: Well written tests over all. 
I'm unsure why this file is in the feature folder.

app.rb
```diff
@@ -0,0 +1,28 @@
+require 'sinatra/base'
+require 'sinatra/flash'
+
+require_relative './lib/user'
+require_relative './database_connection_setup'
```
Matt-Warnock, line 5: Nice a clear easy to read. Good work!

PR #1863, opened 2021-11-14
Ryan Grimes Chitter PR

RPGrimes:


spec/features/add_peep_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'pg'
+
+feature 'add new peep' do
+  scenario 'user can add new peeps to the page' do
+    connection = PG.connect(dbname: 'chitter_data_test')
+
```
JacobCH93, line 6: Good use of testing database, allows you to make almost a 'double' to test your page.

views/index.erb
```diff
@@ -0,0 +1,28 @@
...
+<body>
+<h1> Chitter! </h1>
+<h5> Pronounced like CHampagne or CHef or maCHine </h5>
+
+<p> Sign up below to enter Chitter! </p>
+
+<form action='/add-user' method='post'>
+  <label>Your username:</label>
+  <input type='text' name='username' placeholder='Your Username e.g. Memelord3000'>
+  <label>Your password:</label>
```
JacobCH93, line 22: Clean and DRY HTML here. Well done!

views/peeps.erb
```diff
@@ -0,0 +1,47 @@
+<html>
+<head>
+<style>
+body {background-color: midnightblue;
+      font-family: sans-serif;
```
JacobCH93, line 5: I like that you went into the aesthetics of the page, shows extra effort

Gemfile
```diff
@@ -6,8 +6,19 @@ group :test do
...
+  gem 'capybara'
+  gem 'sinatra'
+  gem 'sinatra-reloader'
+  gem 'sinatra-contrib'
+  gem 'rack'
+  gem "thin"
+  gem "puma"
+  gem 'reel'
+  gem 'http'
+  gem 'webrick'
```
JacobCH93, line 18: Many GEMs, would just question and maybe think if they are all necessary

JacobCH93 (commented):
Really good code! And without even seeing the page I can tell it would be clear and well presented. Good use of tests to ensure everything is working correctly.

Gemfile
RPGrimes: I agree I probably don't require all of these but I had many gem issues in the Bookmarks challenge this week so added more to cover all bases. In future, I need to make more of a concerted effort to only add relevant gems in future.

PR #1871, opened 2021-11-15
weekend challenge week4

Andybird88:
Weekend Challenge week4. Completed the first 4 user stories and then ran out of time. Would have liked more time to spend making look a lot better than it does.


blessingaliu (commented):
Excellent work done, the demo of the website showed that all the functionality implemented worked very well, an improvement would be to practice writing tests before coding :)

PR #1872, opened 2021-11-15
Ginny Chitter challenge

ginnyamazed:
Read me file NOT updated.
1st user story not implemented
I created my full database tables initially, including having primary and foreign keys, which were not nullable.  This caused problems when I went to implement the test environment.  I had not yet setup a method to pull chit_user IDs and automatically load that into a PEEP, which meant in order to have the first functionality of seeing a peep possible I had to have a chitter user table that was not being cleared because I needed the chitter user ID to stay the same.  This worked okay until I started to implement chitter user functionality - in which I then had to consistently manually clear rows from the table so that I didn't receive duplicate key errors.  
I started to break working functionality when I tried to send the create user model data to the controller - I currently have that stuff commented out because I am pulling incorrect variables.
Additionally, because of having the test database reset the peeps table, I properly had working and displaying on the localhost side started to show empty because I started to receive a failing Peep test due to how the spec was written (I could not get a passing test once I started to have it initialise a new instance of a Peep because I couldn't get the spec expect statement to match what the output was).  I initially ignored this as the code was actually working and the webpage was working (when it was pointing at the production database with info pre-loaded) - but once webpage went to test database, the failing rspec did not enter the test data to have info show on this page.
If I had more time, I would 
1) try to get the latter working first - i.e. get the peep_spec.rb working & passing tests.  
2) update readme
3) properly send model data to the controller
4) have sign_in page check for a user against the database (currently a user can sign in, but it is not being checked against database)
5) add feature to create new peep


lib/user.rb
```diff
@@ -0,0 +1,28 @@
...
+  def initialize(name, user_name, email, password, id)
+    @name = name
+    @user_name = user_name
+    @email = email
+    @password = password
+    @id = id
+  end
+
+  def self.all
+    connection = PG.connect(dbname: "chitter#{'_test' if ENV['ENVIRONMENT'] == 'test'}")
```
sigii1992, line 15: Nice job implementing an if/else statement on one line

lib/user.rb
```diff
@@ -0,0 +1,28 @@
+require 'pg'
+
```
sigii1992, line 2: Good naming. Easy to understand what you are talking about

spec/features/confirm_sign_up_spec.rb
```diff
@@ -0,0 +1,14 @@
+feature 'user can sign_up for chitter account' do
```
sigii1992, line 1: Great test, very well organized

spec/spec_helper.rb
```diff
@@ -1,5 +1,17 @@
 require 'simplecov'
 require 'simplecov-console'
+require 'pg'
```
sigii1992, line 3: Gemfiles can be split into environments. Ensure that all test related gems are in test group, e.g. capybara etc.

views/view_peeps.erb
```diff
@@ -0,0 +1,17 @@
+<h2>Here are the latest Peeps!</h2>
+
```
sigii1992, line 2: Easy to read and understand. Very well formatted

PR #1873, opened 2021-11-15
chitter-challenge Ben M

bmargerison:
**user stories**

As a Maker

As a maker

As a Maker
So that I can better appreciate the context of a peep
I want to see the time at which it was made

As a Maker
I want to sign up for Chitter

As a Maker
So that only I can post messages on Chitter as me
I want to log in to Chitter

**use**

The current user on loading the app is a user object with the name 'anonymous', but this fake user is not in the database and so there is an error message if you try to post as anonymous. 

You must first signup AND login, using the same username. You can then post and the peep will appear with your name.

**issues and improvements**

Unsure on:
* The logged in user was persisted across server calls using class methods set_user/current_user. Not come across this before so not sure if its the best way to implement.

* In order to access the author of a peep, a separate method (Peep.author) calls to the database, and uses the foreign ID in the peeps table to find the user in the users table. Perhaps this can be better implemented with a ruby object as opposed to SQL query?

With more time:
* Would have added edge cases such as user cannot post unless they are logged in.

* Tests could be improved with more expect statements to check the data that is being stored.

* Should separate the job of communicating with the database to a different class.


spec/features/post_peep_spec.rb
```diff
@@ -0,0 +1,8 @@
+feature 'posting peeps' do
+	scenario 'user can post a peep and see it appear in list of peeps' do
+		visit('/')
+		fill_in :peep_message, with: "Its a fine day to be a Maker"
+		click_button "Post"
+		expect(page).to have_content "Its a fine day to be a Maker"
+	end
+end
```
smoni19, line 15: Sensible test for page functionality

app/chitter_web.rb
```diff
@@ -0,0 +1,44 @@
...
+    erb :login
+  end
+
+  post '/login' do
+    User.assign_user(params[:username])
+    redirect '/'
+  end
+
+  run! if app_file == $0
+end
```
smoni19, line 44: Very stripped back controller 👍

README.md
```diff
@@ -1,3 +1,47 @@
...
+
+* In order to access the author of a peep, a separate method (Peep.author) calls to the database, and uses the foreign ID in the peeps table to find the user in the users table. Perhaps this can be better implemented with a ruby object as opposed to SQL query?
+
+With more time:
+* Would have added edge cases such as user cannot post unless they are logged in.
+
+* Tests could be improved with more expect statements to check the data that is being stored.
+
+* Should separate the job of communicating with the database to a different class.
+
```
smoni19, line 44: Appreciate the reflections

db/migrations/01_create_peeps_table.sql
```diff
@@ -0,0 +1,12 @@
...
+ALTER TABLE peeps ADD time TIMESTAMP;
+ALTER TABLE peeps ADD UserID int;
+ALTER TABLE peeps ADD FOREIGN KEY (UserID) REFERENCES users(id);
+
+CREATE DATABASE chitter_test;
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(60));
+ALTER TABLE peeps ADD time TIMESTAMP;
+ALTER TABLE peeps ADD UserID int;
+ALTER TABLE peeps ADD FOREIGN KEY (UserID) REFERENCES users(id);
+
```
smoni19, line 12: Could you not have changed the CREATE TABLE command to include the updates?

lib/peep.rb
```diff
@@ -0,0 +1,53 @@
...
+      connection = PG.connect(dbname: "chitter")
+    end
+    result = connection.exec_params(
+      "SELECT username FROM users WHERE id in 
+      (SELECT UserID from peeps WHERE UserID=$1)",
+      [peep_id])
+    result.first['username']
+  end
+  
+end
```
smoni19, line 53: Appreciate that you've linked the user and peeps tables

spec/peep_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  it "saves the time a peep was made" do
+    Timecop.freeze(Time.local(2021).to_formatted_s(:db))
+    user = double("user", :id => 1)
+    new_peep = Peep.create("Its a fine day to be a Maker", user.id)
+    expect(Peep.all.first.time).to eq Time.local(2021).to_formatted_s(:db)
+  end
+
+end
```
smoni19, line 20: I like that you've used the Timecop gem to isolate your test

PR #1885, opened 2021-12-12
W4 weekend challenge

tc1316:
Wasnt clear if or how to setup database tables in travis.yml


app.rb
```diff
@@ -0,0 +1,72 @@
...
+
+  get '/' do
+    if logged_in?
+      redirect '/chitter'
+    else
+      erb :prompt
+    end
+  end
+
+  post '/signup' do
```
sherwingp, line 28: Your controller is quite fat, try to refactor any logic that isn't about the response to the model where possible

PR #1887, opened 2021-12-12
Chitter Challenge

EviePalaiochorinou:
Only managed to -barely- implement the first user story (not enough time!).
My next step would be to improve the peep posting method by straight up inserting the new peep into the database and then showing all peeps from the database in /my-peeps.
Test coverage so far is about 98%.


app.rb
```diff
@@ -0,0 +1,28 @@
...
+    Peep.new(message: params[:message])
+    redirect '/my-peeps'
+  end
+
+  get '/my-peeps' do
+    erb :my_peeps
+  end
+  
+  run! if app_file == $0
+end
```
FrancescoCollotto, line 28: It might be beneficial to use RESTful routes convention for CRUD route names

spec/features/add_peep_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+        expect(page.has_button?('Post')).to be_truthy
+    end
+
+    scenario 'posts a peep' do
+        fill_in("message", with: "Hello! This is my first peep!")
+        click_button("Post")
+
+        expect(page).to have_content("Hello! This is my first peep!")      
+    end
+end
```
FrancescoCollotto, line 21: Good detailed feature spec test!

FrancescoCollotto (commented):
Good setup and Peep class implementation

PR #1888, opened 2021-12-12
chitter challenge

s-dousse:


README.md
```diff
@@ -139,4 +139,4 @@ let's see how far I can go...
 
 ## Domain Model
 
-![Domain Model] (insert screenshot from the miro board)
+![Domain Model] (https://github.com/s-dousse/chitter-challenge/blob/main/screenshots/Screenshot%202021-12-10%20at%2023.47.31.png)
```
giorgigutsaevi: Well done here! This is very helpful for whoever is checking the project.

lib/peep.rb
```diff
@@ -1,13 +1,39 @@
...
     rs = DatabaseConnection.query("SELECT * FROM peeps ORDER BY id DESC;")
 
-    rs.map { |peep| peep['content'] }
+    array_of_peeps = []
+
+    rs.each do |peep_data|  # rs (result object) is NOT A REAL PEEP instance just data entries. we grab these from the database with each
+      peep = Peep.new(                                    # use the data entries (rs) to create a Peep
+        id: peep_data['id'],
+        content: peep_data['content'],
+        created_at: Time.parse(peep_data['created_at'])  # Takes a string representation of a Time and attempts to parse it
```
giorgigutsaevi: Very good use of Time module. I had no idea .parse automatically gets the current timestamp. well done!

views/homepage.erb
```diff
@@ -4,12 +4,16 @@
   <link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
```
giorgigutsaevi, line 4: I think linking the .css file would work even without including it in the url()? So if we write href="/styles.css", it technically should work.

lib/peep.rb
```diff
@@ -1,25 +1,13 @@
...
   def self.create(content:)
-    if ENV['ENVIRONMENT'] == 'test'
-      con = PG.connect(dbname: 'chitter_app_test')
-    else
-      con = PG.connect(dbname: 'chitter_app')
-    end
-
-    rs = con.exec(
-      "INSERT INTO peeps (content) VALUES ('#{content}');")
+    rs = DatabaseConnection.query("INSERT INTO peeps (content) VALUES ('#{content}');")
```
giorgigutsaevi: Because you are using the exec_params with your .query, should not there be a second argument, with an array of parameters? Not sure if I'm right though.

lib/peep.rb
s-dousse: Thank's! It works well. but didn't manage to feature test this properly so far

views/homepage.erb
s-dousse: Yes, it makes sense, and it works perfectly, thank you!

lib/peep.rb
s-dousse: Good question!
I believe I set up the DB connection with this method: self.query(sql_string, params = [ ])
so the second parameter is set to default? I wonder 😕
I will investigate this 🙈

PR #1892, opened 2021-12-13
week 4 Chitter challenge 

sabrinalord:
Completed user stories:
I want to post a message (peep) to chitter  ✅
I want to see all peeps in reverse chronological order ✅ 
I want to see the time at which it was made ✅ 

Database:
I have one table called Peeps, which is populated on initialization of the Peep class with a timestamp and the message body. I planned to create a second table called Users, to store the userid and the username. 
To link the two tables, I would then need to add the userid from the User's table as a Foreign key in my Peeps table.


app.rb
```diff
@@ -0,0 +1,15 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
+class Chitter < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do 
+  'Hello World'
```
WaqasIkram: I have seen many ppl do erb: index here to maintain separation of responsibilities.

README.md
```diff
@@ -70,6 +73,11 @@ CREATE DATABASE chitter_test;
 CREATE TABLE peeps (peep_id SERIAL PRIMARY KEY, created_at TIMESTAMP, message_body VARCHAR(60), user_id VARCHAR(60));
 
 
+### Current Progress
+
+# ![Site screenshot](https://github.com/sabrinalord/chitter-challenge/blob/main/screenshot/chitter_screenshot.png)
+
+
```
WaqasIkram, line 80: I saw someone doing a separate MYREADME.md, in which they explained how many classes and methods they used. They also used instructions to run the app at the end. 
I am also learing how to do such MYREADME.md with a clear app map.

spec/features/viewing_peeps_spec.rb
```diff
@@ -1,15 +1,17 @@
...
+    Peep.create(peep_body: 'Example peep posted at 3pm', created_at: @fake_timestamp)
+    Peep.create(peep_body: 'Example peep posted at 2pm', created_at: @fake_timestamp)
+
 
     visit '/home'
 
     expect(page).to have_content "Example peep posted at 3pm"
     expect(page).to have_content "Example peep posted at 2pm"
-    expect(page).to have_content "Example peep posted at 1pm"
+
```
WaqasIkram, line 15: nice and detailed feature test.

spec/spec_helper.rb
```diff
@@ -17,7 +19,12 @@
...
 
+
 RSpec.configure do |config|
+  config.before(:each) do
+    setup_test_database
+  end 
+
   config.after(:suite) do
     puts
     puts "\e[33mHave you considered running rubocop? It will help you improve your code!\e[0m"
```
WaqasIkram, line 28: Overall, nice effort but as I can understand its work in progress :) I would try to keep Views and Controllar separate to make it easier to follow.

PR #1893, opened 2021-12-13
site functionality works, need to work on DB

bryn9696:


app.rb
```diff
@@ -0,0 +1,32 @@
...
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  get '/' do
+    erb :index
+  end
+
+  post '/addpeep' do
```
TomasGarciaDev, line 17: you could use RESTful for adding new peeps '/peep/new' or '/peep/add'

views/index.erb
```diff
@@ -0,0 +1,8 @@
+<h1> CHITTER </h1>
```
TomasGarciaDev, line 1: Great that you first sorted the login part of it. Great approach!

lib/user.rb
```diff
@@ -0,0 +1,8 @@
+class User
```
TomasGarciaDev, line 1: Very nice and simple Model's (in a very good way)

TomasGarciaDev (commented):
Great approach to the 'app', I like that is very 'simple' in every good way. Easy to understand and run

PR #1894, opened 2021-12-13
Weekend challenge - Chitter

chris-clement:


app.rb
```diff
@@ -0,0 +1,50 @@
...
+    erb :confirm_sign_up
+  end
+
+  get '/log_in' do
+    erb :log_in
+  end
+
+  post '/confirm_log_in' do
+    if SignUp.log_in_verified?(email: params[:email], password: params[:password])
+      @user_info = SignUp.all.reverse.first
```
leoht, line 33: What you want to retrieve here is actually the user info of the *user who just signed in*, so you have more than one solution:
 * change `SignUp.log_in_verified?` so it returns the `SignUp` object corresponding to this account instead — the condition then changes to verify it is not `nil`
 * add a new method to `SignUp` to retrieve an account by email, so you can use it in the controller:
 ```ruby
@user_info = SignUp.get_by_email(params[:email])
```

If you do this, you might also want to save the user ID into the `session` object so you keep a trace of who's logged in across requests (by storing this session ID, you can then retrieve the user profile in further HTTP requests made to the app)

File: app.rb [hunk 2]
```diff
@@ -0,0 +1,50 @@
...
+    end
+  end
+
+  get '/user_page/:id' do
+    @peeps = Peep.all
+    @user_info = SignUp.all.reverse.first
+    erb :user_page
+  end
+
+  post '/:id/:user_name/:handle/peep' do
```
leoht, line 46: For POST requests (e.g posting a new peep), It's usually better to not pass the parameters in the path, but in the request body:
```ruby
post '/peep' do
   handle = params[:handle]
   user_name = @user_info.user_name
   # etc
end
```

This means you'd also change the HTML form to create a peep to have an action set to `/peep` and have HTML inputs for each value you need to send

spec/features/logging_in_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+    example_user = SignUp.create(email: 'example1@fake_email.com', password: 'Password123', user_name: 'Joe Biden', handle: 'sleepy_joe')
+    visit '/'
+    click_on 'Log In'
+    fill_in 'email', with: 'example1@fake_email.com'
+    fill_in 'password', with: 'Password123'
+    click_on 'Submit'
+    expect(page).to have_content 'Hi Joe Biden, thanks for logging in'
+  end
+  scenario 'User logs in with incorrect info' do 
+    example_user = SignUp.create(email: 'example1@fake_email.com', password: 'Password123', user_name: 'Joe Biden', handle: 'sleepy_joe')
```
leoht, line 12: Nice to use the model class to do the required setup for this test :)

app.rb
```diff
@@ -0,0 +1,50 @@
...
+    Peep.create(text: params[:text], user_name: params[:user_name], handle: params[:handle], time_posted: Time.new)
+    redirect '/user_page/id:'
```
leoht, line 48: Here the redirect URL shouldn't contain a named parameter, but the final value you want to use for this path - e.g `redirect /user_page/123`, so you'll need to interpolate the current user ID in the redirect path

db/migrations/create_database_and_tables.sql
```diff
@@ -0,0 +1,12 @@
+CREATE DATABASE chitter_manager;
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, text VARCHAR(140), handle VARCHAR(50), user_name VARCHAR(60));
+ALTER TABLE peeps ADD time_posted TIMESTAMP;
```
leoht, line 3: This those SQL statements are part of the same migration, they'll always be executed together — so it's fine to move the "ALTER TABLE" new column to be part of the initial "CREATE TABLE" so the table is created with the `time_posted` column initially too :)

leoht (commented):
Great work! I left a few comments mainly on minor things to improve the way you're checking & managing the user info when logging in, and when posting a new peep with a POST request — other than this, the code is quite readable and clean, the tests are well-written and quite thorough, so well done :) 
Let me know if there's anything unclear on the feedback, or if you'd like to discuss it further at some point

PR #1895, opened 2021-12-13
Chitter - weekend challenge

FrancescoCollotto:


lib/database_connection.rb
```diff
@@ -0,0 +1,11 @@
+require 'pg'
+
```
EviePalaiochorinou, line 2: Oh that's your 3rd class! You forgot to mention it in the read me and I was wondering. Very nice. Separation of concerns.

spec/features/login_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+
+    expect(page).to have_content 'Welcome, Ruby'
+  end
+   
+  scenario "a user can not sign in if not already registered or wrong email" do
+    visit '/users/login'
+    fill_in(:email, with: 'notregistered@email.com')
+    fill_in(:password, with: 'password')
+    click_button('Sign in')
+
```
EviePalaiochorinou, line 18: Couldn't you have raised an error here in the test instead? Not sure if possible, just asking.

views/layout.erb
```diff
@@ -0,0 +1,69 @@
+<!DOCTYPE html>
```
EviePalaiochorinou, line 1: Your code is very clean and easy to read! I have nothing to add since I did not get that far. Looks perfect to me!

PR #1900, opened 2021-12-13
Chitter-Challenge 

WaqasIkram:


app.rb
```diff
@@ -0,0 +1,91 @@
...
+class ChitterChat < Sinatra::Base
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  enable :sessions, :method_override
+  register Sinatra::Flash
+
+  get '/' do
+    erb :"sessions/new"
```
sabrinalord, line 21: Nice use of REST naming convention

app.rb
```diff
@@ -0,0 +1,91 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require 'sinatra/flash'
```
sabrinalord, line 3: I cloned and did bundle install but when I run rackup I am getting an error that says 'cannot load such file -- sinatra/flash (LoadError)'. I think sinatra-flash may be missing from the gem file

app.rb
```diff
@@ -0,0 +1,91 @@
...
+require './lib/chitter'
+require './lib/user'
+require 'date'
+require 'setup_test_database.rb'
```
sabrinalord, line 7: I had a load error on 'setup_test_database.rb', I don't think you need to require setup_test_db here as it can be run using the before in spec_helper config.

app.rb
```diff
@@ -0,0 +1,91 @@
...
+  get '/' do
+    erb :"sessions/new"
+  end
+
+  post '/sessions' do
+    if User.find(params[:username], params[:password]) == false
+      flash[:notice] = "Please submit a valid username or password."
+      redirect('/')
+    else
+      user = User.authenticate(username: params[:username], password: params[:password])
```
sabrinalord, line 29: nice use of authenticate

db/01_create_tweets_table.sql
```diff
@@ -0,0 +1,6 @@
+CREATE TABLE tweets(
```
sabrinalord, line 1: I would add some database setup info maybe here on in the README - to avoid people calling the database the wrong thing

lib/chitter.rb
```diff
@@ -0,0 +1,41 @@
+require 'date'
+require 'pg'
+require './lib/database_connection.rb'
+
+class Chitter
```
sabrinalord, line 5: I think Chitter could have a different name, I'm not sure what object Chitter is representing from the user story. Maybe a Tweet or a Peep object would be more readable?

lib/chitter.rb
```diff
@@ -0,0 +1,41 @@
...
+
+  def self.all(id:)
+    connect = PG.connect(dbname: 'chitter_test')
+    chitter = connect.exec("SELECT * FROM tweets WHERE user_id = #{id} ORDER BY time DESC;")
+    chitter.map do |tweet|
+      Chitter.new(tweet: tweet['tweet'], time: tweet['time'], user_id: tweet['user_id'])
+    end
+  end
+
+  def self.view_all
```
sabrinalord, line 28: nice clear method name here

views/tweets/index.erb
```diff
@@ -0,0 +1,170 @@
...
+              <span><p><%= @username %></p> <br>
+              <%= tweet.tweet %></span> <span class ='time'><%= tweet.time %></span>
+            </div>
+          <% end %>
+        <% end %>
+      </div>
+  </main>
+</body>
+
+<style>
```
sabrinalord, line 57: very beautiful styling!

sabrinalord (commented):
A fully functioning Chitter/Twitter app! This is really great, the CSS looks very clean, and all the functionality is working. I would add a bit more setup info on the README as I was unsure what to name the database at first. Other than that I think it's a really great submission and the data flows well accross both tables. Good job!

PR #1901, opened 2021-12-13
Chitter aapp

Fadaka:


app.rb
```diff
@@ -0,0 +1,20 @@
...
+require 'sinatra/reloader'
+require './lib/tweet'
+
+class Chitter < Sinatra::Base
+    configure :development do
+        register Sinatra::Reloader
+    end
+
+    get '/' do
+        'tweet tweet'
```
chris-clement, line 11: homepage at the moment is quite bare, looking at your ruby code I know I should go to /tweets but a link from the homepage would be helpful

db/migrations/01_create _tweet_form.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE tweet_timeline(id SERIAL PRIMARY KEY, tweet VARCHAR(60))
```
chris-clement, line 1: Good use of instructions on how to set up table. You should also include how to set up the database so I know it should be called 'tweet_form' and tweet_form_test'

lib/tweet.rb
```diff
@@ -0,0 +1,14 @@
...
+
+class Tweet
+    def self.all
+        if ENV['ENVIRONMENT'] == 'test'
+            connection = PG.connect(dbname: 'tweet_form_test')
+        else
+            connection = PG.connect(dbname: 'tweet_form')
+        end
+        result = connection.exec('SELECT * FROM tweet_timeline')
+        result.map {|tweet_timeline| tweet_timeline['tweet']}
```
chris-clement, line 11: nice use of map and relevant names given, makes the code easy to follow

spec/tweet_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+            all_tweets = Tweet.all
+
+            expect(all_tweets).to include('hello I am on twitter 1')
+            expect(all_tweets).to include('hello I am on twitter 2')
+            expect(all_tweets).to include('hello I am on twitter 3')
+
+
+        end
+    end
+end
```
chris-clement, line 21: good structure to your tests, I can understand what is being tested here from your use of describe, breaking it down by method and clear it statements

Gemfile.lock
```diff
@@ -36,6 +64,7 @@ GEM
     rubocop-ast (1.11.0)
       parser (>= 3.0.1.1)
     ruby-progressbar (1.11.0)
+    ruby2_keywords (0.0.5)
```
chris-clement, line 67: lots of Gems being used here, what are they all for? Interested to know where they are used in your program and what they do

chris-clement (commented):
Nice work on the Chitter challenge. I think it would benefit from having a homepage which maybe redirects you to the /tweets page or have a link which takes you there. See my specific comments for details

PR #1903, opened 2021-12-14
Incomplete Chitter Challenge

christopherxwebb:
There is some obvious work to do here but I am opening a pull request in order to get some feedback on the problems that persist.

Currently, I am having trouble with the Chitter feed: a host of hashes are being displayed. I would like some advice on how to fix this and what's going wrong under the hood more generally.

I purposely wanted to keep everything on the same page (like the feed on Twitter). I realise it would be a lot simpler to spread everything out over a few different pages/views but chose not to do this.

Any help would be greatly appreciated.


app.rb
```diff
@@ -0,0 +1,24 @@
...
+require './lib/peep'
+
+class Chitter < Sinatra::Base
+
+  configure :development do
+    register Sinatra::Reloader 
+  end
+
+  get '/' do
+    @peep = Peep.all
```
eoinmakers, line 12: Will ask you a couple of questions based on your question around why your logic is not communicating with your view / controller! 

First up, do you get a result from this method call? What do you see when you write p @peep underneath this line, and run your code with rpec?

app.rb
```diff
@@ -0,0 +1,24 @@
...
+    erb :homepage
+  end
+
+  post '/' do
+    @newpeep= Peep.create(peep: params[:peep])
```
eoinmakers, line 17: Do you get a successful result from this method call, to create a new peep? 

If not, if you were to add a line above this to print params (**p params**), do you see a peep message stored in params[:peep]?

views/homepage.erb
```diff
@@ -0,0 +1,24 @@
...
+
+<h1> Peep Feed </h1>
+<ul>
+<%@newpeep%>
+  <% @peep.each do |peep| %>
+   <li>
+      <%=peep.peep%>
+   </li>
+   <% end%>
+</ul>
```
eoinmakers, line 23: Do you see this displayed on your homepage?

PR #1913, opened 2022-01-17
Chitter Challenge- Mat

mikitaro-27:


lib/peep.rb
```diff
@@ -0,0 +1,42 @@
+require 'pg'
+require 'time'
+
+class Peep
+  def self.add(content:)
+    return false if is_empty?(content)
```
eoinmakers, line 6: Great work.

lib/peep.rb
```diff
@@ -0,0 +1,42 @@
...
+    prs_data = database_connect.exec_params("INSERT INTO peeps (content, created_at) VALUES ($1, now()) RETURNING id, content, created_at;", [content])
+    Peep.new(id: prs_data[0]['id'], content: prs_data[0]['content'], created_at: Time.parse(prs_data[0]['created_at']))
+  end
+
+  def self.get_all
+    output = database_connect.query("SELECT * FROM peeps;")
+    peeps = output.map { |peep_data| Peep.new(id: peep_data['id'], content: peep_data['content'], created_at: Time.parse(peep_data['created_at']))}
```
eoinmakers, line 13: Concisely written method.

spec/features/viewing_peeps_spec.rb
```diff
@@ -0,0 +1,29 @@
...
+    visit '/'
+
+    peeps = page.all('.peep')
+   
+    expect(peeps[0]).to have_content "Third peep!"
+    expect(peeps[1]).to have_content "Second peep!"
+    expect(peeps[2]).to have_content "First peep!"
+  end
+  
+  #Should I be stubbing time values to see whether?
```
eoinmakers, line 15: You can use Timecop to freeze times - we try to stub non-deterministic values in our tests including time. Timecop's not the only solution, as you could also create stubs instead - although sometimes this requires a bit more of a refactor.

eoinmakers (commented):
This is a great Chitter implementation, nice work. 

In particular, I thought each class was clearly defined and had one sole responsibility each. Your ReadMe was descriptive and shows the app in use too. I made a small suggestion around stubbing Time in your (overall comprehensive) tests, which would be one good improvement to make. Aside from that, this is great. Nice work!

PR #1921, opened 2022-01-17
chitter-challenge

cmkap:


spec/features/viewing_all_cheet_spec.rb
```diff
@@ -30,6 +30,6 @@
 
   scenario 'displays in reverse chronological order' do
     visit('/')
-    expect(page).to have_content("Latest cheets:\n\"The last word\" cheeted by Duck_Tape, @")
+    expect(page).to have_content("Latest cheets:\n\"The last word\" cheeted by Duck_Tape, at")
```
izaowl, line 33: I would keep it simple. Test it just for one sentence on the page. I believe you need to create it first using the class which manages all logic for that message.

PR #1929, opened 2022-02-20
Jan 22 cohort, Jack

winters145:


db/01_create_peeps_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE bookmarks(id SERIAL PRIMARY KEY, peep VARCHAR(60), post_time TIME);
```
mmguinness, line 1: Name of table could be peeps rather than bookmarks for clarity in your code.

README.md
```diff
@@ -1,23 +1,31 @@
...
-As usual please start by forking this repo.
+In order to user this web-app please follow these instructions:
 
-We are going to write a small Twitter clone that will allow the users to post messages to a public stream.
+* Clone this repo
+* Run 'bundle' to install the required gems
+* Connect to psql
+* Create the database using the psql command CREATE DATABASE chitter;
+* Connect to the database using the pqsl command \c chitter;
+* Run the query saved in the file db/01_create_peeps_table.sql
```
mmguinness, line 12: This file should also note about setting up the test database. I see that table is called peeps, not bookmark.

spec/chitter_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it 'saves and returns the posted peep' do
+      peep = Peep.post("Hello, Chitter!")
+      expect(peep.content).to eq ("Hello, Chitter!")
+      expect(peep.post_time).to eq Time.now.strftime("%H:%M:%S")
+    end
+  end
+
+  describe '#list' do
+    it 'lists all the posted peeps' do
+
```
mmguinness, line 15: It would be a more robust test to add a second peep to test, as the test wording includes 'lists all'.

PR #1930, opened 2022-02-20
Francesco - First three user stories

now-what:
Completed the first three user stories.
Didn't get the chance to implement signup / login / logout.


app.rb
```diff
@@ -0,0 +1,34 @@
+require 'sinatra/base'
```
ShirleyMacfarlane, line 1: As a general comment: I found your code very well thought out and put together. It is very easy to both understand and follow.

ShirleyMacfarlane: I liked your use of docs and also external style sheets.

lib/peep.rb
```diff
@@ -0,0 +1,33 @@
...
+      Peep.new(message: peep['message'], time: peep['time'])
+    end
+  end
+
+  def self.write(text:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
```
JohnForster, line 29: It would be really nice if you could store your db name as an env variable, so you could do something along the lines of:
```ruby
db_name = ENV['DB_NAME']
connection = PG.connect(dbname: db_name)
```

spec/peep_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'peep'
+
+describe Peep do
+
+  describe '#all' do
```
JohnForster, line 5: Minor note, but `#all` indicates an _instance method_, but this is a _class method_, which would be `::all`

JohnForster (commented):
This all looks excellent, I can't see any real problems in the code whatsoever. If you struggled to implement the login/logout functionality, maybe we can discuss that separately.

PR #1931, opened 2022-02-20
Chitter Challenge Shirley 

ShirleyMacfarlane:


now-what (commented):
Great implementation - you've set up user registration off the bat and although I am not 100% sure on how it works (as I haven't worked on it yet), the cod is clean and easy to follow. Some of your tests don't work but we've started working on debugging them.

PR #1933, opened 2022-02-20
Chitter Challenge - Peter O

harmlessgoose:


README.md
```diff
@@ -1,3 +1,11 @@
+Setup
+--------
+1. Connect to psql
+2. Create two databases using the psql command CREATE DATABASE chitter; and CREATE DATABASE chitter_test;
+3. Connect to the database using the pqsl command \c chitter;
+4. Run the queries saved in /db/migrations in order
+
```
SarahM55, line 7: Nice job remembering to add this!

spec/spec_helper.rb
```diff
@@ -1,14 +1,32 @@
...
-  # SimpleCov::Formatter::HTMLFormatter
+  SimpleCov::Formatter::HTMLFormatter
 ])
 SimpleCov.start
 
 RSpec.configure do |config|
+  config.before(:each) do
+    setup_test_database
+  end
+
```
SarahM55, line 29: Cleaning the database between tests, this is great :)

app.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  get '/peeps/new' do
+    erb :"peeps/new"
+  end
+
+  post '/peeps' do
+    Peep.create(message: params[:message])
+    redirect '/peeps'
+  end
```
SarahM55, line 26: Nice use of RESTful routes :D

spec/database_helpers.rb
```diff
@@ -0,0 +1,7 @@
+require 'pg'
+
+def persisted_data(id:)
+  connection = PG.connect(dbname: 'chitter_test')
+  result = connection.query("SELECT * FROM peeps WHERE id = #{id};")
+  result.first
+end
```
SarahM55, line 7: Good use of a 'helper' to extract this code into its own file

SarahM55 (commented):
This is a really great attempt! Impressed that you managed to finish three of the user stories and work out how to add the time stamps :) Your code is really well set out and easy to follow, you've made sure that your feature tests are doing more than one thing instead of having a new scenario for each. One tiny change you could have made was to redirect your '/' route to '/peeps/new' so that the website would 'flow' without having to add each route on to the url, if that makes sense! Great work :D

PR #1934, opened 2022-02-20
Chitter Challenge - Claire Nelson

nelsonclaire:


lib/peep.rb
```diff
@@ -0,0 +1,53 @@
...
+      created_at: result[0]['created_at'],
+      user_id: peep['user_id']
+    )
+  end
+
+  def self.all
+    result = DatabaseConnection.query("SELECT * FROM peeps ORDER BY created_at DESC;")
+    list = result.map do |peep|
+      epoch = Time.parse(peep["created_at"]).to_i
+      date = Time.at(epoch).strftime("%e %b %Y at %H:%M")
```
Dash1704, line 43: Nice use of google to return the time the peeps were created

public/stylesheet.css
```diff
@@ -0,0 +1,152 @@
+* {
+  box-sizing: border-box;
+}
```
Dash1704, line 3: Lovely use of CSS to make it look pretty

PR #1937, opened 2022-02-20
Chitter displaying posts and times

Haroon2021:


CBuchan5 (commented):
Overall good looking code. Perhaps try and have a go at sprucing it up with HTML. Also don't forget to update the README, even if it's a simple 'create table' instruction.

spec/features/Add_chitter_peep_spec.rb
```diff
@@ -5,5 +5,6 @@
         fill_in('peep', with: 'test peep')
         click_button('Submit')
         expect(page).to have_content 'test peep'
+
```
CBuchan5, line 8: Good clear test for adding peeps, good job :)

PR #1947, opened 2022-02-21
Chitter Challenge for Chris Middleton

CTJMID:
Ran out of time, but a good start I think.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,16 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+
+
+class Chitter < Sinatra::Base 
+  configure :development do 
+    register Sinatra::Reloader 
+  end
+  
+
+  get '/' do
+    'Chitter'
+  end
+
+  run! if app_file == $0
+end
--- a/Gemfile
+++ b/Gemfile
@@ -2,7 +2,13 @@ source 'https://rubygems.org'
 
 ruby '3.0.2'
 
+gem 'sinatra'
+gem 'sinatra-contrib'
+gem 'webrick'
+
 group :test do
+  gem 'pg'
+  gem 'capybara'
   gem 'rspec'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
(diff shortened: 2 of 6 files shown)


lib/peeps.rb
```diff
@@ -0,0 +1,36 @@
...
+      conn = PG.connect(dbname: 'chitter')
+    end
+    records = conn.exec("SELECT * FROM peeps")
+    records.map do |peep|
+      Peep.new(id: peep['id'], content: peep['content'], time: peep['time_stamp'])
+    end     
+  end
+
+  def self.create(content:)
+    if ENV['ENVIRONMENT'] == 'test'
```
AliCo-Hash, line 26: DRY. Can put this into a method to grab from.

app.rb
```diff
@@ -1,16 +1,21 @@
...
 
 class Chitter < Sinatra::Base 
   configure :development do 
     register Sinatra::Reloader 
   end
   
-
   get '/' do
-    'Chitter'
+    @peeps = Peep.all.reverse
```
AliCo-Hash, line 11: Cool way of having peeps in reverse chronological order.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,33 @@
+
+require 'peeps'
+
+describe Peep do
+  describe '#self.all' do
+    it 'returns all peeps' do
+      connection = PG.connect(dbname: 'chitter_test')
```
AliCo-Hash, line 7: Is there a point to this line if in our peep.rb we have an if statement to connect to the chitter_test db when rspec is run.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+
+      connection.exec("INSERT INTO peeps (content) VALUES('This is my first peep');")
```
AliCo-Hash, line 9: Can be refactored with the new create method.

app.rb
AliCo-Hash: Can add a feature test using .first to check order.

PR #1954, opened 2022-03-20
Chris' chitter challenge

clovellbsc:
creating twitter clone with activerecord, sinatra, capybara, rspec and html. Added ability to log in, log out, create posts associated with account. Unable to make posts or comment on posts without being signed in. Ability to comment on posts.

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,84 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/user'
+require './lib/post'
+require './lib/comment'
+require 'sinatra/flash'
+
+class Chitter < Sinatra::Base
+
+  enable :sessions
+  register Sinatra::Flash
+
+  get '/' do
+    @user = User.find_by_id(session[:user_id])
+    @post = Post.order(created_at: :desc).pluck(:peep, :created_at, :id, :user_id)
+    @comment = Comment.order(created_at: :asc)
+    erb :index
+  end
+
+  post '/add_post' do
+    new_post = Post.new(peep: params[:peep], user_id: session[:user_id])
+    new_post.save
+    redirect '/'
+  end
+
+  get '/sign-up' do
+    erb :sign_up
+  end
+
+  get '/add-comment/:id' do
+    @post_id = params[:id]
+    @post_text = Post.find_by_id(@post_id).peep
+    @post_user_id = Post.find_by_id(@post_id).user_id
+    @username = User.find_by_id(@post_user_id).username
+    erb :new_comment
+  end
+
+  post '/commenting/:id' do
+    Comment.create(comment: params[:comment], post_id: params[:id], user_id: session[:user_id])
+    redirect "/"
+  end
+
+  post '/signing-up' do
+    @new_user = User.create(
+      username: params[:username], 
+      email: params[:email],
+      first_name: params[:first_name],
+      last_name: params[:last_name],
+      password: params[:password]
+    )
+    @new_user.save
+    
+    if User.find_by_id(@new_user.id)
+      session[:user_id] = @new_user.id
+      redirect '/'
+    else 
+      flash[:notice] = 'Please ensure you have filled all the fields'
+      redirect '/sign-up'
+    end
+  end
+
+  post '/logging-in' do
+    @user = User.find_by(username: params[:log_in_username])
+
+    if @user.nil?
+      flash[:notice] = 'Please check your username'
+      redirect '/sign-up'
+    elsif @user.authenticate(params[:log_in_password])
+      session[:user_id] = @user.id
+      redirect '/'
+    else
+      flash[:notice] = 'Please check your password'
+      redirect '/sign-up'
+    end
+  end
+
+  post '/logging-out' do
+    session.clear
+    flash[:notice] = 'Successfully logged out'
+    redirect '/'
+  end
+
+  run! if app_file == $0
+end
--- a/lib/comment.rb
+++ b/lib/comment.rb
@@ -1,0 +1,27 @@
+require 'active_record'
+
+if ENV['ENVIRONMENT'] == 'test'
+  ActiveRecord::Base.establish_connection(
+    adapter: 'postgresql',
+    host: 'localhost',
+    username: ENV["USERNAME"],
+    password: ENV["SQL_PASSWORD"],
+    database: 'chitter_test'
+  )
+else 
+  ActiveRecord::Base.establish_connection(
+    adapter: 'postgresql',
+    host: 'localhost',
+    username: ENV["USERNAME"],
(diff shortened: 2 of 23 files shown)


app.rb
```diff
@@ -0,0 +1,84 @@
...
+require './lib/comment'
+require 'sinatra/flash'
+
+class Chitter < Sinatra::Base
+
+  enable :sessions
+  register Sinatra::Flash
+
+  get '/' do
+    @user = User.find_by_id(session[:user_id])
```
leoht, line 14: What happens if `session[:user_id]` is `nil`? Even if this in theory *should not* happen on this page, often unusual cases will come up and can break the application. There should probably be a guard condition handling this, and maybe redirecting to a login page

app.rb
```diff
@@ -0,0 +1,84 @@
...
+    redirect '/'
+  end
+
+  get '/sign-up' do
+    erb :sign_up
+  end
+
+  get '/add-comment/:id' do
+    @post_id = params[:id]
+    @post_text = Post.find_by_id(@post_id).peep
```
leoht, line 32: Here you're calling twice `Post.find_by_id` to get the post information (first the "peep" text, then the user ID) - although I think ActiveRecord optimises for this in some way (and even though this is hidden to you by the library), it means making two SQL requests to the database - which is not ideal, as your application needs to scale. It might be better to call it only once, and then assign the different instances variables if needed:
```ruby
@post = Post.find_by_id(@post_id)
@post_text = @post.peep
@post_user_id = @post.user_id
```

app.rb
```diff
@@ -0,0 +1,84 @@
...
+
+  post '/signing-up' do
+    @new_user = User.create(
+      username: params[:username], 
+      email: params[:email],
+      first_name: params[:first_name],
+      last_name: params[:last_name],
+      password: params[:password]
+    )
+    @new_user.save
```
leoht, line 51: The `.create` method already saves the new record into the DB, so you don't need to call `save` afterwards

lib/comment.rb
```diff
@@ -0,0 +1,25 @@
+require 'active_record'
+
+if ENV['ENVIRONMENT'] == 'test'
+  ActiveRecord::Base.establish_connection(
```
leoht, line 4: Rather than being in all model files, the connection setup should probably be factored in a file that would be required by the app when initialised — you can also use the method `ActiveRecord::Base.connected?` to make sure this is only run once

lib/post.rb
```diff
@@ -0,0 +1,29 @@
...
+    adapter: 'postgresql',
+    host: 'localhost',
+    username: ENV["USERNAME"],
+    password: ENV["SQL_PASSWORD"],
+    database: 'chitter'
+  )
+end
+
+class Post < ActiveRecord::Base
+  self.table_name = 'posts'
```
leoht, line 22: By default, AR checks for a table than is the lowercase and pluralised version of the class name, and a primary key column named `id`, which means you can probably remove these two lines of setup as they describe the default behaviour

app.rb
leoht: Also, since we're getting a list of posts - should the variable be named `@posts`?

views/index.erb
```diff
@@ -0,0 +1,63 @@
...
+        </label>
+        <input type="submit" value="Send Peep">
+      </form>
+      <br>
+      <br>
+    <% end %>
+      <h3> Your Chitter Feed </h3>
+    <ul>
+      <% @post.each do |peep| %>
+      <%time = peep[1].strftime("%d/%m/%Y %k:%M")%>
```
leoht, line 44: AR generates "on the fly" attribute accessors for the model class, depending on the column names it finds in the table — which means you can use `peep.id`, `peep.created_at`, `peep.peep`, etc, rather than accessing it like an array index, so it makes the code easier to read

app.rb
```diff
@@ -0,0 +1,84 @@
...
+
+class Chitter < Sinatra::Base
+
+  enable :sessions
+  register Sinatra::Flash
+
+  get '/' do
+    @user = User.find_by_id(session[:user_id])
+    @post = Post.order(created_at: :desc).pluck(:peep, :created_at, :id, :user_id)
+    @comment = Comment.order(created_at: :asc)
```
leoht, line 16: This one is a bit more advanced — you can optimise the query to the DB by "telling" AR you're going to need also the comments associated with each post. This way AR will perform a SQL JOIN to retrieve the comments information:
```ruby
@posts = Post.order(created_at: :desc).pluck(:peep, :created_at, :id, :user_id).includes(:comments)
```

And later, when you need to access and list the comments for a given post:
```ruby
post.comments.each do |comment|
  # ...
end
```

This has two advantages:
 * you don't need to filter manually the comments by their post ID
 * the comments are already loaded - otherwise AR would need to make an additional SQL query to SELECT the comments associated to each post, for every post we iterate on in the view. This is a very common pitfall of ActiveRecord (and ORMs in general) [called the N+1 problem](https://blog.appsignal.com/2018/04/24/active-record-performance-the-n-1-queries-antipattern.html), so it's good to learn about it early and recognise when it can happen and how it can be avoided

leoht (commented):
Great job! My feedback is mainly on how you've used ActiveRecord, since you asked specifically about this — there are a few ways you can simplify and improve its use, and some words about the "N+1" query optimisation when loading posts' comments on the index page — let me know if you have any further questions on that feedback, or on AR, and happy to discuss it :)

app.rb
clovellbsc: > What happens if `session[:user_id]` is `nil`? Even if this in theory _should not_ happen on this page, often unusual cases will come up and can break the application. There should probably be a guard condition handling this, and maybe redirecting to a login page

I have an if else statement in the homepage therefore if it is nil, the page will only display the posts and no ability to add posts or comments with a login button and sign up button that take you to a login page, therefore, it shouldn't be an issue. Or am I missing something?

app.rb
clovellbsc: Ah, that's silly of me

lib/comment.rb
clovellbsc: All done, now thank you. Again, this makes a lot of sense

lib/post.rb
clovellbsc: I had left these in as a precaution but you're right, I shouldn't be keeping unnecessary code

views/index.erb
clovellbsc: That's really good to know, I took ages looking for how to do it the way I did, good to know that I can sort this in a much more manageable way.

app.rb
clovellbsc: That's really interesting and amazing to know. Thank you! I have just read the article and it makes a lot of sense. Definitely one to keep an eye out for in the future.

PR #1967, opened 2022-03-21
Chitter - work in progress

Lorn888:


app.rb
```diff
@@ -0,0 +1,20 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/peep'
+
```
anushanesargi, line 4: Given it is Work In Progress, it is still a neat piece of code

spec/peep_spec.rb
```diff
@@ -0,0 +1,21 @@
+require 'peep'
+
+describe Peep do
+  describe '.all' do
+    it 'returns all peeps' do
+
+      connection = PG.connect(dbname: 'chitter_test')
+
+      # Add the test data
+      connection.exec("INSERT INTO peeps (peep) VALUES ('Peep test 1');")
```
anushanesargi, line 10: You would need a method inside the class peep to insert the methods.

PR #1969, opened 2022-04-18
Full sign up, log in and log out functionality, with some HTML and CSS adaptation

Josenewmano:


app.rb
```diff
@@ -0,0 +1,69 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require_relative './lib/peep'
+require_relative './lib/user'
+require_relative './lib/database_connection_setup'
+
+class Chitter < Sinatra::Base
```
almorcrette, line 7: Beautifully 'skinny'. Well done.

app.rb
```diff
@@ -0,0 +1,69 @@
...
+    user = User.find(session[:user_id])
+    Peep.create(content: params['content'], username: user.username, name: user.name)
+    redirect '/'
+  end
+
+  get '/sign_up' do
+    erb :sign_up
+  end
+
+  post '/add_user' do
```
almorcrette, line 29: Recommending reviewing RESTful routes as this doesn't follow REST conventions

app.rb
```diff
@@ -0,0 +1,69 @@
...
+    erb :index
+  end
+
+  post '/add_peeps' do
+    user = User.find(session[:user_id])
+    Peep.create(content: params['content'], username: user.username, name: user.name)
+    redirect '/'
+  end
+
+  get '/sign_up' do
```
almorcrette, line 25: Review RESTful routes as this doesn't follow convention

app.rb
```diff
@@ -0,0 +1,69 @@
...
+      session[:user_id] = user.id
+      redirect '/'
+    end
+  end
+
+  get '/log_in' do
+    erb :log_in
+  end
+
+  post '/new_session' do
```
almorcrette, line 45: Not RESTful

db/migrations/02_add_posted_time_to_messages.sql
```diff
@@ -0,0 +1,2 @@
+ALTER TABLE peeps 
+ADD COLUMN posted VARCHAR(60);
```
almorcrette, line 2: Would it be better for the database to store a timestamp rather than a varchar?

lib/peep.rb
```diff
@@ -0,0 +1,31 @@
...
+    result = DatabaseConnection.query("SELECT * FROM peeps;")
+    array_of_peeps = result.map do |peep| 
+      Peep.new(id: peep['id'], content: peep['content'], 
+               posted: peep['posted'], username: peep['username'], name: peep['name'])
+    end
+    array_of_peeps.reverse
+  end
+
+  def self.create(content:, username:, name:)
+    posted = Time.now.strftime("%I:%M %p, %d/%m/%Y")
```
almorcrette, line 24: Just a thought - where is it best to convert a timestamp to a string? You have done it when the peep is created so that what is in the database is a string. Perhaps it would maintain more information to keep the timestamp persisting in the database and the model, and only convert it to the string you want when taking it to the view (not sure if that would be in the controller or the view you would do that)...

views/index.erb
```diff
@@ -0,0 +1,93 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style>
```
almorcrette, line 4: Best practice would be to extract a style sheet for your CSS to keep CSS and HTML code separate. But I'm just trying to give you constructive feedback. You've already gone way beyond me...

views/log_in.erb
```diff
@@ -0,0 +1,43 @@
+<!DOCTYPE html>
+<html>
+<head>
+<style>
```
almorcrette, line 4: Extract CSS into separate stylesheet

Gemfile
```diff
@@ -2,10 +2,19 @@ source 'https://rubygems.org'
...
 ruby '3.0.2'
 
+gem 'sinatra'
+gem 'sinatra-contrib'
+gem 'webrick'
+gem 'pg'
+gem 'launchy'
+gem 'bcrypt'
+
 group :test do
```
almorcrette, line 12: Good to have the test suite grouped

spec/features/home_page_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+  scenario 'log in button' do
+    visit('/')
+    click_button('Log in here...')
+    expect(page).to have_content "Enter your username and password below:"
+  end
+  
+  scenario 'user can add a peep to the page after logging in' do
+    User.create(username: 'Sarahness', name: 'Sarah Smith', email: 'email@email.com', password: 'pwpwpwpw')
+    
+    visit('/log_in')
```
almorcrette, line 27: Could you use spec_helper methods more to make these tests read more easily, be shorter, avoid repetition?

app.rb
```diff
@@ -0,0 +1,69 @@
...
+    user = User.authenticate(username: params[:username], password: params[:password])
+    if user.nil?
+      redirect '/log_in_failure'
```
almorcrette, line 48: Not using Flash, could look into this.

almorcrette (commented):
Can confirm the following features implemented:

* Features
  * [x] Sign up
  * [x] Log in
  * [x] Log out
  * [x] Peeping
  * [x] Listing Peeps
  * [x] Display time of peep
  * [ ]  Email when tagged

* Bonus Features
  * [ ] Replying to peeps
  * [x] CSS styling
  
Planning
- [x] CRC cards

README
- [x] Database setup instructions including for test
- [x] Info about planning / technical approach
- [ ] Instructions for getting going (cloning, `bundle`, `rackup`, `rspec` for testing) - recommend including this to provide full instructions to new reviewing on using your app

ORM
- [x] Built own ORM
- [ ] Use of out-of-the-box ORM (e.g. ActiveRecord)

Test coverage
- [x] 100%!
- [x] feature tests and unit tests look and read differently. Feature tests are longer, more procedural and named more like doing something 
- [x] Feature tests in features folder
- [x] Feature tests interacting with entire stack 
- [x] Unit tests are not in features folder 
- [x] Spec helpers used for database
- [ ] Using spec_helpers to stay DRY on tests - could use this technique more

MVC
- [x] Skinny controller
- [x] SRP on models and their methods
- [ ] RESTful routes - controller not following REST. Recommend this
- [ ] Use of flash. Not using this. Could consider this
- [ ] Splitting routes into separate controller files. Not done this. See review / rubric for guidance.
- [x] Using semantic HTML elements to divs where possible
- [ ] splitting CSS into stylesheets to keep CSS and HTML separate
- [ ] view helpers
- [ ] nice to have: Rakefil has appropriate tasks??

config
- [x] test group in Gemfile
- [x] spec_helper configured with all dependencies
- [x] spec_helper connecting to testing framework
- [x] clean database for testing, creating data for individual tests, running test with that data

db/migrations/02_add_posted_time_to_messages.sql
Josenewmano: Yes, it would, and I did initially, but was gettting issues with my tests, so I changed it. Your idea of using the strfttime in the html would have prevented these issues, and made it easier to adapt the code later

views/index.erb
Josenewmano: I need to research how to do this! But it sounds like it would make the code DRYer

spec/features/home_page_spec.rb
Josenewmano: Yes, I should have done that, although some of the Sarahs are subtly different to each other

app.rb
Josenewmano: I made a decision not to use Flash, as I think it's going to die pretty soon

spec/features/home_page_spec.rb
Josenewmano: The Sarahs did turn out to be all the same, but trying to add it in after the wipe table method in the rspec before config didn't work - I remember now that I also had issues making the database_helper methods use the DatabaseConnection class methods. It seems that the connection to the database isn't actually yet made at that point, and only when the tests actually happen? Would you be able to shed any light?

almorcrette: Hey! An example of doing this is here (link below), used on the Battle pair challenge. You can find some guidance on it in the relevant section of the challenge.

I should have been clearer: the concept is to create a file called web_helper.rb inside spec/features which just has some methods which run the repetitive code. You require this file where you have used the repetitive code and replace the code with the declarative method names stored in the helper file to be DRYer. Having the file in spec is ok because as long as you don't include spec in the name, Rspec will not think it is a test itself.

https://github.com/almorcrette/Battle/blob/main/spec/features/web_helpers.rb

PR #1982, opened 2022-05-15
Chitter challenge

ErlantzR:
User stories completed: 3


db/migrations/01_create_peeps_table.sql
```diff
@@ -0,0 +1 @@
+CREATE TABLE peeps(id SERIAL PRIMARY KEY, message VARCHAR(140), time VARCHAR(60));
```
lukestorey95, line 1: Maybe consider having message  column as not null - Users shouldn't be able create a blank peep

lib/peep.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def self.post(message:)
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    time = Time.new.strftime('%a, %d %b %Y %H:%M:%S')
```
lukestorey95, line 33: PostgreSQL has the NOW() default value that sets the time in the database for you - might be worth considering so that you don't have to generate the time every time you want to create a peep

spec/peep_spec.rb
```diff
@@ -0,0 +1,39 @@
+require 'peep'
+
+describe Peep do
+
+  describe '#.show' do
```
lukestorey95, line 5: #show is for an instance method, .show is for a class method, so think this should just be .show

lib/peep.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  def self.show
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+    end
+
+    result = connection.exec("SELECT * FROM peeps ORDER BY id DESC;")
```
lukestorey95, line 20: I hadn't thought of sorting by ID, that's clever as ID is generated sequentially!

lukestorey95 (commented):
A really great start, and I think you've set yourself up well if you had time to work on the next few users stories. I've left a couple of comments around the formatting of the tests (me being fussy) and the structure of your database.

PR #1996, opened 2022-05-23
Jimmy Chitter Challenge

jimmy-lyons:
Basic functionality


lib/peeps.rb
```diff
@@ -0,0 +1,43 @@
...
+    )
+    Peeps.new(
+      peep_id: result[0]['id'],
+      time_stamp: result[0]['time_stamp'],
+      user_id: result[0]['user_id'],
+      peep: result[0]['peep']
+    )
+  end
+
+  def self.all
```
ErlantzR, line 28: (very stupid comment) I think you don't need the empty array for DatabaseConnection.query (it is by default an empty array)

PR #1998, opened 2022-06-12
Kay Dent Chitter Challenge partial solution

kaykakaraka:


README.md
```diff
@@ -1,15 +1,56 @@
 Chitter Challenge
```
raphaella-rose, line 1: I think this README is really detailed and makes it easy to navigate your project. :)

app.rb
```diff
@@ -0,0 +1,40 @@
+require 'sinatra/base'
```
raphaella-rose, line 1: Your controllers are 'skinny' and easy to read. :)

lib/peep.rb
```diff
@@ -0,0 +1,45 @@
...
+    chars = post.chars
+    new_chars = []
+    chars.each do |char|
+      char = "''" if char == "'"
+      new_chars << char
+    end
+    new_chars.join
+  end
+
+  def self.choose_database
```
raphaella-rose, line 36: This method could be moved into a database setup object.

lib/user.rb
```diff
@@ -0,0 +1,37 @@
+class User
+
```
raphaella-rose, line 2: Maybe you could create an initialize method to store the name, username as instance variables so you could remove lines 11-21?

views/chitter/welcome.erb
```diff
@@ -0,0 +1,4 @@
+<% if !@username.nil? %>
+Welcome to Chitter, <%=@username%>
+<% end %>
+<a href='/chitter/compose'>Compose a peep</a> <p></p>
```
raphaella-rose, line 4: I'm not sure if the p tag needs to be here

raphaella-rose (commented):
I think this is great and well rounded. Your README made it clear and easy to follow the code and everything seems to work well. There are some places where you could refactor to cut out some code, particularly in the peep.rb file. Really good job :)

PR #2005, opened 2022-06-13
chitter challenge partial solution

raphaella-rose:


db/migrations/01_create_chitter_table.sql
```diff
@@ -0,0 +1 @@
+REATE TABLE chitter(id SERIAL PRIMARY KEY, peep VARCHAR(120));
```
kaykakaraka, line 1: missing the C at the beginning

lib/peep.rb
```diff
@@ -0,0 +1,25 @@
+require 'pg'
+
+class Peep
+  def self.post
+    if ENV['ENVIRONMENT'] == 'test'
+      connection = PG.connect(dbname: 'chitter_test')
+    else
+      connection = PG.connect(dbname: 'chitter')
+   end
```
kaykakaraka, line 9: can extract this 'if' statement to a separate method as it is repeated

spec/peep_spec.rb
```diff
@@ -0,0 +1,24 @@
...
+      expect(peep).to include('Hello from user1')
+    end
+  end
+
+  describe '.new' do
+    it 'creates new peep' do
+      Peep.new(peep: 'Hello from user1')
+
+      expect(Peep.post).to include 'Hello from user1'
+    end
```
kaykakaraka, line 21: I think these two tests test the same thing - I wonder if there's anything that shows their function is different?

app.rb
```diff
@@ -0,0 +1,31 @@
...
+    redirect '/chitter'
+  end
+
+  get '/chitter/new' do
+    erb :'chitter/new'
+  end
+
+  run! if app_file == $0
+
+end
```
kaykakaraka, line 31: very 'skinny' and readable

spec/features/posting_peep_spec.rb
```diff
@@ -0,0 +1,9 @@
+feature 'posting peep' do
+  scenario 'A user can post a peep to Chitter' do
+    visit('/chitter/new')
+    fill_in('peep', with: 'Hello from user1')
+    click_button('Post')
+
+    expect(page).to have_content 'Hello from user1'
```
kaykakaraka, line 7: could also use this test to specify how you want to see the peep - ie. not in an array

spec/features/posting_peep_spec.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+end
```
kaykakaraka, line 9: tests are decoupled from implementation

Gemfile
```diff
@@ -1,11 +1,18 @@
...
+gem 'sinatra-flash'
+gem 'pg'
+
 
 group :test do
   gem 'rspec'
   gem 'simplecov', require: false
   gem 'simplecov-console', require: false
+  gem 'capybara'
 end
```
kaykakaraka, line 17: test gems are separated well

spec/peep_spec.rb
kaykakaraka: you could for example have 'eq' the second one (and this means you will need to clear your database before each test)

views/chitter.erb
```diff
@@ -0,0 +1 @@
+<%= @peep %>
```
kaykakaraka, line 1: Could this have a link to the chitter/new page so that multiple peeps can be added?

spec/features/posting_peep_spec.rb
kaykakaraka: and there's a clear differentiation between feature tests and unit tests

kaykakaraka (commented):
What you have looks very good! Before doing the next user stories you could add in something to clear the database before each test and do a small refactor of the methods you have. You might find it easier if your tables have different names from your database. It's a good start and of good quality, vey easy to follow and read.

PR #2012, opened 2022-06-13
Pull

brownc2:


spec/spec_helper.rb
```diff
@@ -1,6 +1,17 @@
+# Set the environment to "test"
+ENV['RACK_ENV'] = 'test'
```
BenjaminNeustadt: I wonder whether it would be a good idea to anticipate having two databases at this point that link up to the program, one for testing.

views/chitter/index.erb
```diff
@@ -0,0 +1,39 @@
+<html>
+
+<style>
+body {
+    background-color: #c4c4c4;
+  
+}
+
+.text {
+    position: absolute;
```
BenjaminNeustadt, line 7: The site looks really good, all this CSS makes it look great. I wonder whether you could have all your css on a style sheet inside a public folder, that SInatra would go and find before running the program?

spec/chitter_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'chitter'
+
+describe '.all' do
+  it 'returns a list of bookmarks' do
+    connection = PG.connect(dbname: 'chitter_test')
+    
+
+    # Add the test data
+    connection.exec("INSERT INTO chat_log (chat_log) VALUES('Hello');")
+    connection.exec("INSERT INTO chat_log (chat_log) VALUES('That is true');")
```
BenjaminNeustadt, line 10: Great tests, really clear and easy to understand.

spec/setup_test_database.rb
```diff
@@ -0,0 +1,8 @@
+require 'pg'
+
+p "Setting up test database..."
+
+def setup_test_database
+  connection = PG.connect(dbname: 'bookmark_manager_test')
```
BenjaminNeustadt, line 6: It might be a good idea to take the conditional for the connection in the model chitter.rb, and place it into here, since this is where we're dealing with that stuff. 
```
if ENV['ENVIRONMENT'] == 'test'
      connection = PG.connect(dbname: 'chitter_test')
    else
    end
```

BenjaminNeustadt (commented):
Really nice. I would suggest having a longer README, maybe even detailing your plan, and also I would suggest longer and more detailed commit messages. Really nice work, and great css 👍

PR #2014, opened 2022-06-19
Change model peep

BenjaminNeustadt:
1. Refactored peep_spec.rb unit test to include Peep.create instead of using sql to insert a new instance of a peep inside the database. 
2. Included in the peep_spec the method .to_s, in order for the test to pass
3. All tests are passing.
4. Created new instructions for db/migrations with 02_create_users_table
5. Deleted unnecessary instructions from the db/migrations folder that are obsolete.


BenjaminNeustadt (commented):
Approved changes, will fix conflicts locally. (conflicts will include double instances of particular tests)

PR #2017, opened 2022-06-20
David Johnson

fridayshoes:


lib/peeps.rb
```diff
@@ -0,0 +1,28 @@
+require 'pg'
+
+class Peeps
+
+  attr_reader :id, :peep, :created_on
+  
+  def initialize(id:, peep:, created_on:)
+    @id  = id
+    @peep = peep
+    @created_on = created_on
```
michaelcychan, line 10: I think the variable names are easy to read. Good name!

views/peeps/peep_list.erb
```diff
@@ -0,0 +1,6 @@
+<p>Peeps List</p>
+<ul>
+  <% @peeps.each do |message| %>
+    <li><%= message.peep %> <%= message.created_on[0...-10] %></li>
```
michaelcychan, line 4: I like you chose to display the more relevant part of the timestamp rather than the whole timestamp from database.

spec/peeps_spec.rb
```diff
@@ -0,0 +1,14 @@
+require 'peeps'
+
+describe Peeps do
+  describe '.all' do
+    it 'returns all peeps' do
+      peeps = Peeps.all
+
```
michaelcychan, line 7: I think it is better if you show how the test would be expecting these output.

spec/features/viewing_chitter_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+end
+
+feature 'Viewing Peeps on Chitter' do
+  scenario 'A user can see a list of peeps' do
+    visit('/peeps')
+    
+    expect(page).to have_content "hello everyone"
+    expect(page).to have_content "Still got this bloomin lurgy and high temperature."
+    expect(page).to have_content "Why is he sniffing so much?"
+    expect(page).to have_content "I am in the curry house"
```
michaelcychan, line 15: I think as you have shown time in the views, you can also expect(page) to have the date as content. That would be a proof that every variable you decide to show is actually showing.

michaelcychan (commented):
Nice chitter clone!

PR #2028, opened 2022-07-18
hello

shaunywho:


app.rb
```diff
@@ -0,0 +1,104 @@
...
+    register Sinatra::Reloader
+    also_reload 'lib/post_repository'
+    also_reload 'lib/user_repository'
+  end 
+
+  enable :sessions
+
+  def check_logged_in
+    if session[:user_id] == nil
+      redirect '/'
```
karolinaspirulina, line 23: could be a good time to use the shorter syntax of  `redirect '/' if session[:user_id] == nil ` for clarity

app.rb
```diff
@@ -0,0 +1,104 @@
...
+    end
+  end 
+
+  def set_session(username, user_id)
+    session[:username] = username
+    session[:user_id] = user_id
```
karolinaspirulina, line 29: i really like that you've isolated the session code!

app.rb
```diff
@@ -0,0 +1,104 @@
...
+  end 
+
+  get '/' do
+    return erb(:index)
+  end 
+
+  get '/message_board' do
```
karolinaspirulina, line 36: do you need to print the session[:user_id]? could be nice to use some whitespace for clarity

app.rb
```diff
@@ -0,0 +1,104 @@
...
+    id = repo.login(username, password)
+    @invalid_entry = false
+    if id != nil
+      session[:user_id] = id
+      session[:username] = username
+      redirect '/message_board'
+    else
+      @invalid_entry = true
+      return erb(:login)
+    end
```
karolinaspirulina, line 66: Nice! I like the use of @invalid_entry

lib/post_repository.rb
```diff
@@ -0,0 +1,99 @@
...
+  end 
+
+  def all_sorted
+    all.sort_by{|post| Date.parse(post.post_date)}
+  end 
+
+
+  
+  private
+
```
karolinaspirulina, line 77: i like these mapping methods & that they're private :)

seeds/database_setup.sql
```diff
@@ -0,0 +1,2 @@
+CREATE TABLE IF NOT EXISTS users (id SERIAL PRIMARY KEY, email TEXT, username TEXT, password TEXT);
```
karolinaspirulina, line 1: i think it would be nice if you added some keywords to. your column definitions such as NOT NULL or use a bigger range of data types to reflect the use/necessity of the data better. Maybe make use of some default values too to make it easier on yourself in the long run such as for post_date.

PR #2031, opened 2022-08-15
Chitter challenge - Naomi

naomischlosser:
Password check with bcrypt to login is not working (it always returns false). Only got as far as creating the homepage that shows boxes to login and signup, but none of these are actually working yet. I started with the page once you are logged in that shows the peeps, but this is also not finished.


lib/user_repository.rb
```diff
@@ -0,0 +1,38 @@
...
+    return user.password == BCrypt::Password.new(encrypted_submitted_password)
+  end
+
+  def find_by_email(email)
+    sql = 'SELECT * FROM users WHERE email = $1;'
+    params = [email]
+    result = DatabaseConnection.exec_params(sql, params)[0]
+
+    user = User.new
+    user.id = result['id']
```
ladantork, line 29: Might be is better if set the id to integer     user.id = result['id'].to_i

lib/user_repository.rb
```diff
@@ -0,0 +1,38 @@
...
+    user.email = result['email']
```
ladantork, line 30: Looks like there is something wrong with email here

lib/peep_repository.rb
```diff
@@ -0,0 +1,30 @@
...
+      peep = Peep.new
+      peep.id = record['id']
+      peep.content = record['content']
+      peep.time = record['time']
+      peep.user_id = record['user_id']
+
+      peeps << peep
+    end
+
+    return peeps
```
ladantork, line 20: Need to return all peeps in reverse chronological order

PR #2032, opened 2022-08-15
Databases

ladantork:


docs/chitter_repository_class_design.md
```diff
@@ -0,0 +1,283 @@
+# Albums Model and Repository Classes Design Recipe
```
naomischlosser, line 1: Good shout of adding the .md files to the docs folder! It keeps the folder stucture clean.

I also love the names for your methods and classes (they are the same as mine haha)

spec/seeds/peeps_seeds.sql
```diff
@@ -0,0 +1,19 @@
+DROP TABLE IF EXISTS peeps;
+
+-- Table Definition
+CREATE TABLE peeps (
+  id SERIAL PRIMARY KEY,
+  content text,
+  time time,
```
naomischlosser, line 7: What about using timestamp? It can store both date and time. Because Ruby has the function called Time.now that also includes the timezone, it might be interesting to add this as well. You can always decide what information you actually want to display in the app.

spec/seeds/users_seeds.sql
```diff
@@ -0,0 +1,15 @@
+DROP TABLE IF EXISTS users;
+
+-- Table Definition
+CREATE TABLE users (
+  id SERIAL PRIMARY KEY,
+  name text,
+  email varchar
```
naomischlosser, line 7: I didn't know about this datatype. When reading a bit more about it, it says it's a variable- length character string. How is that controlled, do you know? Just curious.

lib/peep_repository.rb
```diff
@@ -0,0 +1,42 @@
...
+class PeepRepository
+   
+    def all
+        peeps = []
+      # Executes the SQL query:
+      sql='SELECT id, content, time, user_id FROM peeps;'
+      result_set = DatabaseConnection.exec_params(sql, [])
+      result_set.each do |record|
+        peep = Peep.new
+        peep.id = record['id'].to_i
```
naomischlosser, line 12: I am not sure if to_i is necessary here.

lib/peep_repository.rb
```diff
@@ -0,0 +1,42 @@
...
+        peep.content = record['content']
+        peep.time = record['time']
+        peep.user_id = record['user_id']
+        peeps << peep
+      end
+      return peeps
+    end
```
naomischlosser, line 19: Would you mind adding enters between certain blocks of code (e.g. between line 9 and 10 above) and between methods (e.g. between line 19 and 20). I think your code will be a bit more readable, especially when files become longer and more complex.

spec/peep_repository_spec.rb
```diff
@@ -0,0 +1,23 @@
...
+describe PeepRepository do
+  before(:each) do 
+    reset_peeps_table
+  end
+
+  it 'returns all peeps' do
+    repo = PeepRepository.new
+    peeps = repo.all
+    expect(peeps.length).to eq(5)
+
```
naomischlosser, line 19: You might want to check a few more variables here than just the length.

spec/user_repository_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    expect(users.first.name).to eq("sara")
+  end
+
+
+  it ' find specific user' do
+    repo = UserRepository.new
+
+    user = repo.find(2)
+    expect(user.id).to eq(2)
+  end
```
naomischlosser, line 27: Have a look at the 'Creating and signing in users' John posted in the July channel on Friday. It tells you a lot about how to approach this, because the next step will be to login a user and you will need to find the user based on their email, not their id.

PR #2033, opened 2022-08-15
Chitter Challenge!!!

ZacMossHK:


app.rb
```diff
@@ -0,0 +1,185 @@
...
+
+  post '/login' do
+    # verifies login info
+
+    # returns error status 400 if no user is logged in or parameters are given
+    return status 400 if !session[:user_id].nil? ||
+      params[:password].nil? || params[:username].nil?
+    Application.escape_html_all_params(params)
+
+    # checks username is in database, nil if not
```
ruben-wilson, line 67: why is there a bang only the session user_id?

app.rb
```diff
@@ -0,0 +1,185 @@
...
+    redirect to '/login?password_error=true' if params[:password] != user.password
+    session[:user_id] = user.id
+    redirect to '/'
+  end
+
+  get '/log-off' do
+    # logs the user off
+
+    # resets :user_id to nil
+    session.delete(:user_id)
```
ruben-wilson, line 83: you might need to delete all the session values u created , like id and password

email_test.rb
```diff
@@ -0,0 +1,14 @@
+require 'sendgrid-ruby'
+include SendGrid
+
+from = Email.new(email: 'zac@zacmoss.co.uk')
+to = Email.new(email: 'ss@email.com')
+subject = '345'
```
ruben-wilson, line 6: maybe wants to be in your spec file

PR #2034, opened 2022-08-15
chitter-challenge

BrijeshPujara:
Partial completion/WIP


lib/peeps_repo.rb
```diff
@@ -0,0 +1,48 @@
...
+    result_set = DatabaseConnection.exec_params(sql, [peep.content, peep.date, peep.user_id])
+
+    return peep
+  end
+
+  def delete(id)
+    sql = 'DELETE FROM peeps WHERE id = $1;'
+    DatabaseConnection.exec_params(sql, [id])
+  end
+
```
SophLewDev, line 47: Looking at the user stories, you might not need this method, but it's handy!

spec/integration/app_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  context 'GET /' do
+    it 'returns 200 OK and a form' do
+      response = get('/')
+      expect(response.status).to eq(200)
+      expect(response.body).to include('<form action="/signup" method="POST">')
+    end
+  end
+end
```
SophLewDev, line 27: Maybe do tests first for the routes before implementing them on app.rb

PR #2044, opened 2022-09-20
Chitter Challenge 

iniffur:


lib/peep_repository.rb
```diff
@@ -0,0 +1,41 @@
+require_relative 'peep'
+
+class PeepRepository
+  def all
+    peeps = []
+
+    query = 'select * from peeps;'
```
itsmosho, line 7: good use of * to quickly select all columns from peep table

app.rb
```diff
@@ -0,0 +1,126 @@
...
+    register Sinatra::Flash
+    enable :sessions
+    also_reload 'lib/peep_repository'
+    also_reload 'lib/user_repository'
+  end
+
+  get '/peeps' do
+    peeps_repo = PeepRepository.new
+    @peeps = peeps_repo.all
+    @sorted_peeps = @peeps.sort_by { |peep| peep.peep_time }.reverse!
```
itsmosho, line 25: good way to use the reverse method to list peeps in reverse chronological order

app.rb
```diff
@@ -0,0 +1,126 @@
...
+    users.each do |user|
+      if @new_user_username == user.username 
+        status 400
+        return 'ERROR: Username is already taken'
+      elsif @new_user_email == user.email
+        status 400
+        return 'ERROR: Email is already in use'
+      end
+    end
+
```
itsmosho, line 96: Could create signup method in user repository, then call it in app to prevent large chunks of code in your application. Also good to be able to unit test the repository to know the code works. But this example does fulfil the function

itsmosho: look at password encryption to keep passwords safe in database

PR #2045, opened 2022-09-20
completed the chitter challenge with basic functionality

jonas-diete:


app.rb
```diff
@@ -0,0 +1,150 @@
...
+class Application < Sinatra::Base
+  # This allows the app code to refresh
+  # without having to restart the server.
+  configure :development do
+    register Sinatra::Reloader
+    also_reload 'lib/peep_repository'
+    also_reload 'lib/user_repository'
+  end
+
+  enable :sessions
```
44jovi, line 18: Interestingly when I left this `enable :sessions` outside of the `configure` block, it wouldn't work for me

app.rb
```diff
@@ -0,0 +1,150 @@
...
+
```
44jovi, line 19: This whole file is really clean - I like that you sectioned it with comments (e.g. '#Login Form)

app.rb
```diff
@@ -0,0 +1,150 @@
...
+    end
+    # check password matches
+    if BCrypt::Password.new(user.password) == submitted_password
+      return user
+    else
+      return nil
+    end
+  end
+
+  def check_sign_up_credentials(repo, email, username, password1, password2)
```
44jovi, line 121: I'm thinking these methods should live in the UserRepository, so that your app.rb is as dedicated to redirecting the user through the different HTTP routes as possible

44jovi (commented):
The code is really clean and the logic flow is sound, great job on this

PR #2051, opened 2022-09-20
partial solution

Sadat15:


db/migrate/20220920084842_rename_author_to_user_id.rb
```diff
@@ -0,0 +1,9 @@
+class RenameAuthorToUserId < ActiveRecord::Migration[7.0]
+  def change
+    remove_column :peeps, :author_name
+
+    add_column :peeps, :user_id, :bigint
+
+    add_foreign_key :peeps, :users
+  end
+end
```
ritaaktay, line 9: interesting why you added this file rather than changing the other migrations `CreatePeeps` & `CreateUsers`

app.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  get "/" do
+    @peeps = Peep.all.order(created_at: :desc)
+    if session[:session_user_id]
+      @user = User.find(session[:session_user_id])
+    end
+    erb :index
+  end
+
+  post "/new_peep/:id" do
```
ritaaktay, line 21: I think the convention of creating a new resource in CRUD / REST apps is POST request to `/peeps` or` /users` etc. And then you POST path can redirect to `/user/:id` or `/peep/:id` once it has created it (or back to the submission page with error if not created)

app.rb
```diff
@@ -0,0 +1,62 @@
...
+    end
+  end
+
+  get "/sessions/new" do
+    erb :"sessions/new"
+  end
+
+  post "/sessions" do
+    user = User.where("email": params["email"]).first
+    if user.authenticate(params["password"])
```
ritaaktay, line 53: interesting so when you say `has_secure_password` ActiveRecord adds an `authenticate` method that checks for BCrypt encryption?

ritaaktay (commented):
Some more tests could have been useful!

PR #2052, opened 2022-09-26
Not finished

GuillerminaLorenzo:


app.rb
```diff
@@ -0,0 +1,91 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/database_connection'
+require './lib/user_repository'
+require './lib/user'
+require './lib/peep_repository'
+require './lib/peep'
+
+DatabaseConnection.connect('chitter_test')
```
44jovi: Should this point to the real database (e.g. 'chitter')

app.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  enable :sessions
+  
+  get '/' do
+    repo = PeepRepository.new
+    @peeps = repo.all.sort_by(&:post_time).reverse
+
+    return erb :index, :layout => :main_layout if session[:user_id].nil?
+    return erb :index, :layout => :main_layout
```
44jovi, line 25: Is this repeated so you can insert a custom page for if a user is not logged in?
Otherwise it looks like you only need line 25 here.

app.rb
```diff
@@ -0,0 +1,91 @@
+require 'sinatra/base'
+require 'sinatra/reloader'
+require './lib/database_connection'
+require './lib/user_repository'
+require './lib/user'
+require './lib/peep_repository'
+require './lib/peep'
+
```
44jovi, line 8: Very clear code/layout in this app.rb

lib/peep_repository.rb
```diff
@@ -0,0 +1,37 @@
...
+      peep
+    end
+  end
+
+  def create(peep)
+    sql = 'INSERT INTO peeps (content, post_time, user_id) VALUES ($1, $2, $3) RETURNING id;'
+    param = [peep.content, peep.post_time, peep.user_id]
+    results = DatabaseConnection.exec_params(sql, param)
+    peep.id = results.first['id'].to_i
+    return peep
```
44jovi, line 25: I think `return peep` isn't needed here - this could be causing the issue with not being able to create a peep in app.rb. 

I think you can just do `return nil` as this method's job is just to insert a peep into the database.

lib/peep_repository.rb
```diff
@@ -0,0 +1,37 @@
...
+    result_set = DatabaseConnection.exec_params(sql, params)
+    result = result_set[0]
+    peep = Peep.new
+    peep.id = result['id'].to_i
+    peep.content = result['content']
+    peep.post_time = result['post_time']
+    peep.user_id = result['user_id'].to_i
+    peep
+  end
+end
```
44jovi, line 40: Another very clean layout here on this file

PR #2054, opened 2022-10-09
"Straight Up" and "Harder" features implemented

kateusacova:
See README for details.


app/models/user.rb
```diff
@@ -0,0 +1,21 @@
...
+
+  validates :real_name, 
+  presence: true,
+  length: { in: 6..30 }
+
+  validates :password,
+  length: { in: 6..30 }
+
+  has_secure_password
+end
```
Nameo91, line 21: All inputs are well validated

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,113 @@
...
+    if logged_in?
+      status 400
+    else
+		  erb :login
+    end
+	end
+
+	post "/login" do
+		user = User.find_by(:username => params[:username])
+
```
Nameo91, line 52: find_by(:username) can be put together find_by_username.

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,113 @@
...
+    
+    if @peep.save
+      redirect "/"
+    else  
+      erb :create_peep
+    end
+	end
+
+  private
+
```
Nameo91, line 99: Love you separated method. very readable

app/controllers/application_controller.rb
```diff
@@ -0,0 +1,113 @@
...
+    usernames.each do |username|
+      if words.include?("@#{username}")
+        tagged_users << username
+      end
+    end
+    tagged_users
+  end
+
+end
```
Nameo91, line 121: Indentation can make code more readable

Nameo91 (commented):
Well written erb files and variable names. The values for tables are well validated. If you can add spec for app, I will visualise what we should expect from the app.

app/controllers/application_controller.rb
kateusacova: Github's fault! :(

PR #2058, opened 2022-10-10
Calum Wilmot partial Chitter solution

Calumwi:


app.rb
```diff
@@ -0,0 +1,67 @@
...
+    login_email = params[:email]
+    login_password = params[:password]
+
+    user = repo.login(login_email, login_password)
+    if user != nil
+        return erb(:peeper)
+    end
+    return erb(:failed_login)
+  end
+end
```
piperrosearmstrong, line 67: I see that you are yet to test the app.rb. In future, it would be good to adopt a test-driven approach.

piperrosearmstrong, line 67: It's great to see that your sign-up function works - a good next addition might be a failed log-in page.

app.rb
```diff
@@ -0,0 +1,67 @@
...
+  post '/peep' do
+    repo = PeepRepository.new
+    content = params[:content]
+    repo.create(content)
+    return erb(:peeper)
+  end
+
+#
+
+  post '/login' do
```
piperrosearmstrong, line 55: This is a great attempt at the login section. If you feel that you would like to finish it - seek coach support.

PR #2064, opened 2022-10-10
Partial Chitter Challenge

saritahub:


Recipes/Chittter_Plan.md
```diff
@@ -0,0 +1,257 @@
...
+
+3. Include time of peep
+- Add timestamp parameter to each peep 
+
+4. Sign up for Chitter (using a form)
+- Email
+- Password
+- Full name
+- Username  
+(e.g. samm@makersacademy.com, password123, Sam Morgan, sjmog).
```
eggs-benny, line 58: Really clear goal setting!

Recipes/Table_Recipe.md
```diff
@@ -0,0 +1,194 @@
...
+username: VARCHAR
+fullname: text
+password: VARCHAR
+
+Table: peeps
+id: SERIAL
+message: VARCHAR
+username: VARCHAR
+fullname: text 
+timestamp: TIMESTAMP
```
eggs-benny, line 127: good use of VARCHAR and TIMESTAMP

app/Models/peep.rb
```diff
@@ -0,0 +1,40 @@
...
+      connection = PG.connect :dbname => 'chitter_test'
+      result = connection.exec("SELECT * FROM peeps;")
+      result.map do |row|
+        row
+        new(id: row["id"],
+            message: row["message"],
+            timestamp: row["timestamp"])
+      end
+    rescue PG::Error => e
+      puts e.message
```
eggs-benny, line 16: Is there a clearer name than 'e' you could use here?

spec/Features/create_peeps_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+
+feature 'Creating peeps' do
+  before { Peep.create(message:'The CREATE message', timestamp: '1980-10-10 00:00:00')}
+  scenario "Add a peep and see it on the page" do
+    visit('/peeps')
+    click_button('PostPeep')
+    expect(page).to have_content('The CREATE message')
+    expect(page.current_path).to have_content('/peeps')
+  end
```
eggs-benny, line 19: Although I haven't encountered this approach before, it feels like it's very readable and clear on what it will do!

spec/spec_helper.rb
```diff
@@ -1,5 +1,44 @@
 require 'simplecov'
 require 'simplecov-console'
+require '/Users/saritaradia/Desktop/Projects/chitter-challenge/chitter-challenge/spec/database_setup.rb'
+require '/Users/saritaradia/Desktop/Projects/chitter-challenge/chitter-challenge/app.rb'
+require '/Users/saritaradia/Desktop/Projects/chitter-challenge/chitter-challenge/spec/Support/database_cleaner.rb'
```
eggs-benny, line 5: I'm not 100% sure, but these file paths look like they might only work on your local system - if someone were to fork this, you might need more general requires or something like require_relative './spec/database_setup.rb' ?

eggs-benny (commented):
Overall a really clear project setup and work so far. I can already picture how your final system would work. As I mentioned on the call, it's also great that you've approached this in your own way, and shows that you have the confidence to try and write from scratch vs copy & pasting and not really understanding!

PR #2065, opened 2022-10-10
Dave T pull request

Dmum303:


lib/peep_repo.rb
```diff
@@ -0,0 +1,31 @@
+require 'peep'
+
+class PeepRepository
+    def initialize
```
PatriciaGN, line 4: I don't think you need the initialize method here.

lib/peep_repo.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    end
+
+    def all
```
PatriciaGN, line 8: Very concise and readable methods :)

PatriciaGN (commented):
Nice work, so far it all seems tidy and readable!

PR #2083, opened 2022-11-07
Chitter complete

kwatts949:
Completed: post a message, see all peeps in reverse chronological order, see the time, sign up, log in, log out (sort of!)

Not completed: email when tagged, need to add username to peep


spec/chitter_table.sql
```diff
@@ -0,0 +1,9 @@
+CREATE TABLE users ( id SERIAL PRIMARY KEY,
+name text, username text, email text, password text
+);
+
+CREATE TABLE peeps ( 
+  id SERIAL PRIMARY KEY, title text,
+  content text, time int, user_id int,
```
mixxxxxer: It will be better to use the current time for a new message # time DATE DEFAULT NOW() -> return today date, or time DATETIME DEFAULT NOW() -> return current date with time for each new message

PR #2086, opened 2022-12-04
chitter-challenge

yulingyou:


cbai123:
In the repository classes the find and all methods for posts and users are quite long, perhaps part of it could be put into a private method to reduce the size of method.

cbai123:
All tests passed except for 2 in post_repository_spec, it expected create_at to equal a timezone of +08 but was receiving +01. I think this problem comes in during seeding the tables, the date put in has no timezone specified, so maybe in my computer it’s defaulting to +1.

cbai123:
Website works when using rackup.

Good use of checking for invalid input.

cbai123:
Could maybe add tests in app_spec that check when a new post is added it can be found on the home page.

PR #2088, opened 2022-12-04
chitter challenge with CSS

shussain894:


app.rb
```diff
@@ -0,0 +1,84 @@
...
+  post '/login' do #using the login method below, it will allow the user to login to the site and it will redirect them to peeps
+    login
+    redirect '/peeps'
+  end
+
+  def login
+    repo = UserRepository.new
+    all = repo.all
+    all.each do |user|
+      if user.password != params[:password]
```
MthBarber, line 75: User.password and user.email should be switched to help authentication

PR #2093, opened 2022-12-05
Partial Chitter Challenge submission

RemoteUnderscore:
Unfortunately due to being away all weekend with family as well as experiencing a massive gem conflict issue this is a partial submission.


jxc136 (commented):
Issues with gems stymied progress - but the database has been well modelled and implemented well. Avoiding comments on anything that is obviously WIP - but this is a good start Mike!

lib/post_repository.rb
```diff
@@ -0,0 +1,59 @@
...
+    post.user_id = result_set[0]['user_id'].to_i
+
+    return post
+  end
+
+  def create(post)
+    sql = 'INSERT INTO posts (message, message_time, user_id) VALUES($1, $2, $3);'
+    result_set = DatabaseConnection.exec_params(sql, [post.message, post.message_time, post.user_id])
+
+    return post
```
jxc136, line 39: Create methods (usually) won't return anything - watch out that this doesn't trip you up when you develop the app class!

lib/post_repository.rb
```diff
@@ -0,0 +1,59 @@
...
+    post.message = result_set[0]['message']
+    post.message_time = result_set[0]['message_time']
+    post.user_id = result_set[0]['user_id'].to_i
+
+    return post
+  end
+
+  def create(post)
+    sql = 'INSERT INTO posts (message, message_time, user_id) VALUES($1, $2, $3);'
+    result_set = DatabaseConnection.exec_params(sql, [post.message, post.message_time, post.user_id])
```
jxc136, line 37: Considering looking into DateTime.Now for setting the date and time automatically upon a message creation - this would avoid any user error in input breaking things.

PR #2098, opened 2022-12-05
Chitter Challenge

aimeehorwood:
Straight up section: 3/4 complete
Harder : 1/2 complete

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,87 @@
+require "sinatra/base"
+require "sinatra/reloader"
+require "./lib/database_connection"
+require_relative "lib/peep_repository"
+require_relative "lib/user_repository"
+require "date"
+require "bcrypt"
+
+DatabaseConnection.connect("chitter_test")
+
+class Application < Sinatra::Base
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+    also_reload "lib/peep_repository"
+    also_reload "lib/user_repository"
+  end
+
+  get "/peep/new" do
+    @peeps = PeepRepository.new
+    @users = UserRepository.new
+    return erb(:homepage)
+  end
+
+  post "/peep/new" do
+    new_peep = Peep.new
+
+    new_peep.created = DateTime.now.strftime "%Y/%m/%d %H:%M"
+    new_peep.content = params[:content]
+    new_peep.user_id = session[:user_id]
+
+    repo = PeepRepository.new
+
+    repo.create(new_peep)
+    return erb(:homepage)
+  end
+
+  get "/" do
+    session.clear
+    @peeps = PeepRepository.new
+    @users = UserRepository.new
+    return erb(:homepage)
+  end
+
+  get "/signup" do
+    return erb(:signup)
+  end
+
+  post "/signup" do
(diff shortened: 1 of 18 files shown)


views/signup.erb
```diff
@@ -0,0 +1,28 @@
+<link href="<%= url('/styles.css') %>" rel="stylesheet" type="text/css" />
+<h1> Sign up for Chitter </h1>
+<form action="/signup" method="POST">
+
+    <div>
+      <label> Name </label>
+      <input type="text" name= "name" />
```
atcq9876, line 7: A nice way to ensure users don't submit any blank fields is to put 'required' at the end of these input lines, e.g. `<input type="text" name= "name" required>`
This will prevent them from submitting while the field is blank, raising a dynamic error message on the page, and can save you from having to write similar code into app.rb
Check here for more info: https://www.w3schools.com/tags/att_input_required.asp

app.rb
```diff
@@ -0,0 +1,87 @@
+require "sinatra/base"
+require "sinatra/reloader"
+require "./lib/database_connection"
+require_relative "lib/peep_repository"
+require_relative "lib/user_repository"
+require "date"
+require "bcrypt"
+
+DatabaseConnection.connect("chitter_test")
```
atcq9876, line 9: Did you have any issues with connecting to the database? I took (what I think is) the same approach as you in using the template for database connection (including in lib/database_connection) that we used in previous weeks, but this was causing me a problem where test comments were actually being added to my database when I ran rspec - so I looked back at the web-applications music-library-app and copied all the database_connection info from there and this solved the problem. I'd recommend taking a look at that too (or my lib/database_connection  file and 'DatabaseConnection.connect' line in my app.rb) for future reference if you encounter similar problems

app.rb
```diff
@@ -0,0 +1,87 @@
...
+    @users = UserRepository.new
+    return erb(:homepage)
+  end
+
+  get "/signup" do
+    return erb(:signup)
+  end
+
+  post "/signup" do
+    if (params[:email].empty? || params[:password].empty? || params[:name].empty? || params[:user_name].empty?)
```
atcq9876, line 51: Check my comment in the signup.erb file re being able to save code here

lib/database_connection.rb
```diff
@@ -0,0 +1,28 @@
+# file: lib/database_connection.rb
```
atcq9876, line 1: See comment in app.rb re database connection

spec/seeds/seeds_chitter.sql
```diff
@@ -0,0 +1,15 @@
+
```
atcq9876, line 1: Nice way of inserting values into tables concisely, I repeat a lot of code in mine

views/login.erb
```diff
@@ -0,0 +1,20 @@
+<link href="<%= url("/styles.css") %>" rel="stylesheet" type="text/css" />
+<h1> Sign in to Chitter  </h1>
+<form action="/login" method="POST">
+
+    <div>
+      <label> Email </label>
+      <input type="email" name= "email" />
```
atcq9876, line 7: Try deleting the spaces after name= in these files?

PR #2109, opened 2022-12-05
partial solution

Elle8888:


anisha-11:
The code is done well and is very readable and can be understood well. 
For future , maybe you can put 'Time.now' instead of having to input the time manually when making a peep.

PR #2134, opened 2023-01-23
chitter-challenge-incomplete

Caroline-Amanquah:


lib/database_connection.rb
```diff
@@ -0,0 +1,26 @@
...
+# This class is a thin "wrapper" around the
+# PG library. We'll use it in our project to interact
+# with the database using SQL.
+
+class DatabaseConnection
+  # This method connects to PostgreSQL using the 
+  # PG gem. We connect to 127.0.0.1, and select
+  # the database name given in argument.
+  def self.connect
+    @connection = PG.connect({ host: '127.0.0.1', dbname: chitter_challenge_test })
```
HusnainRashid, line 12: You don't have to change the name of the parameter. Just add this code to the method.   
 if ENV['ENV'] == 'test'
      database_name = 'music_library_test'
    else
      database_name = 'music_library'
    end

lib/peep.rb
```diff
@@ -0,0 +1,3 @@
+class Post
```
HusnainRashid, line 1: class Peep

PR #2136, opened 2023-02-13
Chitter app completed

mirandaweston:


README.md
```diff
@@ -1,24 +1,10 @@
-Chitter Challenge
```
sameerasood, line 1: Really detailed git commit messages.

app.rb
```diff
@@ -0,0 +1,76 @@
...
+require 'rack/test'
+
+DatabaseConnection.connect('chitter_database_test')
+
+class Application < Sinatra::Base
+
+  def set_all_peeps
+    peep_repo = PeepRepository.new
+    user_repo = UserRepository.new
+    peeps = peep_repo.all.reverse!
```
sameerasood, line 24: Really interesting to see different approaches used to achieve same results.(to get peeps displayed in reverse chronological order)

views/index.erb
```diff
@@ -0,0 +1,53 @@
+<!DOCTYPE html>
```
sameerasood, line 1: Well laid out index page

spec/user_repository_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+    @repo = UserRepository.new
+  end
+
+  it 'retrieves all users' do
+    expect(@repo.all.length).to eq 3
+    expect(@repo.all.first.name).to eq 'user one'
+    expect(@repo.all.last.name).to eq 'user three'
+  end
+
+  it 'creates a new user with hashed password' do
```
sameerasood, line 26: Good implementation of mocking.

app.rb
```diff
@@ -0,0 +1,121 @@
...
+require 'sinatra/reloader'
+require 'rack/session/cookie'
+require_relative 'lib/database_connection'
+require_relative './lib/user_repository'
+require_relative './lib/peep_repository'
+require_relative './lib/peep'
+require_relative './lib/user'
+require 'rack/test'
+
+DatabaseConnection.connect('chitter_database_test')
```
leoht, line 11: The Sinatra application should connect to the "production" database (instead of the test one) — so when we run the server, the app can connect to the right DB when creating / reading peeps stored in it

app.rb
```diff
@@ -0,0 +1,121 @@
...
+class Application < Sinatra::Base
+
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+  end
+
+  def set_all_peeps
+    peep_repo = PeepRepository.new
```
leoht, line 22: I would suggest moving this method into one of the Repository classes, for example it could be a method `PeepRepository.find_all_peeps_with_users` — this is so we leave the concern / responsibility of querying peeps and users to the Repository classes, and the Sinatra application class is only dealing with HTTP concerns, and delegating the work to other classes. This would make the Sinatra app class lighter, and it would be easier to change and modify the code fetching the peeps alongside with their user, as it would be centralised into the Repository class code

spec/integration/application_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+    connection.exec(peep_seeds_sql)
+    connection.exec(user_seeds_sql)
+  end
+
+  context 'GET /' do
+    it 'shows a list of peeps with usernames and time they were posted' do
+      response = get('/')
+
+      expect(response.status).to eq 200
+      expect(response.body).to include 'loving chitter'
```
leoht, line 30: It might be good including a few more assertions here, to check for a few other peeps contents, instead of only one — in case there'd be a bug preventing all peeps to be correctly displayed (but just that one 'loving chitter'), we'd end up with a false positive test

spec/integration/application_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+
+      expect(response.status).to eq(200)
+      expect(response.body).to include('<form action="/sign-up" method="POST">')
+    end
+  end
+
+  context 'POST /sign-up' do
+    it 'creates a new user in the database' do
+      response = post('/sign-up', name: 'user four', email:'user_four@gmail.com', pass_word: 'password4', username: 'user_four')
+
```
leoht, line 67: If we want to verify the user has been correctly added (signed-up), we could then make another request to log in, and check whether we manage to login successfully with the same credentials? Currently this test doesn't really assert the user has been correctly created, only that we get a 200 OK response

views/index.erb
```diff
@@ -0,0 +1,52 @@
...
+<html>
+  <head>
+    <title>Peeps</title>
+    <style>
```
leoht, line 5: I'd suggest moving the CSS code into an external CSS file, to keep the HTML and CSS code separated and make it easier to update later — you can follow this guidance to load an external CSS file on the page: https://github.com/makersacademy/web-applications/blob/main/pills/loading_css_or_static_files.md

leoht (commented):
Really great work on that one! I left a few comments on a few things to improve, do let me know if you have any questions on the feedback :)

spec/integration/application_spec.rb
mirandaweston: Agreed! Done

PR #2139, opened 2023-02-13
pull exhibits partial chitter chatter prog

ChalkandFeather:
To be revised


lib/maker_repository.rb
```diff
@@ -0,0 +1,26 @@
+class MakerRepository
+  def all
+    makers = []
+
+    # Send the SQL query and get the result set.
+    sql = 'SELECT id, name, genre FROM makers;'
```
edwardfgt: The SQL here references the column titles from the previous project, I believe the information that you want is id, name, username, email password

spec/seeds/makers_seeds.sql
```diff
@@ -0,0 +1,21 @@
+DROP TABLE IF EXISTS makers; 
+
+-- Table Definition
```
edwardfgt: It could be good to separate your table setup from your seeds since you were having issues with your database. Once the tables are set up, each time you run your tests only the information in the table would then be refreshed

edwardfgt (commented):
Overall the code is very cleanly written and easy to read.

I think your indentation and file layout is very logical.

PR #2141, opened 2023-02-13
Chitter Challenge 

Ajmakovsky:


spec/app_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  # hello 
+  include Rack::Test::Methods
+
+  # We need to declare the `app` value by instantiating the Application
+  # class so our tests work.
+  let(:app) { Application.new }
+
+  # run seed files and start database connection 
+  before(:each) do 
+    chitter_seeds = File.read('spec/seeds/chitter_seeds.sql')
```
joel-powell, line 16: Can be nice to break these 3 lines out into a method so that it's clearer that they achieve a single goal

By having a method name like replace_tables it's very clear that before each test you are resetting the tables

joel-powell, line 16: Can be nice to break these 3 lines out into a method so that it's clearer that they achieve a single goal

By having a method name like replace_tables it's very clear that before each test you are resetting the tables

spec/peep_repository_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+  connection.exec(seed_sql)
+end
+
+describe PeepRepository do 
+
+  before(:each) do 
+    reset_peeps_table
+  end 
+
+  it 'returns all peeps - unsorted' do
```
joel-powell, line 16: nice clear test names

spec/peep_repository_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+                                                                                                          
+      peep = repo.find(1)
+
+      expect(peep.id).to eq "1"
+      expect(peep.contents).to eq "hello peeps"
+      expect(peep.time).to eq Time.new(2023, 1, 8, 4, 5, 6)
+      expect(peep.user_id).to eq "1"
+    end 
+
+    it 'returns peep with id 1 ' do
```
joel-powell, line 40: test here is for ID 1 but the test tests for ID 3

lib/peep_repository.rb
```diff
@@ -1,27 +1,97 @@
...
+    end 
+
+    # @sorted_peeps = 
+    # sorted_peeps = @peeps.sort_by { |peep| peep.time }
+    # return sorted_peeps
+    return peeps
     # Returns an array of peep objects.
   end
 
+  def sort_peeps(peeps)
```
joel-powell, line 51: Nice concise sorting method here!

lib/user_repository.rb
```diff
@@ -6,21 +6,53 @@ class UserRepository
...
   def find(id)
     # executes the SQL query: 
-    # SELECT id, name, email, username, and password FROM users WHEN id = $1; 
+    sql = 'SELECT id, name, email, username, password FROM users WHERE id = $1;' 
+    sql_params = [id]
+
+    result_set = DatabaseConnection.exec_params(sql, sql_params)
+
+    user = User.new
+    user.id = result_set[0]['id']
```
joel-powell, line 38: can use the .first method here to avoid the double square brackets

views/show_peeps.erb
```diff
@@ -0,0 +1,18 @@
+<html>
+  <head></head>
+  <body>
+  <h1>Peeps:</h1>
+  
+    <% @sorted_peeps.each do |peep| %>
```
joel-powell, line 6: Could implement a join query for peeps to users to avoid the nested loop here

PR #2144, opened 2023-02-13
Shamima[Chitter-challenge(incomplete)]

Shamima14:


spec/seeds/seeds_users.sql
```diff
@@ -0,0 +1,23 @@
+DROP TABLE IF EXISTS users;
```
SimpleLuke, line 1: The table can not be dropped because of the posts foreign key. You may try to use CASCADE or TRUNCATE both tables at the same time.

spec/seeds/seeds_posts.sql
```diff
@@ -0,0 +1,24 @@
+-- Drop the table if it exists
+DROP TABLE IF EXISTS posts;
```
SimpleLuke, line 2: The table can not be dropped because of the posts foreign key. You may try to use CASCADE or TRUNCATE both tables at the same time.

spec/user_repository_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+  describe UserRepository do
+    before(:each) do 
+      reset_users_table
+    end
+
+    it 'checks if the email is in use' do
+        repo = UserRepository.new
+        found_user = repo.find('mike@strangerthings.com')
+
+        if found_user.nil?
```
SimpleLuke, line 19: Instead of putting the condition in the test, you may try add the logic for checking existing user in the repository.

lib/user_repository.rb
```diff
@@ -0,0 +1,44 @@
...
+        user.id = result_set[0]['id'].to_i
+        user.name = result_set[0]['name']
+        user.username = result_set[0]['username']
+        user.email = result_set[0]['email']
+        user.password = result_set[0]['password']
+
+        return user
+    end
+
+    def does_username_exist(username)
```
SimpleLuke, line 25: You may try to add a test to this method.

lib/post_repository.rb
```diff
@@ -0,0 +1,49 @@
+require_relative 'post'
+
+class PostRepository
```
SimpleLuke, line 3: Class name and repository names are clear and easy to understand.

lib/user.rb
```diff
@@ -0,0 +1,5 @@
+class User
+
+    # Replace the attributes by your own columns.
```
SimpleLuke, line 3: Nice direction with comment.

lib/user_repository.rb
```diff
@@ -0,0 +1,44 @@
+require_relative 'user'
+
+class UserRepository
```
SimpleLuke, line 3: Class name and repository names are clear and easy to understand.

PR #2153, opened 2023-03-12
Sarah - Chitter challenge / not complete

sarahdavies186:
Functionality for all user stories except the advanced. Password is not secure yet. Still some bugs. CSS not complete.


app.rb
```diff
@@ -0,0 +1,98 @@
...
+    also_reload "lib/user_repository.rb"
+  end
+
+  get "/" do
+    erb(:index)
+  end
+
+  get '/feed' do
+    @peeps = PeepRepository.new.all.reverse
+    @user_repo = UserRepository.new
```
leoht, line 30: What data will you specifically need from `@user_repo` in your view? It's usually better to "give" to the view the least access possible to any database resources, and only pass it the data it needs to display information to the user — in that case, maybe `@peeps` and a variable `@current_user` set in the route block would be enough, instead of letting the view have access to the whole repository class methods

app.rb
```diff
@@ -0,0 +1,98 @@
...
+    erb(:login)
+  end
+
+  post '/login' do
+    email = params[:email]
+    password = params[:password]
+
+    users = []
+    user_repo = UserRepository.new
+    users << user_repo.find_by_email(email)
```
leoht, line 60: If you only need access to the one user with the given email address (the user currently trying to sign in), maybe you don't need an array of `users`?

app.rb
```diff
@@ -0,0 +1,98 @@
...
+    #not working properly
+    if users[0].password == password && users[0].email == email
+      user = users.first
+      session["user_id"] = user.id
+      redirect '/login_home'
+    else
+      @error = true
```
leoht, line 67: Since we're rendering a specific view `:login_failed` for this specific case, maybe having a variable `@error` set to true is redundant?

lib/user_repository.rb
```diff
@@ -0,0 +1,61 @@
+require_relative './user'
+
+class UserRepository
+  def all
+    users = []
+
+    sql = "SELECT * FROM users;"
+    result_set = DatabaseConnection.exec_params(sql, [])
+    result_set.each do |record|
+      user = User.new
```
leoht, line 10: The code creating a new `User` object and assigning values from the hash returned by the DB is reused a few times in the other methods — maybe you could factor it into a private method (it could be named `convert_hash_to_user_object`) to simplify / DRY the code?

views/login_home.erb
```diff
@@ -0,0 +1,35 @@
...
+    <title>Chitter</title>
+    <link rel="stylesheet" href="/style.css">
+  </head>
+  <body>
+    <h1>Chitter</h1>
+    <a href="/logout">Logout</a>
+    <h3>Start peeping now</h3>
+
+    <form action="/peeps" method="post">
+
```
leoht, line 16: Minor one - it'd be good to fix the indentation to improve the readability of the HTML structure (you can use any extension to auto-lint / format the code with your editor, to make things easier or automated whenever you save a file)

spec/integration/app_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+          '<input type="submit" name="button" value="Submit" />',
+        )
+      end
+    end
+
+    context "POST /" do
+      it "should add the new peep" do
+        _response = post('/peeps', message: 'hello')
+        response = get('/login_home')
+        expect(response.status).to eq(200)
```
leoht, line 107: This assertion doesn't verify that the new peep has been created — in case something was going wrong, we could still get a 200 OK response from `/login_home` but maybe the entry wouldn't be there (which would be a false positive) — how could you also verify that the new entry has been created in this test?

leoht (commented):
Really nice work, well done — I have a few points of feedback below, please let me know if you want to discuss it further or if you have any questions :)

PR #2165, opened 2023-04-10
project setup

toppy007:
Week 4 Solo Project Commited

+ Login
+ Sign Out
+ Message Board
+ Tags
+ Date
+ Sign Up
+ Database

### DIFF
--- a/app.rb
+++ b/app.rb
@@ -1,0 +1,117 @@
+# file: app.rb
+
+require 'sinatra'
+require "sinatra/reloader"
+require 'bcrypt'
+require_relative 'lib/database_connection'
+require_relative 'lib/user_repository'
+require_relative 'lib/message_repository'
+
+DatabaseConnection.connect
+
+class Application < Sinatra::Base
+
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+    also_reload 'lib/user_repository'
+    also_reload 'lib/message_repository'
+  end
+
+  get '/' do
+    repo = MessageRepository.new
+    @messages = repo.all.sort_by(&:created_at).reverse
+
+    @tags = []
+    
+    user_info = UserRepository.new
+    tag = user_info.all
+    tag.each do | username |
+        @tags << "@#{username.username}"
+    end
+    return erb(:index)
+  end
+
+  post '/' do
+    if invalid_request_parameters?
+      status 400
+      return ''
+    end
+
+    redirect '/login' if session[:user_id] == nil
+    
+    title = params[:title]
+    content = params[:content]
+    tags = params[:tags]
+
+
+    repo = MessageRepository.new
+    new_message = Message.new
+    new_message.title = params[:title]
+    new_message.content = params[:content]
+    new_message.posted_by = session[:user_id] 
+    new_message.tags = params[:tags]
+    
+    repo.create(new_message)
+
+    redirect '/'
+  end
+
+  get '/sign_up' do
+    return erb(:sign_up)
+  end
+
+  post '/sign_up' do
+    email = params[:email]
+    password = params[:password]
+    username = params[:username]
+    name = params[:name]
+
+    repo = UserRepository.new
+    new_user = User.new
+    new_user.email = params[:email]
+    new_user.password = params[:password]
+    new_user.username = params[:username]   
+    new_user.name = params[:name]  
+
+    repo.create(new_user)
+
+    return erb(:login)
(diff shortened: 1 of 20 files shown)


app.rb
```diff
@@ -0,0 +1,117 @@
...
+    if stored_password == password
+      session[:user_id] = user.username
+      redirect '/'
+    else
+      return erb(:login_error)
+    end
+  end
+
+  get '/log_out' do
+    destroy
```
tomcarmichael, line 104: I love the naming of this as 'destroy', but I would suggest it could be 'destroy_session' in order to make it a bit easier to guess what it's doing without checking the method definition

app.rb
```diff
@@ -0,0 +1,117 @@
...
+  post '/login' do
+    email = params[:email]
+    password = params[:password]
+
+    repo = UserRepository.new
+    user = repo.find(email)
+
+    stored_password = BCrypt::Password.new(user.password)
+    if stored_password == password
+      session[:user_id] = user.username
```
tomcarmichael, line 96: Perhaps it'd make more sense to set the session variable name to session[:username] to be more consistent with the data that you're assigning to it?

app.rb
```diff
@@ -0,0 +1,117 @@
...
+
+  get '/sign_up' do
+    return erb(:sign_up)
+  end
+
+  post '/sign_up' do
+    email = params[:email]
+    password = params[:password]
+    username = params[:username]
+    name = params[:name]
```
tomcarmichael, line 69: As far as I can tell lines 66 - 69 aren't doing anything useful and could be deleted , based on what you've done in the following lines below?

app.rb
```diff
@@ -0,0 +1,117 @@
...
+    return erb(:index)
+  end
+
+  post '/' do
+    if invalid_request_parameters?
+      status 400
+      return ''
+    end
+
+    redirect '/login' if session[:user_id] == nil
```
tomcarmichael, line 42: nice and concise, code looks neat

app.rb
```diff
@@ -0,0 +1,117 @@
...
+    
+    title = params[:title]
+    content = params[:content]
+    tags = params[:tags]
```
tomcarmichael, line 46: as per comment below, I think these variables dont get used and these three lines could be deleted?

lib/user_repository.rb
tomcarmichael, line 1: looks good to me!

spec/message_repository_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+    connection.exec(seed_sql)
+end
+
+RSpec.describe MessageRepository do
+
+    before(:each) do 
+        reset_messages_table
+    end
+
+    context 'testing the MessageRepository class' do
```
tomcarmichael, line 16: A minor note here and something I've only come to understand (or think I understand) more recently .. the context here should be the context in which the user is accessing the program, eg. 'when the user is logged in'. So that our test names describe the behaviour we want from our program, under certain conditions / contexts

spec/message_repository_spec.rb
```diff
@@ -0,0 +1,47 @@
...
+            message = repo.all
+
+            expect(message.length).to eq 3 
+            expect(message[0].id).to eq 1 
+            expect(message[0].title).to eq 'paired programming' 
+            expect(message[0].content).to eq 'learning to pair program'
+            expect(message[0].tags).to eq '@mat'
+        end
+
+        it 'create a new message' do
```
tomcarmichael, line 28: if you word this as 'creates a new message', it just reads slightly better when running the tests

views/index.erb
```diff
@@ -0,0 +1,155 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="UTF-8">
+        <meta name="viewport" content="width=device-width, initial-scale=1.0">
+        <title>Chitter Message Board</title>
+        <style>
+            body {
+                font-family: helvetica, sans-serif;
```
tomcarmichael, line 9: Styling looks great in the browser, but would suggest moving all the CSS code to a separate stylesheet file so you can reuse it more easily as the application grows

tomcarmichael (commented):
Think you've done a great job Chris! Tests are all passing and describe the functionality well. Think you've mostly written clean, concise and readable code. Just suggested removing some redundant lines of variable assisgnment where the variables do not subsequently get used, and where you've repeated the same CSS code, that could all be put in a single stylesheet file.

PR #2171, opened 2023-04-12
Makers Chitter Challenge March 23 Cohort

jonpillay:
All functionality working as pre spec, except for email. Needs CSS.


bcrypt.rb
penguat, line 1: I'm not sure how this contributes to Chitter - could we delete it?

controllers/application_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    #also_reload 'lib/user'
+    enable :sessions
+  end
+
+  get '/' do
+    peeps_repo = PeepRepository.new
+    @user_repo = UserRepository.new
+    @peeps = peeps_repo.all
+    erb(:"app/index")
+  end
```
penguat, line 25: I love this, super clean and clear!

controllers/application_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+    peep = Peep.new
+    peep.content = content
+    peep.author = author
+    peep_repository = PeepRepository.new
+    if peep.content.match(/@\w+/)
+      puts "we are here"
+    end
+    peep_repository.create(peep)
+    redirect '/'
+  end
```
penguat, line 39: This feels like it's doing a lot - could it be broken up to make it more digestible?

penguat (commented):
Hi Jon - i think there are a few places where you could have reviewed this yourself to see unnecessary code! Your structure looks good, and your code is clear, although you don't always break it up with newlines. You might consider when you need to make new methods to help out too.

PR #2172, opened 2023-04-12
Work In Progress Chitter app

tomcarmichael:
Began implementation of home page, using sessions to display a form to create new peep only if the user is logged in. Began implementation of login and register pages.


app.rb
```diff
@@ -0,0 +1,101 @@
...
+  enable :sessions
+
+  configure :development do
+    register Sinatra::Reloader
+    also_reload "lib/peep_repository"
+    also_reload "lib/user_repository"
+  end
+
+  get '/' do
+    @all_peeps = PeepRespository.new.all_by_rev_date_order_with_author
```
toppy007: method names could be shorter

app.rb
```diff
@@ -0,0 +1,101 @@
...
+    return erb(:index)
+  end
+
+  get '/login' do
+    if session[:user_id]
```
toppy007: think about developing a controller method/class.

spec/integration/application_spec.rb
```diff
@@ -0,0 +1,168 @@
...
+
+  include Rack::Test::Methods
+
+  let(:app) { Application.new }
+
+  describe "homepage" do 
+    it "displays all peeps in reverse chronological order" do
+      response = get('/')
+      expect(response.status).to eq(200)
+      expect(response.body).to match(/Big Brother is watching you @wsmith[\s\S]*@wsmith & @smhanna - this is jam hot, this is jam hot[\s\S]*We shall meet in the place where there is no darkness/)
```
toppy007, line 27: not sure I agree with such a large expects comment.
maybe break this up.

lib/peep_repository.rb
toppy007, line 1: is it better you make a request for the data then sort it accordingly.

spec/seeds.sql
toppy007, line 1: I would create a file called seeds to store all my database information with the spec.
e.g put chitter.sql and chitter_test.sql in the same location.

chitter.sql
```diff
@@ -0,0 +1,32 @@
...
+  message text,
+  posted_at timestamp,
+-- The foreign key name is always {other_table_singular}_id
+  user_id int,
+  constraint fk_user foreign key(user_id)
+    references users(id)
+    on delete cascade
+);
+
+-- Create the join table.
```
toppy007, line 25: this table seems unnecessary but I'd be interested to know why you've added it. p.s think this might be my understanding.

lib/database_connection.rb
```diff
@@ -0,0 +1,37 @@
...
+# with the database using SQL.
+
+class DatabaseConnection
+  # This method connects to PostgreSQL using the 
+  # PG gem. We connect to 127.0.0.1, and select
+  # the database name given in argument.
+  def self.connect
+      # If the environment variable (set by Render)
+    # is present, use this to open the connection.
+    if ENV['DATABASE_URL'] != nil
```
toppy007, line 14: nice to see this conditional statement ready to use.

lib/user.rb
```diff
@@ -0,0 +1,3 @@
+class User
+  attr_accessor :id, :name, :username, :email, :password
```
toppy007, line 2: from the Peeps table I can see you have the columns user and user_id, then in Users table you have the column username. reckon you can drop a few of these columns as the project progresses. I'm guilty of the same thing.

lib/user_repository.rb
```diff
@@ -0,0 +1,52 @@
...
+  end
+
+  def login(username, submitted_password)
+    user = find_by_username(username)
+    return { success: false, reason: "invalid username "} if !user
+    stored_password = BCrypt::Password.new(user.password)
+    if stored_password == submitted_password
+       return { success: true, username: user.username, user_id: user.id }
+    else
+      return { success: false, reason: "incorrect password "}
```
toppy007: I like the reason statement. will take this for my own project.

schema_recipe.md
toppy007, line 1: well planned out. might be worth adding a .md file to the seeds db file i suggested adding with some of the info below in it.

spec/user_repository_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  context "when creating a new user" do
+    it "adds the user to the DB and returns nil" do
+      expect(repo.create(user)).to be_nil
+      expect(repo.all.length).to eq 4
+      expect(repo.all.last.name).to eq "John Smith"
+      expect(repo.all.last.email).to eq "john@smith.com"
+    end
+    xit "fails to add them given any missing info" do
```
toppy007: look at developing a validation method to run on the inputs incorporating a conditional statement to decide to run the create new user method.

  def invalid_request_parameters?
    return true if params[:title].nil? || params[:title].empty? ||
                    params[:content].nil? || params[:content].empty? 
    false
  end

views/login_denied.erb
toppy007, line 1: This is quite cool! im not sure how dynamic HTML is at controlling conditional statements will be interesting to get your take on it!

toppy007 (commented):
I know this is a work in progress so its a bit messy at the moment. looks like your got a nice structure of recording inputs and then requesting the data from the db. looks like your ready to begin displaying the required information as required.

PR #2180, opened 2023-04-12
Chitter Challenge (Not completed)

ghost:


app.rb
siqbal181, line 1: For the get '/' route, perhaps instead of doing the 'if' statements to return some text, you could instead initialize the user and do the 'if' statements in the erb file.

app.rb
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  get '/logout' do
+    session.destroy
+    redirect "/"
+  end
+
+  def invalid_peep_request_parameters?
+    # Are the params nil?
+    return true if params[:peep] == nil
```
siqbal181: Maybe add in some || pipes to lessen the if statements

app.rb
```diff
@@ -0,0 +1,162 @@
...
+  post '/signup' do
+    if invalid_signup_request_parameters?
+      status 400
+      return ''
+    end
+
+    repo = UserRepository.new
+
+    # check if the username is already taken
+    if repo.username_exists?(params[:username].downcase)
```
siqbal181, line 62: maybe take the downcase into the separate method to reduce the line length?

app.rb
```diff
@@ -0,0 +1,162 @@
...
+    user.email = params[:email].downcase
+    user.password = params[:password]
+
+    repo.create(user)
+    return erb(:user_created)
+  end
+
+  get '/login' do
+    # check if user already logged in
+    redirect '/' if session[:username] != nil
```
siqbal181: maybe we can take the redirect into another method and reduce code amount?

app.rb
```diff
@@ -0,0 +1,162 @@
...
+    end
+
+    repo = UserRepository.new
+    user = repo.find(params[:email])
+
+    if user
+      if BCrypt::Password.new(user.password) == params[:password]
+        session[:username] = user.username
+        session[:user_id] = user.id
+        @heading = "Hello, #{user.username}"
```
siqbal181: Maybe add the HTML messaging to the .erb file instead?

app.rb
```diff
@@ -0,0 +1,162 @@
...
+        redirect '/'
+      else
+        redirect '/login'
+      end
+    else
+      redirect '/login'
```
siqbal181: If time - can we try and tidy the if else returns

app.rb
```diff
@@ -0,0 +1,162 @@
...
+    # Are the params nil?
+    return true if params[:username] == nil || params[:name] == nil || params[:email] == nil || params[:password] == nil || params[:confirm_password] == nil
+    # Are they empty strings?
+    return true if params[:username] == "" || params[:name] == "" || params[:email] == "" || params[:password] == "" || params[:confirm_password] == ""
+    return false
+  end
+
+  def invalid_login_request_parameters?
+    # Are the params nil?
+    return true if params[:email] == nil || params[:password] == nil
```
siqbal181: Maybe in erb file add minlength to form params, so then ==nil is not needed?

lib/user_repository.rb
```diff
@@ -0,0 +1,62 @@
...
+  def username_exists?(username)
+    sql = "SELECT username FROM users WHERE username = $1;"
+    res = DatabaseConnection.exec_params(sql, [username])
+    if res.num_tuples.zero?
+      return false
+    end
+    return true
+  end
+
+  def email_exists?(email)
```
siqbal181, line 40: Maybe we can try pull in the exists methods and dynamically be able to check either username or password in one

siqbal181 (commented):
Hey Gary! - Great structure to the code and I like the way that you have added a layout to avoid repetitive HTML. I have left a few comments on perhaps how you could tighten some of the code. Let me know if you have any questions on slack. Sidra

PR #2184, opened 2023-04-13
Ongoing development

SunainaR:


Gemfile.lock
```diff
@@ -1,14 +1,42 @@
 GEM
   remote: https://rubygems.org/
   specs:
+    activemodel (7.0.4.3)
```
toppy007, line 4: OMG! this was a brave set to make. looking forward to you explaining it to me.

app.rb
```diff
@@ -0,0 +1,74 @@
...
+      return 'Incorrect email/password'
+    else
+      
+      return redirect('/')
+    end
+  end
+
+  private
+  def password_encryption(password)
+    encrypted_password = BCrypt::Password.create(password)
```
toppy007, line 70: Like how you've made the password method private.

docs/design.md
```diff
@@ -0,0 +1,157 @@
...
+email: text
+username: text
+password: text
+
+'users' can have many 'peeps' but a peep can only belong to one user: users (1) -> peeps (M)
+
+## 5. Write SQL for seed files
+
+psql -h 127.0.0.1 chitter_test < spec/seeds/table_creation.sql
+psql -h 127.0.0.1 chitter_test < spec/seeds/seeds_chitter.sql
```
toppy007, line 83: Couldn't find this file in the the repo!

app.rb
```diff
@@ -0,0 +1,74 @@
+require 'sinatra'
+require "sinatra/reloader"
+require 'sinatra/activerecord'
+require_relative './lib/peep'
+require_relative './lib/user'
+require_relative 'lib/database_connection'
+require 'bcrypt'
+
+# Need to edit database_connection later so this would work when deployed
+set :database, { adapter: "postgresql", database: "chitter_test" }
```
toppy007, line 10: Yeah noticed this! is this because you didn't want to fill up you deployed database with a load of rubbish when testing in rack-up?

not sure if this is best practice?

app.rb
```diff
@@ -0,0 +1,74 @@
...
+      password: password_encryption(params[:password])
+    )
+    return erb(:user_created)
+  end
+
+  get '/login' do
+    return erb(:login)
+  end
+
+  post '/login' do
```
toppy007, line 58: I think you need to have another method in this section 

decryption method (to take the encrypted password in the database and convert it back to the original password)

Dont worry the BCrypt gem does all the work just need to learn the method!

stored_password = BCrypt::Password.new(user.password)

not sure if that makes any sense or even if I'm correct tho but can't login when deploying the code on localhost!

lib/peep.rb
```diff
@@ -0,0 +1,3 @@
+# Active Record Model Class for peeps table
+class Peep < ActiveRecord::Base
```
toppy007, line 2: This looks like a game changer! nice one for going outside the comfy zone

lib/user.rb
```diff
@@ -0,0 +1,3 @@
+# Active Record Model Class for users table
+class User < ActiveRecord::Base
```
toppy007, line 2: This looks like a game changer! nice one for going outside the comfy zone

public/style.css
```diff
@@ -0,0 +1,11 @@
+body {
+  font-family: Verdana, Arial, Helvetica, sans-serif;
```
toppy007, line 2: OMG, both my brothers are graphic designers! They would be going crazy! three different type families in one project! very debatable

toppy007: Like how you've created a separate file system for the CSS. I should have done that

toppy007 (commented):
Hey Sunny. Thanks for letting me have a look at your project. It's good and great to see you got active record working! Brave direction to go so nice to see you took the challenge on. P.S I don't really know what I'm doing in these reviews!

First thing I would tackle is when I deployed the app.rb with rackup was being unable to pass the password conditional statement but I think thats a quick fix because I can see its in the database when I've changed the connection setting in the app.py

HTML is pretty limited in its style so don't worry! It made me smile to see three typeface families in your CSS file.

PR #2185, opened 2023-05-11
Charlie Parker Chitter Challenge

cshjp:
There is only one commit as I had to redo all my gemfiles when I had already completed everything but the app.rb, and it was easier to move everything over to a new chitter file than potentially mess up the one I had been working on.


app.rb
```diff
@@ -0,0 +1,78 @@
...
+    peeps
+  end
+
+  post '/' do
+    return erb(:login) if session[:user_id].nil?
+    repo = PeepRepo.new
+    peep = Peep.new
+    peep.message = params['peep']
+    peep.time = DateTime.now
+    peep.user_account_id = 1
```
cshjp, line 28: this was supposed to be peep.user_account_id = session[user_id]

